Issue #2, opened 2019-07-15
SAMRAI_config.h not found


Dekken:
built with 

```cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$PWD  ..```

fix - add to CMakeLists.txt

```
include_directories(
    ${CMAKE_CURRENT_SOURCE_DIR}/build/include
)
```

PR #8, opened 2019-09-11
add samrai include if build in tree

Dekken:
fix for https://github.com/PHAREHUB/PHARE/issues/2


Dekken:
required by #7 adding there

PR #23, opened 2019-09-19
Amrlib

nicolasaunai:
moves solvers, physical models and the MultiPhysicsIntegrator to a separate lib 'amr'
These classes are "easy" AMR components that use the easy interface provided by the samrai_interface module (messengers etc.) and don't know about low level SAMRAI details.
Now only the samrai_interface module makes explicit reference to SAMRAI and its API.

still WIP


src/amr/messengers/hybrid_messenger.h
```diff
@@ -6,20 +6,20 @@
...
-#include "evolution/messengers/messenger_info.h"
-#include "evolution/messengers/mhd_messenger.h"
+#include "messengers/hybrid_messenger_strategy.h"
+#include "messengers/messenger.h"
+#include "messengers/messenger_info.h"
+#include "messengers/mhd_messenger.h"
 #include "hybrid/hybrid_quantities.h"
-#include "physical_models/hybrid_model.h"
-#include "physical_models/physical_model.h"
+//#include "physical_models/hybrid_model.h"
```
Dekken: :(

tests/core/data/ions/test_main.cpp
```diff
@@ -30,7 +30,7 @@ using MaxwellianParticleInitializer1D = MaxwellianParticleInitializer<ParticleAr
 
 
 
-double density([[maybe_unused]] double x)
+double density(double x)
```
Dekken: :+1:

PR #46, opened 2019-10-07
python

nicolasaunai:
This PR links PHARE to the user script job.py for initialising the simulation.
PHARE reads init.py which imports job.py and takes user inputs from there to fill a PHAREDict using PyPHARE.
A test shows an example of a standard pharedict (to be completed with models and diagnostics later) obtained from the pharein simulation object.
The PhareDict will then be passed to various objects in the code for their initialization

### DIFF
--- a/src/initializer/data_provider.h
+++ b/src/initializer/data_provider.h
@@ -52,6 +52,8 @@ namespace initializer
         std::unique_ptr<PHAREDict<2>> phareDict2D;
         std::unique_ptr<PHAREDict<3>> phareDict3D;
 
+    private:
+        PHAREDictHandler() = default;
 
 
 
@@ -77,13 +79,13 @@ namespace initializer
         void stop()
         {
             if constexpr (dim == 1)
-                PHAREDictHandler::INSTANCE().phareDict1D.release();
+                phareDict1D.release();
 
             else if constexpr (dim == 2)
-                PHAREDictHandler::INSTANCE().phareDict2D.release();
+                phareDict2D.release();
 
             else if constexpr (dim == 3)
-                PHAREDictHandler::INSTANCE().phareDict3D.release();
+                phareDict3D.release();
         }
 
 
--- a/src/initializer/data_provider.h
+++ b/src/initializer/data_provider.h
@@ -2,7 +2,6 @@
 #define DATA_PROVIDER_H
 
 #include "cppdict/include/dict.hpp"
-//#include "models/physical_state.h"
 
 #include <map>
 #include <string>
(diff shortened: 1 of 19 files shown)


nicolasaunai:
I think it is good for review now. I need to change PhareIn UniformModel and and add another model to generate python functions for initial condition and plug that to the init.py and pharedict, but that will come in a separate PR.

src/initializer/data_provider.h
```diff
@@ -35,64 +34,80 @@ namespace initializer
...
 
-    template<std::size_t dim>
-    using VectorFunction = typename VectorFunctionHelper<std::array<double, 3>, dim>::type;
 
 
 
-    template<std::size_t dim>
-    using PHAREDict = cppdict::Dict<int, double, std::size_t, std::string, ScalarFunction<dim>,
-                                    VectorFunction<dim>>;
+    class PHAREDictHandler
```
Dekken: A private default constructor is perhaps a good idea so it blocks accidental non static access.

src/initializer/data_provider.h
```diff
@@ -35,64 +34,80 @@ namespace initializer
...
+                phareDict3D = std::make_unique<PHAREDict<3>>();
+            }
         }
-        else if constexpr (dim == 2)
+
+        template<std::size_t dim>
+        void stop()
         {
-            return phareDict2D;
+            if constexpr (dim == 1)
```
Dekken: start uses direct access (without INSTANCE()) - consider common usage

src/initializer/pyphare.cpp
```diff
@@ -7,29 +7,33 @@
 
 
 using PHARE::initializer::ScalarFunction;
-using PHARE::initializer::VectorFunction;
+// using PHARE::initializer::VectorFunction;
```
Dekken: :sob:

tests/amr/multiphysics_integrator/test_main.cpp
```diff
@@ -223,35 +253,60 @@ PHARE::initializer::PHAREDict<1> createIonsDict()
         = std::string{"MaxwellianParticleInitializer"};
     dict["ions"]["pop0"]["ParticleInitializer"]["density"] = static_cast<ScalarFunctionT>(density);
 
-    dict["ions"]["pop0"]["ParticleInitializer"]["bulkVelocity"]
-        = static_cast<VectorFunctionT>(bulkVelocity);
+    dict["ions"]["pop0"]["ParticleInitializer"]["bulk_velocity_x"]
```
Dekken: it might be too much work, but it's typical to use constexpr types for strings like this to avoid typos
e.g.
```
constexpr char* IONS_STR ="ions";
```

tests/initializer/test_initializer.cpp
```diff
@@ -48,8 +51,80 @@ TEST(ARestartTimeDataProvider, isADataProvider)
 
 
 
+TEST(APythonDataProvider, providesAValidTree)
+{
+    PHAREDictHandler::INSTANCE().init<1>();
+
+    char const* name                         = "init.py";
+    std::unique_ptr<PythonDataProvider> pydp = std::make_unique<PythonDataProvider>(2, name);
```
Dekken: superfluous, but why a unique_ptr?

PR #49, opened 2019-10-16
Phare main update for DictHandler

Dekken:


Dekken:
handled as part of diagnostics https://github.com/PHAREHUB/PHARE/pull/49

Issue #54, opened 2019-11-01
pharein submodule inconsistency


Dekken:
```
 git clone https://github.com/PHAREHUB/PHARE -b master

error: Server does not allow request for unadvertised object 9f1913f4e9c0d1fbe353ae566f476fee07ad2f9f
Fetched in submodule path 'subprojects/pharein', but it did not contain 9f1913f4e9c0d1fbe353ae566f476fee07ad2f9f. Direct fetching of that commit failed.
```

Dekken:
stopped happening

PR #55, opened 2019-11-01
tracking pharein

Dekken:
https://github.com/PHAREHUB/PHARE/issues/54


Dekken:
might have been github connectivity issue

PR #72, opened 2019-11-19
assert not nan in computeBulkVelocity

Dekken:


Dekken:
this PR is expected to fail for "test-messenger".

PR #90, opened 2019-12-14
Python diagnostics

nicolasaunai:
Not finished yet

- makes fluid, electromag and particle diagnostics  available from python job.py
- fixes simulation.py so that diagnostics are uniquely identified by their name (broken before)
- changes `rho_s` and `flux_s` where `_s` stands for `species` (former name of `populations`) to `density` and `flux`. 
- adds the `diag_type` `"bulk_velocity"`
- now `population_name` is the parameters that says to which population `diag_type` operates. User can chose among all population names declared in the initialization model and among (`"all"`, `"ions"`). If choosing `"all"` the `diag_type` (e.g. `"density"`) applies to all populations. If choosing `"ions"`, it applies to the total ions. Hence for `"ions"` only  `"density"` and `"bulk_velocity"` are permitted.


Dekken:
replaced by https://github.com/PHAREHUB/PHARE/pull/95

PR #99, opened 2020-01-14
time step needed when calling faraday

nicolasaunai:
Solving Faraday's equation 

Needs the time step defined on the patch.

This PR adds the time step as an argument to the operator(). Was hard-coded to 1 before.


Dekken:
merge?

PR #105, opened 2020-01-28
Add SAMRAI generated config file include dir

jeandet:
This is only needed when SAMRAI is used as subproject.

### DIFF
--- a/res/cmake/samrai.cmake
+++ b/res/cmake/samrai.cmake
@@ -16,12 +16,7 @@ if (NOT SAMRAI_FOUND)
     option(ENABLE_TESTS "Enable Samrai Test" OFF ) # disable SAMRAI Test so that we can use the googletest pulled after
 
     add_subdirectory(${SAMRAI_SRCDIR})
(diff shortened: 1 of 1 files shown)


Dekken:
This is done in https://github.com/PHAREHUB/PHARE/pull/104/files, which is also a prerequisite for the "*_SS" builds to complete

Dekken:
if you want to merge it @nicolasaunai go ahead. I'll rebase #104

jeandet:
In my understanding [this](https://github.com/PHAREHUB/PHARE/blob/bd7df63a620f74719b69ae84a82f9b6b16e7eb81/res/cmake/samrai.cmake#L24) is quite wrong since you may end up mixing a subproject build with a global install and it makes assumptions that this directory is created when you reach this line.

Dekken:
> In my understanding [this](https://github.com/PHAREHUB/PHARE/blob/bd7df63a620f74719b69ae84a82f9b6b16e7eb81/res/cmake/samrai.cmake#L24) is quite wrong since you may end up mixing a subproject build with a global install and it makes assumptions that this directory is created when you reach this line.

I don't see how

Dekken:
this change has been put in https://github.com/PHAREHUB/PHARE/pull/104/files

PR #108, opened 2020-01-30
Allow RNG seeding in maxwellian particle initialization

Dekken:


src/core/data/ions/particle_initializers/maxwellian_particle_initializer.h
```diff
@@ -84,6 +83,19 @@ namespace core
 
 
     private:
+        // can be relocated if needed
+        static std::mt19937_64 getRNG(std::optional<size_t> const& seed)
```
nicolasaunai: if I understand correctly this will lead to the same seed being used by all processes in case of parallel executions?

Dekken: this has nothing to do with MPI, so your assumption is correct.

I would think having a constant seed across all processes is a valid use case.

perhaps in python when we set the seed, we can use a flag/string for "MPI_RANK" to be explicit in our expectations.

```
ph.MaxwellianFluidModel(
    bx=bx, by=bx, bz=bx,
    ex=bx, ey=bx, ez=bx,
    protons={"charge":-1, "density":density, **vvv, "init":{"seed":1337}},
    alpha={"charge":-1, "density":density, **vvv, "init":{"seed":"mpi_rank"}},
)
```

as an example

Dekken: or rather we can expose some functions via pybind to return the mpi_rank or another value if MPI is not in use
e.g.

```
get_mpi_rank_or(value = None) # returns passed value if no MPI
get_mpi_rank() # returns 0 if no MPI
```

as None is a valid case which enables actual (pseudo) randomization.

Dekken:
once approved I'll squash

src/core/data/ions/particle_initializers/maxwellian_particle_initializer.h
```diff
@@ -84,6 +83,24 @@ namespace core
...
+                std::seed_seq seed_seq{randSeed(), randSeed(), randSeed(), randSeed(),
+                                       randSeed(), randSeed(), randSeed(), randSeed()};
+                return std::mt19937_64(seed_seq);
+            }
+            return std::mt19937_64(*seed);
+        }
+
+        template <typename T = float>
+        static std::uniform_real_distribution<T> get_unif_dist(){
+          return std::uniform_real_distribution<T>(0, 1. - std::numeric_limits<T>::epsilon());
```
nicolasaunai: 1.- epsilon is a bit magical here.
I know what it's for but we should probably either put a comment to explain the context or return that from a properly named function?

PR #111, opened 2020-02-03
add average for fields and vecfields and slight refactoring

nicolasaunai:
This PR adds the code that is necessary to average two fields and vecfields necessary in the predictor predictor corrector solver

image

### DIFF
--- a/src/core/data/field/field.h
+++ b/src/core/data/field/field.h
@@ -78,7 +78,7 @@ namespace core
     template<typename Field>
     void average(Field const& f1, Field const& f2, Field& avg)
     {
-        std::transform(std::begin(f1), std::end(f2), std::begin(f2), std::begin(avg),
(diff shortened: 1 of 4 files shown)


src/core/data/field/field.h
```diff
@@ -71,6 +72,21 @@ namespace core
         std::string name_{"No Name"};
         PhysicalQuantity qty_;
     };
+
+
+
+
+    template<typename Field>
+    void average(Field const& f1, Field const& f2, Field& avg)
```
Dekken: could cause conflicts if another templated "average" function is ever made. (with three parameters)
perhaps as a static function on the class?

Dekken: or alternatively something like 
```
    template<typename T>
    void average(Field<T> const& f1, Field<T> const& f2, Field<T>& avg)
```

would block conflicts

PR #119, opened 2020-02-11
add setLayout and LayoutHolder and refactors numerical algos

nicolasaunai:
Faraday, Ohm, Ampere, etc. need a GridLayout to work. This GridLayout is necessarily built from a patch, in a patch level loop. They need to have an internal GridLayout set, and unset.

The current code sets their grid layout pointer, but does not unset it.

This PR proposes to unset it automatically as is done for resourcesPointers by setOnPatch.
It also refactors faraday etc. to all inherit from a single class holding the GridLayout pointer for them

### DIFF
--- a/src/core/numerics/ampere/ampere.h
+++ b/src/core/numerics/ampere/ampere.h
@@ -5,6 +5,7 @@
 #include <iostream>
 
 #include "core/data/grid/gridlayoutdefs.h"
+#include "core/data/grid/gridlayout_utils.h"
 #include "core/data/vecfield/vecfield_component.h"
 #include "core/utilities/index/index.h"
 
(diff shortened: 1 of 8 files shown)


src/core/numerics/ampere/ampere.h
```diff
@@ -248,33 +179,23 @@ namespace core
...
-            if (!impl_.hasLayoutSet())
+            if (!this->hasLayout())
             {
                 throw std::runtime_error(
                     "Error - Ampere - GridLayout not set, cannot proceed to calculate ampere()");
             }
             std::cout << "I'm solving ampere " << GridLayout::dimension << "\n";
 
-            impl_(B, J);
+            this->operator()<VecField, GridLayout::dimension>(B, J);
```
Dekken: cosmetic:

Vecfield here is inferrable from B/J if the template definitions are swapped around, like

```
template<std::size_t dim, typename VecField, std::enable_if_t<dim == 1, int> = 0>`
```

Issue #12, opened 2019-09-16
Electrons


nicolasaunai:
# Objective

The objective of this issue is to implement the `Electrons`.
`Electrons` is an data object, like `Ions`, it is part of the `HybridState`.
It is used in the hybrid PPC solver to 

- get the electron density
- get the bulk velocity
- get electron pressure

Electron moments could a priori be obtained from different assumptions. The most frequent are:

- _quasi-neutrality_, which says that:
    - Ne = Ni
    - Ve = Vi - J/(Ne*e)

- isothermal pressure closure
   - which says that the electron temperature Te is a spatio-temporal constant
   - Pe = Ne*Te

The pressure could be obtained from another closure equation, such as adiabatic, polytropic, etc. More rarely, the electron density and bulk velocity could also be obtained from other assumptions (like evolving the electron fluid equation, or evolving particles etc.).

The was moments are calculated should therefore, a priori, not be known at the solver level.

Most probably, the different possible ways of getting electron moments will need informations from other `HybridState` quantities.

# Problem ?

In the PPC solver, starting from all quantities known at time `n`, `electronMomentModel` is called after `Faraday` and `Ampère`, just before `Ohm`. Hence, we have Ne = Ni at `n`, **V**i at `n` and **J** at `n+1`. Then, **V**e = **V**i-**J**/N is calculated with a mix of quantities given at either `n`, or `n+1`. It should not be a big issue as it is needed for the calculation of **E** in Ohm, but it is not very nice. ..

PR #131, opened 2020-02-27
Electrons

nicolasaunai:
#12 

Remains to be done:

put Electrons in the HybridState
have the pressure closure (compile time for now)


nicolasaunai:
rebase onto master should fix the clang failure

Issue #120, opened 2020-02-15
possible MPI / smallest patch size issue


Dekken:
refinement_boxes": {"L0": {"B0": [(10,), (14,)]}}

 doesn't work with

    mpirun -n 2

it only starts to work when 14 is changed to 19

nicolasaunai:
still open?

Dekken:
it's fixed in https://github.com/PHAREHUB/PHARE/pull/106

perhaps we can fix it in isolation, the refinement_box tests on master (tests/simulator/py) should suffice

do we want to reproduce a minimal test to show that flush can be synchronous?

PR #136, opened 2020-03-03
rm hdf5 flush in case call is not collective

Dekken:
https://github.com/PHAREHUB/PHARE/issues/120


Dekken:
incorporates https://github.com/PHAREHUB/PHARE/pull/135

tests/simulator/py/refinement_boxes.py
```diff
@@ -19,7 +19,7 @@ def dup(dic):
         return dic
 
     valid1D = [
-        dup({"refinement_boxes": {"L0": {"B0": [(10,), (50,)]}}}),
+        dup({"refinement_boxes": {"L0": {"B0": [(10,), (14,)]}}}),
```
nicolasaunai: could we instead of 10 and 14 use 10 and nbrGhost-1 or something?
and maybe add a comment that this test case is done so we have only 1 patch on the refined level 1?

Dekken: done

PR #145, opened 2020-03-10
init pybind hierarchy data exports

Dekken:
if LLNL/SAMRAI#112 is not merged, this ticket requires SAMRAI libraries to be shared, at least on linux
see: https://github.com/Dekken/llnl.samrai/blob/master/mkn.sh#L15

This includes a refactor of MPI functions used in diagnostics, but does not remove the existing code from diagnostics - TBD tho

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,7 @@ add_subdirectory(src/initializer)
 add_subdirectory(src/phare)
 add_subdirectory(src/amr)
 add_subdirectory(src/simulator)
+add_subdirectory(src/python3)
 
 #* Documentation option
 include("${PHARE_PROJECT_DIR}/res/cmake/doc.cmake")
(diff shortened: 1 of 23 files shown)


Dekken:
we may be able to prevent a copy if we move the vectors on the fields
currently we only expose the data as an array and the size

Dekken:
this relies on https://github.com/LLNL/SAMRAI/pull/112

Dekken:
we can possibly get around the SAMRAI VariableDatabase instance problem if we put a forwarding function somewhere and only use that. to be tested.

Dekken:
if https://github.com/LLNL/SAMRAI/pull/112 is not merged, this ticket requires SAMRAI libraries to be shared, at least on linux
see: https://github.com/Dekken/llnl.samrai/blob/master/mkn.sh#L15

Dekken:
there's some extra tests that could be added to this when #106 is merged

PR #150, opened 2020-03-15
hdf5 pains

Dekken:


res/cmake/highfive.cmake
```diff
@@ -11,6 +11,15 @@ if(HighFive)
   endif()
 
   message("HighFive enabled - checking HDF5")
+  # when using -DHDF5_ROOT=/path/to/hdf5 when hdf5 is build with cmake
+  #  variables that are expected to be set, do not appear to be
```
nicolasaunai: Is this something we want to report here https://forum.hdfgroup.org/c/hdf5 ? (I would say yes to se what they say?)
is it specific to 1.12.x ? or does it also apply to 1.10.x ?
why is it not reported already? for instance in highfive's GH?

Issue #160, opened 2020-03-30
cppdict deprecation


Dekken:
cppdict:is_in is -> renamed and private
https://github.com/LaboratoryOfPlasmaPhysics/cppdict/pull/10/files#diff-41ba0660fec8ee545fc00b145e2eb6ebR20

+ remove https://github.com/PHAREHUB/PHARE/blob/master/src/core/utilities/types.h#L92

PR #161, opened 2020-03-30
pre cppdict deprecation of is_in

Dekken:
https://github.com/PHAREHUB/PHARE/issues/160

### DIFF
--- a/src/initializer/pyphare.cpp
+++ b/src/initializer/pyphare.cpp
@@ -14,15 +14,7 @@ template<typename T>
 using py_array = py::array_t<T, py::array::c_style | py::array::forcecast>;
 
 
-template<typename T1, typename... T2>
-constexpr bool is_in()
-{
-    return std::disjunction_v<std::is_same<T1, T2>...>;
-}
-
-template<typename T,
-         typename = std::enable_if_t<is_in<T, int, double, std::string, ScalarFunction<1>,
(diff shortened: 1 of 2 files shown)


Dekken:
https://github.com/LaboratoryOfPlasmaPhysics/cppdict/pull/10

src/initializer/pyphare.cpp
```diff
@@ -13,9 +13,16 @@ namespace py = pybind11;
...
+template<typename T1, typename... T2>
+constexpr bool is_in()
+{
+    return std::disjunction_v<std::is_same<T1, T2>...>;
+}
+
 template<typename T,
-         typename = std::enable_if_t<cppdict::is_in<T, int, double, std::string, ScalarFunction<1>,
-                                                    ScalarFunction<2>, ScalarFunction<3>>>>
+         typename = std::enable_if_t<is_in<T, int, double, std::string, ScalarFunction<1>,
```
nicolasaunai, line 24: isn't this list supposed to equal that of https://github.com/PHAREHUB/PHARE/blob/master/src/initializer/data_provider.h#L45 ?
if yes, it would be nice not to repeat, if no why ?

Dekken: the second template is redundant in my opinion, but I didn't write it so I can't answer you.

Dekken: this has been removed.

Issue #172, opened 2020-04-06
electrons have wrong ref to electromag


nicolasaunai:
when they are used : 

```
 for (auto& patch : level)
            {
                auto layout = PHARE::amr::layoutFromPatch<GridLayout>(*patch);
                auto _      = resourcesManager->setOnPatch(*patch, Bpred, Epred, J, electrons);
                electrons.update(layout);
                auto& Ve = electrons.velocity();
                auto& Ne = electrons.density();
                auto& Pe = electrons.pressure();
                ohm_(Ne, Ve, Pe, Bpred, J, Epred);
                resourcesManager->setTime(Epred, *patch, newTime);
            }
```

electrons.update(layout) fails because electrons are not usable.
This is because electrons, to be usable, need the model Electromag to be usable. 
However, when used, the model Electromag is not the one relevant. Instead, the internal solver EMPred is the one relevant.

For now : 

- either remove the ref to electromag, which was there in prevision of EM being useful for later Pressure closures anyway
- ask for Electromag ref when update. Same could go for Ions etc.

PR #173, opened 2020-04-07
remove electromag in electrons

nicolasaunai:
#172 
I decided to remove the Electromag ref after all.


nicolasaunai:
see #172 
we'll put Electromag back somewhere (probably as an argument for update()) when needed

PR #188, opened 2020-04-21
simplify constexpr is_hybrid_model<Model> check

Dekken:


src/solver/physical_models/hybrid_model.h
```diff
@@ -146,17 +143,15 @@ namespace solver
...
     {
-        using type = decltype(hybrid_model_from_type_list(std::declval<TypeList>()));
-    };
+        return std::is_same_v<Model<Args...>, HybridModel<Args...>>;
+    }
 
-    template<typename TypeList>
-    using type_list_to_hybrid_model_t = typename type_list_to_hybrid_model<TypeList>::type;
+    template<typename Model>
+    bool constexpr is_hybrid_model = _is_hybrid_model((Model*)(nullptr));
```
nicolasaunai, line 154: why Model* ?
how do we know it's going to get a pointer?

Dekken: Model* let's us use the type, without declval or decltype.

I don't get your second question.

Dekken: we need the type as a function parameter, in order to expand the template from
Type
to
Type<Args...>
which is achieved through parameter type inference 
using a pointer is simple as we don't have an instance of the type, nor do we need one.

nicolasaunai: so `Model` is actually `ModelPtr`?

Dekken: template<typename Model> <- is not a pointer, and never should be.

if it was, even the current code wouldn't work.

PR #209, opened 2020-05-08
Create lgtm.yml

Dekken:
Add lgtm.yml

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,28 @@
+path_classifiers:
+  tag:
+    exclude: "subprojects"
+  docs:
+  - "README.md"
+  - "LICENSE.txt"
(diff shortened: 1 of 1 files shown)


Dekken:
The LGTM build seems to timeout, we could consider importing the TC SAMRAI build to reduce the build time, whether these are compatible is uncertain as I believe LGTM is Debian based

Dekken:
TC SAMRAI build may not be incompatible, but it is certainly inconvenient.
1. It installs to /root/.local-samrai, we have no sudo user on LGTM.
2. OpenMPI libs are located differently on Ubuntu.

I've put a SAMRAI Debian build up on PHARCHIVE and testing it currently on LGTM

PR #212, opened 2020-05-12
Alerts From LGTM

Dekken:
https://lgtm.com/projects/g/Dekken/NotPHARE/?mode=list


nicolasaunai:
it seems many files are missing (for instance I didn't find the python data provider). There are no tests also.

Dekken:
> it seems many files are missing (for instance I didn't find the python data provider). There are no tests also.

Perhaps you can consider this PR for what it is, which is the fix for the alerts shown.
If you find there is something missing from LGTM, perhaps make an issue.

nicolasaunai:
> Perhaps you can consider this PR for what it is, which is the fix for the alerts shown.

You should consider explaining what the PR is about in the PR message if you don't want off topic comments ;-)

> If you find there is something missing from LGTM, perhaps make an issue.

creating an issue about something that's being PRed?

Issue #182, opened 2020-04-14
Electric field should be initialized


nicolasaunai:
The electric field is currently initialized with a python function, in a way totally independent of the other variables (electric field, ion moments etc.). Most of the time, it is initialized to 0.

As a result of being zero : 

- the first Faraday leaves the magnetic field untouched since the magnetic field derivatives equals the curl of zero, and so is itself zero, so Bpred in pred1 is equal to B. That's not a super big deal but not ideal.

- After Ohm calculates Epred, the average of E and Epred is computed into Eavg defined at t=n+1/2. Given that E is initially 0, Evag now contains half Epred, so half the electric field only is used to push ions. This is **wrong**.

as a result, the electric field should be initialized, in a way that is consistent with the Ohm's law and the other user input quantities. Several options:

- calculate the electric field in the python script by implementing python ampere (J is needed ohm) and ohm equations. That has the advantage of not touching the cpp code, which would now be initialized from ex,ey,ez functions returning fields consistent with ohm and other inputs. But that has the drawback of having to develop ampere and ohm in python, and other possible dataproviders may not be able to do that.

- calculate ohm in cpp, at intialization, based on other quantities. In this case, it should be done here 

https://github.com/PHAREHUB/PHARE/blob/master/src/solver/level_initializer/hybrid_level_initializer.h#L87 

once the moments have been calculated on all levels, we make a similar patch loop to calculate J with ampere, then another patch loop to calculate E from ohm, and we're good.

Second method is I think preferable.

PR #214, opened 2020-05-13
add initialization of the electric field

nicolasaunai:
#182

### DIFF
--- a/src/core/numerics/faraday/faraday.h
+++ b/src/core/numerics/faraday/faraday.h
@@ -1,5 +1,5 @@
-#ifndef PHARE_CORE_NUMERICS_FARADAY_FARADAY_H
-#define PHARE_CORE_NUMERICS_FARADAY_FARADAY_H
+#ifndef PHARE_FARADAY_H
+#define PHARE_FARADAY_H
 
 #include <cstddef>
 #include <iostream>
--- a/src/simulator/simulator.h
+++ b/src/simulator/simulator.h
@@ -122,10 +122,12 @@ public:
         catch (const std::runtime_error& e)
         {
             std::cerr << "EXCEPTION CAUGHT: " << e.what() << std::endl;
+            std::rethrow_exception(std::current_exception());
         }
         catch (...)
         {
             std::cerr << "UNKNOWN EXCEPTION CAUGHT" << std::endl;
+            std::rethrow_exception(std::current_exception());
         }
     }
 
--- a/src/simulator/simulator.h
+++ b/src/simulator/simulator.h
@@ -110,13 +110,23 @@ public:
 
 
 
-
     void initialize() override
     {
-        if (integrator_ != nullptr)
-            integrator_->initialize();
-        else
-            throw std::runtime_error("Error - Simulator has no integrator");
+        try
+        {
+            if (integrator_ != nullptr)
+                integrator_->initialize();
+            else
+                throw std::runtime_error("Error - Simulator has no integrator");
+        }
+        catch (const std::runtime_error& e)
+        {
+            std::cerr << "EXCEPTION CAUGHT: " << e.what() << std::endl;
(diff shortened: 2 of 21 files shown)


nicolasaunai:
ok this PR solves a bug in the projection methods in gridlayout that was Hidding since we've hard coded ghosts to be 5. When they depended on the interp order, dual or primal nodes had different nbr of ghosts nodes and the functions dualToPrimal() and primalToDual() had to return something different depending on interpOrder... this is not the case anymore but I have kept the code as commented out with a comment to explain that it needs to be restored if/when the ghost nbr changes

src/simulator/simulator.h
```diff
@@ -110,13 +110,23 @@ class Simulator : public ISimulator
...
 
-
     void initialize() override
     {
-        if (integrator_ != nullptr)
-            integrator_->initialize();
-        else
-            throw std::runtime_error("Error - Simulator has no integrator");
+        try
+        {
```
Dekken, line 116: This can leave the class in an invalid state, it should possibly rethrow.

example:

```
    try{
        throw std::runtime_error("VERY BAD");
    }
    catch (const std::runtime_error& e)
    {
        std::cout << e.what() << std::endl;
        std::rethrow_exception(std::current_exception());
        // or 
        //throw e;
    }
    catch (...)
    {
        std::cerr << "UNKNOWN EXCEPTION CAUGHT" << std::endl;
        exit(11); // exit on unknown error? should never happen, state is invalid if it does.
    }
```

Dekken: more a discussion point than requested change.

nicolasaunai: I have no opinion about that, tell me if I need to do copy/paste your snippet

Dekken: here is the complete

```

    void initialize() override
    {
        try
        {
            if (integrator_ != nullptr)
                integrator_->initialize();
            else
                throw std::runtime_error("Error - Simulator has no integrator");
        }
        catch (const std::runtime_error& e)
        {
            std::cerr << "EXCEPTION CAUGHT: " << e.what() << std::endl;
            std::rethrow_exception(std::current_exception());
        }
        catch (...)
        {
            std::cerr << "UNKNOWN EXCEPTION CAUGHT" << std::endl;
            std::rethrow_exception(std::current_exception());
        }
    }
```

nicolasaunai: done

Dekken:
should E still be configurable via python after this?

nicolasaunai:
> should E still be configurable via python after this?

not really...

src/core/numerics/faraday/faraday.h
```diff
@@ -1,5 +1,5 @@
-#ifndef PHARE_CORE_NUMERICS_FARADAY_FARADAY_H
-#define PHARE_CORE_NUMERICS_FARADAY_FARADAY_H
+#ifndef PHARE_FARADAY_H
```
Dekken, line 1: Lost opportunity for "PHARE-ADAY" smh

src/solver/level_initializer/hybrid_level_initializer.h
```diff
@@ -85,8 +99,49 @@ namespace solver
...
+                    auto __  = core::SetLayout(&layout, ohm_);
+                    ohm_(Ne, Ve, Pe, B, J, E);
+                    hybridModel.resourcesManager->setTime(E, *patch, 0.);
+                }
+
+                hybMessenger.fillElectricGhosts(E, levelNumber, 0.);
+            }
         }
-    };
+    }; // namespace solver
```
Dekken, line 144: probable typo

nicolasaunai: I think clang format puts that automatically.... will change

PR #218, opened 2020-05-15
2d Particle schedule test 

Dekken:


tests/amr/data/particles/schedule/test_tag_strategy.h
```diff
@@ -79,29 +78,44 @@ class TagStrategy : public SAMRAI::mesh::StandardTagAndInitStrategy
...
+                                particle.charge = 1.;
+                                particle.v      = {{1.0, 0.0, 0.0}};
 
-                        particle.delta[dirX] = middle - delta;
-                        interior.push_back(particle);
+                                interior.push_back(particle);
 
-                        particle.delta[dirX] = middle + delta;
-                        interior.push_back(particle);
+                                particle.delta[dirX] = middle - delta;
```
nicolasaunai, line 112: cannot be always dirX

tests/amr/data/particles/schedule/test_main.cpp
```diff
@@ -57,30 +70,39 @@ TEST_F(ALevelWithDomainParticles, ghostParticleNumberisZeroBeforeScheduleAndCorr
     std::cout << "filling ghosts\n";
     schedule->fillData(0.);
 
+    size_t eq = std::get<dim - 1>(std::forward_as_tuple(3 * 2, 3 * 5 * 4));
```
nicolasaunai, line 73: explain the numbers, maybe with variables (i.e. not comments, if possible)

tests/amr/data/particles/schedule/test_tag_strategy.h
```diff
@@ -79,29 +78,44 @@ class TagStrategy : public SAMRAI::mesh::StandardTagAndInitStrategy
...
+                                Particle<dimension> particle;
 
-                        particle.weight = 1.;
-                        particle.charge = 1.;
-                        particle.v      = {{1.0, 0.0, 0.0}};
+                                for (size_t i = 0; i < dimension; i++)
+                                    particle.iCell[i] = _3DiCell[i];
 
-                        particle.iCell[dirX] = iCellPos;
+                                particle.delta.fill(middle);
```
nicolasaunai, line 105: Im' not sure it matters for these tests where the particles are exactly, I didn't look... but I don't think we want to put `dimension` times a particle in the middle of the cell, that's weird. Probably better to dispatch them in distinct places.

Dekken: > I don't think we want to put dimension times a particle in the middle of the cell

I don't understand this.

> Probably better to dispatch them in distinct places.

I would state that they are.

nicolasaunai: I misread sorry

tests/amr/data/particles/schedule/test_tag_strategy.h
Dekken: can you elaborate on this please

nicolasaunai: I don't know to why extent the dispatching matters for the tests (I don't remember) but I find it weird if it's 2D to put 3 particles in the cell in the same way it is done in 1D. I'd had imagined something more 2D that dispatch them in the cell with a delta in Y too.

tests/amr/data/particles/schedule/test_main.cpp
Dekken: image

tests/amr/data/particles/schedule/test_tag_strategy.h
Dekken: For what the test does I'm not sure it matters.

consider the test names.

```
ghostParticleNumberisZeroBeforeScheduleAndCorrectAfterSchedule;
hasGhostParticlesInGhostAMRCellsAfterSchedule;
hasGhostParticleFilledFromNeighborFirstCell;
```

tests/amr/data/particles/schedule/test_main.cpp
nicolasaunai: why `Cells{5}`?

Dekken: updated

nicolasaunai: clearer. but sorry I can't find where you get these numbers from.
All I see in the input is that the domain is (0,7) in both directions.

Dekken: ```https://github.com/PHAREHUB/PHARE/pull/218/files#diff-681d5e53cf20eda2a6298fe996f65536R16```

every level0 cell is refined so we end up with 4 4*4 patches given smallest/largest patch_size == 4

4*4 ghost_box is 6*6
6*6 patch border cells == 5*4 (or (6*2) + (4*2))

PR #226, opened 2020-05-20
GCC 10 -fanalyzer

Dekken:
replace pointer dynamic_casts with & casts which throw on failure.
```
src/amr/data/field/field_data.h 
src/amr/data/particles/particles_data.h 
```

emplace_back rather than push_back(move(thing))
```
 tests/core/utilities/box/test_main.cpp 
```

the static analysis logs about the casts was about null deref that I couldn't see happening, but changing it to use ref cast got rid of that it seems. I don't see a point where these functions should be called and we should allow classes other than our own, so throw.

### DIFF
--- a/src/amr/data/field/field_data.h
+++ b/src/amr/data/field/field_data.h
@@ -88,47 +88,35 @@ namespace amr
 
             TBOX_ASSERT_OBJDIM_EQUALITY2(*this, source);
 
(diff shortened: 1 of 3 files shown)


Dekken:
proof for casts throw on failure

http://coliru.stacked-crooked.com/a/54b0a4d01efd29cb

Dekken:
most of the changes in this are white space for dropping some if statements

PR #230, opened 2020-05-26
ND electrons computeBulkVelocity

Dekken:
the changes ndarray_vector.h may be unacceptable, I thought it was useful to remove "if constexpr dim == "

this was a guess: 
https://github.com/PHAREHUB/PHARE/pull/230/files#diff-5bc7e20169f26980f29061f3d35cb9d4R99

### DIFF
--- a/src/core/data/ndarray/ndarray_vector.h
+++ b/src/core/data/ndarray/ndarray_vector.h
@@ -9,140 +9,136 @@
 #include <numeric>
 
 
-namespace PHARE
+namespace PHARE::core
 {
-namespace core
+template<std::size_t dim, typename DataType = double>
+class NdArrayVector
 {
-    template<std::size_t dim, typename DataType = double>
-    class NdArrayVector
+public:
+    NdArrayVector() = delete;
+
+    template<typename... Nodes>
+    NdArrayVector(Nodes... nodes)
+        : nCells_{nodes...}
+        , data_((... * nodes))
     {
-    public:
-        NdArrayVector() = delete;
+    }
 
-        template<typename... Nodes>
-        NdArrayVector(Nodes... nodes)
-            : nCells_{nodes...}
-            , data_((... * nodes))
-        {
-        }
-
-        explicit NdArrayVector(std::array<std::uint32_t, dim> const& ncells)
-            : nCells_{ncells}
-            , data_(std::accumulate(ncells.begin(), ncells.end(), 1, std::multiplies<int>()))
-        {
-        }
-
-
-        NdArrayVector(NdArrayVector const& source) = default;
-        NdArrayVector(NdArrayVector&& source)      = default;
-
-        static constexpr bool is_contiguous = 1;
-
-        auto data() const { return data_.data(); }
-
-        auto size() const { return data_.size(); }
-
-        auto begin() const { return std::begin(data_); }
-        auto begin() { return std::begin(data_); }
-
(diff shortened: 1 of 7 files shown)


Dekken:
> can be merged once the lower/upper XYZ are not taken form XYZ components of Ve

even with the ndarray changes?

nicolasaunai:
> > can be merged once the lower/upper XYZ are not taken form XYZ components of Ve
> 
> even with the ndarray changes?

well... what's the alternative ? I see two : 

- back to if constexpr...
- get rid of the 3 int operator() in field 1D and 2D and make operator(Index<dim>(ix,iy,iz)) that does basically what your sub_array does, i.e. makes an index ignoring all above dim-1.
- something else?

Dekken:
I'll see about #2

src/core/data/ndarray/ndarray_vector.h
```diff
@@ -134,12 +146,7 @@ namespace core
     template<std::size_t dim>
     auto makeNdArray(std::array<std::uint32_t, dim> sizes)
     {
-        if constexpr (dim == 1)
-            return NdArrayVector<1>{sizes[0]};
-        if constexpr (dim == 2)
-            return NdArrayVector<2>{sizes[0], sizes[1]};
-        if constexpr (dim == 3)
-            return NdArrayVector<3>{sizes[0], sizes[1], sizes[2]};
+        return NdArrayVector<dim>{sizes};
```
Dekken, line 149: this function (makeNdArray) seems kinda redundant now

nicolasaunai: did you try to delete it?

Dekken: nope but I will I think

Dekken:
regarding the commit message "might want to profile this"

I believe was running with "-O0 -g3", whereas "-O3 -g0" runs in much less time (as expected)

src/core/data/ndarray/ndarray_vector.h
Dekken: removed

tests/simulator/refined_particle_nbr.py
```diff
@@ -119,9 +119,9 @@ def test_1d(self):
     PREVIOUS_ITERATION_MAX_DIFF_2d = 1.50
 
     ## blocked by "electrons.h computeBulkVelocity()" - not updated for 2d
```
nicolasaunai, line 121: remove?

tools/bench/sim/split.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+#
+# formatted with black
+import os
+os.environ["OMP_NUM_THREADS"] = "1" # export OMP_NUM_THREADS=1
```
nicolasaunai, line 5: could you comment a bit on why we have to set these env var?

Dekken: done

tests/simulator/refined_particle_nbr.py
Dekken: dpme

nicolasaunai:
question : do we keep these 8 commits into master? (I'm not sure they all build)

Dekken:
> question : do we keep these 8 commits into master? (I'm not sure they all build)

I'ma squash

PR #231, opened 2020-05-28
NdArrayVector is templated by dim

rochSmets:
associated tests are not yet done...


src/core/data/ndarray/ndarray_vector.h
```diff
@@ -5,12 +5,216 @@
...
+            this->data_ = std::move(source.data_);
+            return *this;
+        }
+
+
+        template<typename... Indexes>
+        DataType& operator()(Indexes... indexes)
+        {
+            auto params = std::tuple<Indexes...>{indexes...};
+            static_assert(sizeof...(Indexes) == dim);
```
Dekken, line 120: Could replace the op per dim

```
static_assert((... && std::is_unsigned_v<decltype(indexes)>));
```

src/core/data/ndarray/ndarray_vector.h
```diff
@@ -5,12 +5,216 @@
...
+            }
+
+            if constexpr (dim != 1 && dim != 2 && dim != 3)
+            {
+                return 0;
+            }
+        }
+
+        template<typename... Indexes>
+        DataType const& operator()(Indexes... indexes) const
```
Dekken, line 162: https://github.com/PHAREHUB/PHARE/blob/master/src/core/data/ions/ion_population/ion_population.h#L167

this can be used to have only one real function declaration

@nicolasaunai what do you think?

nicolasaunai: probably worth doing in that case would spare many lines

src/core/data/ndarray/ndarray_vector.h
```diff
@@ -5,12 +5,216 @@
...
+                static_assert(std::is_unsigned<decltype(i)>::value);
+                static_assert(std::is_unsigned<decltype(j)>::value);
+                static_assert(std::is_unsigned<decltype(k)>::value);
+
+                return this->data_[k + j * nCells_[2] + i * nCells_[1] * nCells_[2]];
+            }
+
+            if constexpr (dim != 1 && dim != 2 && dim != 3)
+            {
+                return 0;
```
nicolasaunai, line 202: I don't think we want to return 0 (or anything for that matter) in that case.. if dim is invalid we should not even be here in the first place.

src/core/data/ndarray/ndarray_vector.h
nicolasaunai: I think having (Indexes... indexes) with the static_assert that checks we have dim arguments will break @Dekken PR #230

Dekken: I'm going to dry the Index/sub_array thing so it should be ok

rochSmets: I'm not that sure to understand the suggestion ; and it appears to me less talkative. but if it's a better solution, you guys will have to explain it to me.

src/core/data/ndarray/ndarray_vector.h
rochSmets: now, it throws an runtime error,

src/core/data/ndarray/ndarray_vector.h
```diff
@@ -5,12 +5,216 @@
...
+
+        auto end() const { return std::end(data_); }
+        auto end() { return std::end(data_); }
+
+        void zero() { data_ = std::vector<DataType>(data_.size(), {0}); }
+
+
+        NdArrayVector& operator=(NdArrayVector const& source)
+        {
+            if (nCells_[0] != source.nCells_[0])
```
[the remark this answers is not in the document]
rochSmets: yes it does, thx

src/core/data/ndarray/ndarray_vector.h
rochSmets: hmmm, I will need your help on this one : these comments are not explicit to me.

src/core/data/ndarray/ndarray_vector.h
rochSmets: I will also need your help on this one,

Dekken:
This PR shouldn't have all these other commits in it.
I've remade this branch here: https://github.com/Dekken/PHARE/commits/NdArray
such that I believe the commits are ordered properly.

Dekken:
I guess this obsolete now https://github.com/PHAREHUB/PHARE/pull/240

Issue #181, opened 2020-04-14
coarsening moments and NaNs on moment ghosts


nicolasaunai:
This issue concerns the coarsening of moments only.
Moments are defined on primal nodes.

At refinement ratio 2 a coarse grid node needs 3 fine nodes, the one on top, and the two next to it on the left and right side, with coefs 0.5, 0.25 0.25.

![image](https://user-images.githubusercontent.com/3200931/79219555-cb36b680-7e52-11ea-9d9f-d4a6d3d20072.png)

the problem is that for the coarse nod that falls on top of the fine patch border node, that node needs the first fine primal ghost... which is currently set to NaN.

as a result, many coarse node (all those falling on top of a fine border node) will have NaNs, which is bad.

Several options:

- do not set nan on the first primal moment node. But that requires more ghost particles than we have now so that this node gets a complete deposit of moments. This could be costly (more particle communications, more deposits, more pushing...)

- have a special condition for coarse nodes that fall onto fine patch borders when coarsening and make them use only the fine node on top.

![image](https://user-images.githubusercontent.com/3200931/79219845-45ffd180-7e53-11ea-90c2-609feb81d453.png)

second option seems easier even if introducing a special case...

second option variant could use the fine node on top AND the one inside, but the dissymmetry is weird.

nicolasaunai:
Actually, the problem is only for moments...

it could be possible,  since we have the field in the coarsening operator, to check with its name if it's a moment and then we'd have to look wether a fine ghost node appears in the fine indexes to coarsen and then skip the ghost and take only 1 fine node, as said above...

but this seems complicated.
I'm investigating instead another method that consists in not placing a NaN on the first ghost node, but copying the first physical index there...

![image](https://user-images.githubusercontent.com/3200931/83032915-bbf07e80-a036-11ea-8ed6-86b13d23d14d.png)

this way no NaNs anymore when coarsening...

PR #234, opened 2020-06-01
advance further

nicolasaunai:
fix bugs for 2 way coupling evolution


src/core/data/ions/ions.h
```diff
@@ -252,10 +252,23 @@ namespace core
                 }
             };
 
+
+            auto copy = [ix0, ix1](auto& pop) {
```
Dekken, line 256: I think I know what this does, but I don't know why.

Can there be a little comment stating why it's set to the border node values?

nicolasaunai: sure, will add comment in code.

in the mean time, small explanation here.
We had NaN on all moments ghost nodes. Told you that I'd better have NaNs than "wrong" values there. Since none are used in the numerical algorithms, having NaNs there is/was a good idea because if one algo hit one of those nodes, we'd had seen NaNs in domain... and we'd know immediately something's wrong... 
Turns out the coarsening operator takes fine domain nodes and put values on coarse ones. To fill the coarse node lying aligned with the domain border fine node, one needs, for refinement factor 2, ONE ghost node. That ghost node being NaN it was setting NaNs in coarse domain nodes....

I chose to copy the first domain node value to the first ghost node so that no NaN is interpolated on coarse nodes.

all is explained here https://github.com/PHAREHUB/PHARE/issues/181

PR #239, opened 2020-06-09
load python hierarchy from wrangler

nicolasaunai:
Loading python hierarchies is only feasible from h5 diags. This PR makes it possible from the DataWranler

### DIFF
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -4,6 +4,8 @@ from ..core.box import Box
 import numpy as np
 import os
 from ..core.gridlayout import GridLayout
+from .particles import Particles
+from ..data.wrangler import DataWrangler
 
 
 class PatchData:
@@ -142,8 +144,7 @@ class PatchHierarchy:
     def __str__(self):
         s = "Hierarchy: \n"
         for t, patch_levels in self.time_hier.items():
-            print("t = {}".format(t))
-            for ilvl, lvl in enumerate(patch_levels):
+            for ilvl, lvl in patch_levels.items():
                 s = s + "Level {}\n".format(ilvl)
                 for ip, patch in enumerate(lvl.patches):
                     for qty_name, pd in patch.patch_datas.items():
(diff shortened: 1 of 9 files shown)


pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -439,5 +421,62 @@ def hierarchy_from(h5_filename, time=None, hier=None):
...
+    given hierarchy 'hier' will be filled.
+
+    The function fails if the data is already in hierarchy
+    """
+
+    if simulator is not None and h5_filename is not None:
+        raise ValueError("cannot pass both a simulator and a h5 file")
+
+    if h5_filename is not None:
+        return hierarchy_from(h5_filename, time, hier)
```
Dekken, line 471: should this be "hierarchy_fromh5" ?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -399,9 +412,9 @@ def hierarchy_from(h5_filename, time=None, hier=None):
                     add_to_patchdata(hier_patch.patch_datas, h5_patch_grp, basename, layout)
 
             return hier
-                    #hier_patch.patch_datas.update({dataset_name: pdata})
+            # hier_patch.patch_datas.update({dataset_name: pdata})
```
Dekken, line 415: rm

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -173,6 +174,36 @@ def plot(self):
...
+                val = patch.patch_datas[qty].dataset[:]
+                x_val = patch.patch_datas[qty].x
+                ax.plot(x_val, val, label=r"Level {} patch {}".format(il, ip), marker=marker, alpha=alpha, ls=ls)
+                if ylim is not None:
+                    ax.set_ylim(ylim)
+        ax.legend(ncol=4)
+        ax.set_title(title)
+
+        if filename is not None:
+            fig.savefig(filename)
```
Dekken, line 203: should figs be cleared after saving or something?
I think it's a static instance, so they can conflict if done in succession.

nicolasaunai: I'm not following. If done in succession it's the caller's responsibility to change filename accordingly, no ?

Dekken: dunno, hence why I ask

nicolasaunai: ok cool. Since `filename` is an argument I say it's the user responsibility to check it's not dumb to call the function with that name.

PR #240, opened 2020-06-11
Nd array from dekken

rochSmets:


src/core/data/ndarray/ndarray_vector.h
```diff
@@ -5,275 +5,128 @@
...
-            , nx_{nx}
-            , ny_{ny}
-            , nz_{nz}
+        template<typename... Indexes>
+        DataType const& operator()(Indexes... indexes) const
         {
-        }
+            auto params = std::tuple<Indexes...>{indexes...};
+            static_assert(sizeof...(Indexes) == dim);
+            // static_assert((... && std::is_unsigned_v<decltype(indexes)>)); TODO : manage later if
```
Dekken, line 84: perhaps a >=0 check could suffice?

rochSmets: I agree. I tried something like
`static_assert((... && std::greater_equal<int32_t>()), 0);`
which does not compile, but I can hardly figure out the way to write it...

Dekken: I don't think it can be a compile time check, as it's a runtime value check.

```
 static_assert((... && !std::is_floating_point_v<Indexes>));
 assert ((... && (indexes >= 0)));
```

PR #243, opened 2020-06-17
close h5 file handles on dump, to not need to kill the diag manager

Dekken:


pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -143,7 +143,8 @@ def __str__(self):
         s = "Hierarchy: \n"
         for t, patch_levels in self.time_hier.items():
             print("t = {}".format(t))
-            for ilvl, lvl in enumerate(patch_levels):
+            for ilvl, lvl in patch_levels.items():
+                print("ilvl", ilvl, "lvl", lvl)
```
nicolasaunai, line 147: do we want to print here? the function puts everything in s

Dekken: sure, __str__ would return a string rather than print the string wouldn't it

PR #246, opened 2020-06-22
test patch ghost particles are clones of domain particles

nicolasaunai:
image

 - particles in a patch ghost area should be clones of domain particles in overlaped domain cells of neighbor patches


nicolasaunai:
on top of #239

PR #250, opened 2020-06-23
Python Simulator class

Dekken:
Wrapping C++ Simulator, no more static samrai lifecycle


tests/simulator/diagnostics.py
```diff
@@ -5,15 +5,15 @@
 
 from pybindlibs import cpp
```
nicolasaunai, line 6: do we need to keep that one?

Issue #249, opened 2020-06-23
fix periodic warning 


nicolasaunai:
Warning  : 

```
Warning: Simulation is periodic but some functions are not
```
is raised even if equal. Needs to check double_eq.

Dekken:
we already do this

I believe the issue is the results of the init functions at each border
ie. x = 0 / x = domain.size() - 1

PR #253, opened 2020-06-29
periodic border init values

Dekken:
https://github.com/PHAREHUB/PHARE/issues/249


Dekken:
perhaps a global "Werror" for phare might be useful, to consider all warnings as errors for sort of a "strict" mode

PR #261, opened 2020-07-02
isolate h5 file creation for disparate patches across MPI

Dekken:


src/diagnostic/detail/types/fluid.h
```diff
@@ -50,6 +55,31 @@ class FluidDiagnosticWriter : public H5TypeWriter<HighFiveDiagnostic>
...
+
+template<typename HighFiveDiagnostic>
+void FluidDiagnosticWriter<HighFiveDiagnostic>::createFiles(DiagnosticProperties& diagnostic)
+{
+    auto& hi5 = this->hi5_;
+
+    auto checkCreate = [&](auto& tree, auto var) {
+        auto active = tree + var;
+        bool b      = diagnostic.quantity == active;
+        if (b && !fileData.count(diagnostic.quantity))
```
nicolasaunai, line 67: we can now use `and` instead of `&&` I find it easier to read

src/diagnostic/detail/types/fluid.h
```diff
@@ -50,6 +55,31 @@ class FluidDiagnosticWriter : public H5TypeWriter<HighFiveDiagnostic>
...
+            fileData.emplace(diagnostic.quantity, hi5.makeFile(diagnostic));
+    };
+
+    for (auto& pop : hi5.modelView().getIons())
```
nicolasaunai, line 71: should be `auto const& pop`?

src/diagnostic/detail/types/fluid.h
```diff
@@ -60,13 +90,7 @@ void FluidDiagnosticWriter<HighFiveDiagnostic>::getDataSetInfo(DiagnosticPropert
...
     std::string lvlPatchID{std::to_string(iLevel) + "_" + patchID};
 
-    auto checkActive = [&](auto& tree, auto var) {
-        auto active = tree + var;
-        bool b      = diagnostic.quantity == active;
-        if (b && !fileData.count(diagnostic.quantity))
-            fileData.emplace(diagnostic.quantity, hi5.makeFile(diagnostic));
-        return b;
-    };
+    auto checkActive = [&](auto& tree, auto var) { return diagnostic.quantity == tree + var; };
```
nicolasaunai, line 93: shouldn't we generally capture only the one variable that's used in the lambda ? less source of (w)error (with shadowing etc.)

src/diagnostic/detail/types/particle.h
```diff
@@ -62,6 +67,26 @@ class ParticlesDiagnosticWriter : public H5TypeWriter<HighFiveDiagnostic>
 };
 
 
+template<typename HighFiveDiagnostic>
+void ParticlesDiagnosticWriter<HighFiveDiagnostic>::createFiles(DiagnosticProperties& diagnostic)
+{
+    auto& hi5 = this->hi5_;
+
+    auto checkCreate = [&](auto& tree, auto var) {
```
nicolasaunai, line 75: should that lambda somehow be factored out (same as in fluid?)... no pressure... maybe more readable as is than isolated somewhere else?

src/diagnostic/detail/types/particle.h
```diff
@@ -62,6 +67,26 @@ class ParticlesDiagnosticWriter : public H5TypeWriter<HighFiveDiagnostic>
...
+        bool b = diagnostic.quantity == tree + var;
+        if (b && !fileData.count(diagnostic.quantity))
+            fileData.emplace(diagnostic.quantity, hi5.makeFile(diagnostic));
+    };
+
+    for (auto& pop : hi5.modelView().getIons())
```
nicolasaunai, line 81: const& ?

tests/simulator/diagnostics.py
```diff
@@ -20,16 +21,25 @@ def dup(dic):
...
+    )
+    def test_dump_diags_with_killing_dman_1d(self, input):
 
         for interp in range(1, 4):
 
             if rank == 0 and os.path.exists(out):
                 shutil.rmtree(out)
 
-            self.dman, self.sim, self.hier = create_simulator(1, interp, **dup({}))
+            self.dman, self.sim, self.hier = create_simulator(1, interp, **input)
```
nicolasaunai, line 42: we should probably test that all diags we require are indeed created and not only the EM file. That'd also be useful to have the list of names for diag files somewhere !

src/diagnostic/detail/types/fluid.h
Dekken: > shouldn't we...

I don't see much benefit to this. I like [&] because it's minimalist and you get what you need.
If you want to choose an approach, maybe a style guide should exist

src/diagnostic/detail/types/particle.h
Dekken: not sure I get this

tests/simulator/diagnostics.py
Dekken: the filename could be provided by the python diagnostic object, which should be accessible here.
just loop them and check the files exist

src/diagnostic/detail/types/fluid.h
Dekken: k

src/diagnostic/detail/types/particle.h
nicolasaunai: sorry :D 
I meant it looks like this lambda is the same as for the fluid create function, hence maybe could be a function used in both cases rather than duplicating.

Dekken: ok can do

src/diagnostic/detail/types/fluid.h
nicolasaunai: ok I didn't find anything in the cpp core guidelines regarding how to capture things in lambdas.
My argument is that explicit is often better than implicit and when there is only one variable to capture, its explicit that it is *that one* that we want and no other.
But it's a discussion...

tests/simulator/diagnostics.py
Dekken: this fails on particle ghost files, but this may be because of hierarchy from h5 reasons

PR #263, opened 2020-07-08
Block test-updater cmake when building with MPI - 

Dekken:
tries to compile otherwise and fails

this seems to have come about because we test without MPI before with MPI
so the first time it builds test-updater, and compiles properly, but the second time it isn't built as its' already built.


Dekken:
what was happening was test-updater was still being built, but it was never receiving the correct flags, like for "-DPHARE_HAS_HIGHFIVE=1" specifcally, so failed on an #ifndef

Dekken:
it might be simpler just to export "-DPHARE_HAS_HIGHFIVE=1" via CMAKE_CXX_FLAGS

Issue #256, opened 2020-06-30
SAMRAI warning about regrid_interval


nicolasaunai:
We get this warning, probably because we give this superfluous param. 

49: ....SAMRAI/source/SAMRAI/algs/TimeRefinementIntegrator.C line :1621 message: TimeRefinementIntegrator::getFromInput() warning...
49: regrid_interval input parameter not applicable with
49: refined timestepping and will be ignored.

PR #267, opened 2020-07-09
PHARE integrator, drop regrid_interval

Dekken:
https://github.com/PHAREHUB/PHARE/issues/256

image

### DIFF
--- a/src/amr/wrappers/integrator.h
+++ b/src/amr/wrappers/integrator.h
@@ -144,7 +144,6 @@ namespace amr
             db->putDouble("start_time", startTime);
             db->putDouble("end_time", endTime);
             db->putInteger("max_integrator_steps", 1000000);
-            db->putInteger("regrid_interval", 1);
 
 
             timeRefIntegrator_ = std::make_shared<SAMRAI::algs::TimeRefinementIntegrator>(


Dekken:
SAMRAI doc references of "tag_buffer" 
image

Dekken:
tag_buffer has it's own warning if it is not set, I what I believe I have done is just set what it would be set to when the warning is hit

Dekken:
example https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_BuildGithubPrClang/61049?buildTab=log&focusLine=5662&linesState=4995

nicolasaunai:
Ok maybe what you have done kills the warning but I do not understand why there would need to use tag_buffer (even with default value) if refinement boxes are given. Tag buffer just represent the minimum number of tagged cells to be reached before clustering them into refinement boxes... to me that only makes sense when tagging cells, i.e. not when using predefined user boxes....

nicolasaunai:
ok header says : 

```
    * The tag buffer indicates the number of cells on each level by which
    * tagged cells will be buffered after they have selected for refinement.
    * These values are passed into the gridding algorithm routines during
    * hierarchy construction and regridding.  The tag buffer helps to
    * guarantee that refined cells near important features in the solution
    * will remain refined until the level is regridded next.
    *
    * Important note: these values may be specified in the input file.
    * If not, default values are set based on the regrid intervals.
    * However, if the user attempts to specify these values, care must
    * be taken to assure that improper tag buffering will not degrade the
    * calculation.
```

so my understanding is : 

- this keyword does not make any sense when user refinement_boxes are provided, since there is no tagging and regrinding in this case.
- we should therefore not specify this keyword if using refinement_boxes. And IMO SAMRAI should not print any warning. We should see with them.
- if not using refinement_boxes AND not tagging either (forthcoming PR), PHARE should run with only 1 root level, and the keyword should not be used either and no warning issued.
- if running with no refinement_boxes but WITH tagging (forthcoming PR), PHARE should in my opinion NOT (for the moment at least) give samrai the tag_buffer keyword but ask the user to provide the regrid_interval value and SAMRAI will default tag_buffer to regrid_interval.

PR #273, opened 2020-07-10
cppcheck  warnings

Dekken:
might want to start looking at cppcheck again

### DIFF
--- a/src/amr/data/field/field_data_factory.h
+++ b/src/amr/data/field/field_data_factory.h
@@ -171,8 +171,8 @@ namespace amr
 
 
     private:
-        bool const fineBoundaryRepresentsVariable_;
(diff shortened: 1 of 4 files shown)


src/amr/data/field/field_data_factory.h
```diff
@@ -171,8 +171,8 @@ namespace amr
 
 
     private:
-        bool const fineBoundaryRepresentsVariable_;
-        bool const dataLivesOnPatchBorder_;
+        bool const fineBoundaryRepresentsVariable_ = 0;
```
nicolasaunai, line 174: `false` reads better for bool

Dekken: done

Issue #270, opened 2020-07-10
Add wiki page for build config


nicolasaunai:
Add a wiki page under development section to quickly explain the different aspects of the build configuration. For instance, MPI vs non-MPI, choices we have made regarding how dependencies are dealt with (HDF5, MPI, SAMRAI, pybind, ....), etc. known issues/difficulties/ possible future ways of improvements.

PR #274, opened 2020-07-10
cmake refactor/document init

Dekken:
suggestion in tandem to https://github.com/PHAREHUB/PHARE/issues/270


res/cmake/def.cmake
```diff
@@ -1,116 +1,42 @@
...
 
 set (PHARE_HAS_HIGHFIVE "0")
 if (HighFive)
  set (PHARE_HAS_HIGHFIVE "1")
 endif()
 
-# Pybind errors with clang, it is default in GCC
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-  set (CMAKE_CXX_FLAGS -fsized-deallocation)
+if (NOT DEFINED PHARE_MPI_PROCS)
```
nicolasaunai, line 12: shouldn' that rather be in the test config file?

res/cmake/def.cmake
```diff
@@ -1,116 +1,42 @@
...
+  set(PHARE_MPI_PROCS 1)
+  if(testMPI)
+    set(PHARE_MPI_PROCS 2)
+  endif()
 endif()
 
-if(coverage AND NOT MSVC)
-  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pg -DHAVE_EXECINFO_H -g3")
+# Pybind errors with clang, it is default in GCC
```
nicolasaunai, line 19: if that's for pybind shouldn't that rather be in the pybind config file?

res/cmake/func.cmake
```diff
@@ -0,0 +1,72 @@
+
```
nicolasaunai, line 1: does it make sense actually to have a "function" file where all functions are ? wouldn't it be easier to find these functions with the config they are associated with ? like those would go in test cmake ?

res/cmake/coverage.cmake
```diff
@@ -1,38 +1,38 @@
...
-      endif(APPLE)
-
-      add_custom_target(show_coverage
-          COMMAND ${OPPEN_CMD} ${CMAKE_CURRENT_BINARY_DIR}/coverage/index.html
-          DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/coverage/index.html gcovr
-          )
-  ENDIF(coverage)
-
-ENDIF(test)
+if (test AND coverage)
```
nicolasaunai, line 2: we need more comments for the "whys" we do each of these nasty configs....

src/core/numerics/ohm/ohm.h
```diff
@@ -183,8 +183,8 @@ namespace core
 
 
         template<typename Field, typename ComponentTag>
-        auto pressure_(Field const& n, Field const& Pe, MeshIndex<Field::dimension> index,
-                       ComponentTag) const
+        auto pressure_(Field const& n, Field const& Pe,
+                       [[maybe_unused]] MeshIndex<Field::dimension> index, ComponentTag) const
```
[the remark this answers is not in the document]
Dekken: pretty sure it's werror
not sure in what case this function is called and it's never used, that could be a bug

res/cmake/def.cmake
Dekken: it's something used by testing, but it's it only used by testing?
currently maybe, so maybe yes it should be

res/cmake/func.cmake
Dekken: if they all have a file they belong to it makes sense
is there any orphans?

res/cmake/coverage.cmake
Dekken: yes, I will probably do what I did for werror and split it up

src/core/numerics/ohm/ohm.h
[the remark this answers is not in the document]
Dekken: ignore it, probably shouldn't be in this PR

res/cmake/def.cmake
Dekken: I'm not sure it's smart to separate setting variables like CMAKE_CXX_FLAGS all that much

Dekken: perhaps it should be "PHARE_FLAGS" instead of CMAKE_CXX_FLAGS

Dekken:
not sure if the symlink handling should be done in CMake but I think it might better than accidental infinite loops.

it does mean that for #309 it would need a "git reset -hard" before "git pull" I think

res/cmake/coverage.cmake
Dekken: I removed a lot of flags that were (I believe) unnecessary for the point of enabling coverage
so it should be a bit simpler now

PR #251, opened 2020-06-24
Ghosttest

nicolasaunai:
#246 on top of #250

PR #277, opened 2020-07-12
diagnostics string csv attributes refactored to -> vector of type 

Dekken:
dropped AMR periodicity test, deprecated and replaced by python tests

should make 2D+ pyphare/pharesee simpler


Dekken:
not to be merged before https://github.com/PHAREHUB/PHARE/pull/251
should make 2D+ pyphare/pharesee simpler

Dekken:
probably makes casts like these redundant
https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharesee/hierarchy.py#L261

src/core/utilities/mpi_utils.h
```diff
@@ -23,36 +23,58 @@ std::vector<Data> collect(Data const& data, int mpi_size = 0);
...
                                         std::string null_str = "null");
 
 
 template<typename Data>
-std::vector<std::vector<Data>> collectVector(std::vector<Data> const& data, int mpi_size)
-{
+std::vector<std::vector<Data>> collectVector(std::vector<Data> const& data, int mpi_size,
+                                             bool empty = false)
+{ // can't send empty vectors
+    if (data.size() == 0)
```
nicolasaunai, line 63: I think you may want ```MPI_Allgatherv``` instead

https://docs.microsoft.com/en-us/message-passing-interface/mpi-allgatherv-function

Dekken:
> ok looks like a good idea.
> I think investigating if the MPI_Allgatherv function can be used is a good idea. Looks like we can gather vectors of variable **and possibly zero** length

will test it and see

src/core/utilities/mpi_utils.h
Dekken: done

src/core/utilities/mpi_utils.h
```diff
@@ -23,43 +23,101 @@ std::vector<Data> collect(Data const& data, int mpi_size = 0);
 
 size_t max(size_t local, int mpi_size = 0);
 
-template<typename Data>
-void _collect(Data const* const data, std::vector<Data>& values, size_t send = 1,
-              size_t receive = 1)
+int size();
+
+template<typename Data, typename GatherFunc>
+void _gather(GatherFunc&& gather)
```
nicolasaunai, line 29: should probably be 
```
void _gather(GatherFunc const&& gather)
```

Issue #204, opened 2020-05-04
Python tests dictionary duplicate keys


Dekken:
It's possibly not unhelpful to use a dictionary that errors if the same key is attempted to added twice
it would stop silent dict replacements

Dekken:
https://github.com/PHAREHUB/PHARE/pull/284

PR #284, opened 2020-07-15
python tests/ddt - add dict which blocks key reassignment 

Dekken:
prevent silent/hidden overwrites

https://github.com/PHAREHUB/PHARE/issues/204

### DIFF
--- a/tests/simulator/__init__.py
+++ b/tests/simulator/__init__.py
@@ -3,6 +3,18 @@
 import pyphare.pharein as ph, numpy as np
 from pyphare.pharein import ElectronModel
 
+# Block accidental dictionary key rewrites
+class NoOverwriteDict(dict):
+    def __init__(self, dict):
+        for k, v in dict.items():
+            self[k] = v
+
+    def __setitem__(self, k, v):
+        if k in self.keys():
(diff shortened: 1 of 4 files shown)


Dekken:
> * are these changes fixing a bug we have ?
more like bug prevention

> * should that NoOverwriteDict be used also in other places like pharein Simulation.py or maxwellianFluidModel?
can you show an example?

nicolasaunai:
> > * should that NoOverwriteDict be used also in other places like pharein Simulation.py or maxwellianFluidModel?
> 
> can you show an example?

note sure we want this but, Simulation makes a heavy usage of dicts where it seems important not to erase params: 
https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharein/simulation.py#L314

MaxwellianFluidModel also uses user parameters in dicts  : 
https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharein/maxwellian_fluid_model.py#L41-L51
https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharein/maxwellian_fluid_model.py#L117-L134

nicolasaunai (approved):

Issue #299, opened 2020-07-24
IWYU


PhilipDeegan:
there's some strangeness

it's suggesting "iosfwd" for size_t sometimes

see: https://gist.github.com/Dekken/4dc4e65c11681e67a67b5520ee6dd276

Dekken:
IWYU considers ".h" files as C by default, and ".hpp" as C++ by default.

PR #300, opened 2020-07-25
split_1/2d header rearrangement, a few std:: straglers

Dekken:
rearrange split headers to make each file includable individually

moved contents of split.h to splitter.h
now just include each dim split file in split.h


src/amr/data/particles/refine/split_1d.h
```diff
@@ -1,3 +1,4 @@
+// IWYU pragma: private, include "amr/data/particles/refine/split.h"
```
nicolasaunai, line 1: what's that

src/amr/data/particles/refine/split_1d.h
```diff
@@ -6,10 +7,16 @@ Splitting reference material can be found @
 #ifndef PHARE_SPLIT_1D_H
 #define PHARE_SPLIT_1D_H
 
+#include <array>                        // for array, array<>::value_type
```
nicolasaunai, line 10: do we need those "for..xxx" comments? (I get they are automatic but...)

Dekken: need? no, I think it's useful to see at the top of the file tho what is used in the file

src/amr/data/particles/refine/split_1d.h
Dekken: possibly overkill
what this does is, if you want a class that's in split_1d.h, IWYU will tell you that you need "split.h" instead, which I would think is the general case. that you want all dims, not just 1d.

src/amr/data/particles/refine/split_1d.h
nicolasaunai: depends how much stuff we use from a header, I wouldn't want a long list of stuff...
did IWYU do write that in the files? Did you run IWYU manually ? if yes, this kind of comments will never appear again ? Is that really consistent that we let them here and they don't appear anywhere else?

src/amr/data/particles/refine/split_1d.h
nicolasaunai: ok. I'm just concerned about how consistent with the rest of the code base this is. My understanding is that you ran manually IWYU to get that, so this is something that won't be consistently done. So is it something that we want here?

nicolasaunai:
conflicts

src/amr/data/particles/refine/split_1d.h
Dekken: if somethings are considered "private" includes then it might be conceivable to put it in other places.

src/amr/data/particles/refine/split_1d.h
Dekken: it's pasted from https://github.com/PHAREHUB/PHARE/issues/299 

it's easily removable

PR #307, opened 2020-07-28
Run simulator tests that don't require h5 when h5 is not available

Dekken:
This might be why LGTM doesn't have more files showing


Dekken:
I have had issues with the project not updating on LGTM before.
We should have PRs working with this, but might have to wait for updates on the stats

https://discuss.lgtm.com/t/retriggering-a-succesful-build/2921/2

nicolasaunai (approved):

PR #310, opened 2020-07-30
PyPhare 2d box

Dekken:
Hierarchy/geometry is left as unchanged as possible, deprecating some box code in the process.
test_initializtion.test updates to follow with hierarchy/geometry changes inline with nd box

### DIFF
--- a/pyphare/pyphare/core/box.py
+++ b/pyphare/pyphare/core/box.py
@@ -1,4 +1,20 @@
+import numpy as np
 
+
+def is_scalar(arg):
+    return not isinstance(arg, list) and not is_nd_array(arg)
+
+
+def is_nd_array(arg):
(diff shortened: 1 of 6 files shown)


pyphare/pyphare/core/box.py
```diff
@@ -18,49 +36,85 @@ def __mul__(self, box2):
...
+        # deprecated in favor of self.shape()
+        return self.shape()
 
+    def xyz(self):
+        """
+        convert lower/upper to [[x0, x1], [y0, y1], [z0, z1]]
+        """
+        return [[l, u] for l, u in zip(self.lower, self.upper)]
 
+    def corner_points(self):
```
Dekken, line 61: unused, but possibly useful, will remove if suggested

nicolasaunai (approved):

PR #311, opened 2020-07-30
phare include refactor / RuntimeDiagnosticInterface renamed SimulatorDiagnostics

Dekken:
src/phare/include.h -> src/phare/phare.h

RuntimeDiagnosticInterface moved to simulator.h and renamed SimulatorDiagnostics


Dekken:
IWYU wasn't run/done because src/phare/phare.h is something of a convenience header and other things use it

can make a specific file for SimulatorDiagnostics if suggested

nicolasaunai (approved):

Issue #301, opened 2020-07-26
LGTM doesn't include diagnostics


PhilipDeegan:
Diagnostics tests were failing for what seemed to be MPI related problems, but it being a third party system without remote access isn't all that simple to debug

Dekken:
"-DHighFive=OFF"

https://github.com/PHAREHUB/PHARE/blob/master/lgtm.yml#L30

Dekken:
error 

```
[2020-11-13 13:00:06] [build] [----------] 4 tests from Simulator1dTest/0, where TypeParam = SimulatorTestParam<1ul, 1ul, 2ul>
[2020-11-13 13:00:06] [build] [ RUN      ] Simulator1dTest/0.fluid
[2020-11-13 13:00:06] [build] tagging method is set to REFINE_BOXES
[2020-11-13 13:00:06] [build] initializing hierarchy...init level 0 with regriding = 0
[2020-11-13 13:00:06] [build] init level 1 with regriding = 0
[2020-11-13 13:00:06] [build] done !
[2020-11-13 13:00:06] [build] HDF5-DIAG: Error detected in HDF5 (1.10.4) MPI-process 0:
[2020-11-13 13:00:06] [build]   #000: ../../../src/H5FDmpio.c line 317 in H5Pset_fapl_mpio(): not a file access list
[2020-11-13 13:00:06] [build]     major: Property lists
[2020-11-13 13:00:06] [build]     minor: Inappropriate type
[2020-11-13 13:00:06] [build]   #001: ../../../src/H5Pint.c line 3822 in H5P_object_verify(): property list is not a member of the class
[2020-11-13 13:00:06] [build]     major: Property lists
[2020-11-13 13:00:06] [build]     minor: Unable to register new atom
[2020-11-13 13:00:06] [build]   #002: ../../../src/H5Pint.c line 3772 in H5P_isa_class(): not a property list
[2020-11-13 13:00:06] [build]     major: Invalid arguments to routine
[2020-11-13 13:00:06] [build]     minor: Inappropriate type
[2020-11-13 13:00:06] [build] unknown file: Failure
[2020-11-13 13:00:06] [build] C++ exception with description "Unable to set-up MPIO Driver configuration " thrown in the test body.
```

Dekken:
a bit of googling suggests it's possibly because of an old version of OpenMPI

PhilipDeegan:
LGTM doesn't exist anymore, it was bought/incorporated into github
[...]

PR #312, opened 2020-07-30
LGTM - Exclude subprojects as library

Dekken:
LGTM needs two building commits so this should fix that also
should see the LGTM C++ PR job here too

maybe wait for: https://github.com/github/codeql/issues/3997

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,12 +1,8 @@
 path_classifiers:
-  tag:
-    exclude: subprojects
-  docs:
-  - README.md
-  - LICENSE.txt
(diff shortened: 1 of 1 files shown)


Dekken:
not sure if it matters but this format here might be better https://github.com/Dekken/FUR/blob/master/lgtm.yml

Dekken:
Looks like all files (excluding diagnostics) are found now

to follow up https://github.com/PHAREHUB/PHARE/issues/301

nicolasaunai:
should we merge that now?

Dekken:
if it "LGTY" go for it

Issue #321, opened 2020-09-07
cmake - IPO/LTO on by default if supported can fail


Dekken:
Just hit this on Juno

PR #322, opened 2020-09-07
add cmake option 'withIPO' default off

Dekken:
default on can cause failures at link time

https://github.com/PHAREHUB/PHARE/issues/321


Dekken:
maybe should add a message saying if IPO is available aswell as if it's enabled

nicolasaunai (approved):

PR #325, opened 2020-09-07
nd fixMomentGhosts

Dekken:


Dekken:
review python initialization tests

src/core/data/ions/ions.h
```diff
@@ -234,62 +234,253 @@ namespace core
...
+        auto ix0 = layout.physicalStartIndex(QtyCentering::primal, Direction::X);
+        auto ix1 = layout.physicalEndIndex(QtyCentering::primal, Direction::X);
+        auto ix2 = layout.ghostEndIndex(QtyCentering::primal, Direction::X);
+
+        auto iy0 = layout.physicalStartIndex(QtyCentering::primal, Direction::Y);
+        auto iy1 = layout.physicalEndIndex(QtyCentering::primal, Direction::Y);
+        auto iy2 = layout.ghostEndIndex(QtyCentering::primal, Direction::Y);
+
+        auto copy = [&](auto& field) {
+            // topLeft
```
nicolasaunai, line 340: convention is that 0 left/bottom and 1 is right/top

nicolasaunai (commented):
I think it's ok but I would feel more comfortable with a quick and dirty unit test for the boxsetter in each dim to make sure there is no (and there won't be by msitake) any error in all these indexes.... what do you think?

src/core/data/ions/ions.h
Dekken: ok I wasn't sure

Dekken:
> I think it's ok but I would feel more comfortable with a quick and dirty unit test for the boxsetter in each dim to make sure there is no (and there won't be by msitake) any error in all these indexes.... what do you think?

could be added to the diagnostic init value tests perhaps

Dekken:
todo 3d tests

Dekken:
the hollowbox stuff uses NdArrayViews, so it's just embded in the same file for the moment.
maybe move to src/core/utilities/box/box.h
or possibly src/core/utilities/box/hollow_box.h as this needs to include ndarray and box.h doesn't

Dekken:
fixMomentGhosts wasn't updated for 1d to use the hollowbox stuff, nor is it tested for 1d, so TODO (in this PR)

src/core/utilities/box/hollow_box.h
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+private:
+    std::array<Size, dim> sizes_;
+    std::array<Size, dim> buffer_;
+};
+
+template<typename GridLayout>
+auto totalCells(GridLayout const& layout, QtyCentering centering)
```
nicolasaunai, line 85: i don't think this function is named after what it does.
Also, you may want to look at https://github.com/PHAREHUB/PHARE/blob/7d11552dc86f6a8fbb963146eef498583658801a/src/core/data/grid/gridlayout.h#L850

src/core/utilities/box/hollow_box.h
```diff
@@ -0,0 +1,225 @@
...
+    void setBuffer(T const* data, V val)
+    {
+        core::NdArrayView<dim, T> view{data, sizes_};
+
+        if constexpr (dim == 1)
+        {
+            assert(4 == buffer_[0]);
+            for (std::size_t i = 0; i < buffer_[0]; i++)
+                view(i) = val;
+            assert(4 == sizes_[0] - (sizes_[0] - buffer_[0]));
```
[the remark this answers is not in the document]
Dekken: ignore these, forgot to remove them

src/core/utilities/box/hollow_box.h
```diff
@@ -0,0 +1,225 @@
...
+
+        if constexpr (dim == 2)
+            return (sizes_[1] * ((buffer_[0]) * 2)) + (sizes_[0] * ((buffer_[1]) * 2))
+                   - ((buffer_[0]) * (buffer_[1]) * 4);
+
+        throw std::runtime_error("not implemented");
+    }
+
+    template<typename T, typename V>
+    void setBuffer(T const* data, V val)
```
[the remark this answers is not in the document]
Dekken: setBuffer could be setData
buffer_ is the size from the edge

src/core/utilities/box/hollow_box.h
```diff
@@ -0,0 +1,225 @@
...
+            for (std::size_t i = sizes_[0] - buffer_[0]; i < sizes_[0]; i++)
+                view(i) = val;
+        }
+        else if constexpr (dim == 2)
+        {
+            for (std::size_t i = 0; i < buffer_[0]; i++) // bottom
+                for (std::size_t j = 0; j < sizes_[1]; j++)
+                    view(i, j) = val;
+
+            for (std::size_t i = sizes_[0] - buffer_[0]; i < sizes_[0]; i++) // top
```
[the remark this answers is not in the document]
Dekken: typically yes, but in nDarray vector indices aren't typical

src/core/utilities/box/hollow_box.h
```diff
@@ -0,0 +1,225 @@
...
+    {
+    }
+
+    std::size_t cells() const
+    {
+        if constexpr (dim == 1)
+            return ((buffer_[0]) * 2);
+
+        if constexpr (dim == 2)
+            return (sizes_[1] * ((buffer_[0]) * 2)) + (sizes_[0] * ((buffer_[1]) * 2))
```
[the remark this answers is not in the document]
Dekken: done

src/core/utilities/box/hollow_box.h
Dekken: done

src/core/utilities/box/hollow_box.h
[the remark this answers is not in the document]
Dekken: updated

src/core/data/ions/ions.h
```diff
@@ -244,52 +245,28 @@ namespace core
...
-        }
-        else if constexpr (Ions::dimension == 3)
-        {
-            std::cout << "test 3\n";
-        }
-    }
+                auto& flux_comp = pop.flux().getComponent(type);
+
+                auto nGhosts
+                    = layout.physicalStartIndex(flux_comp.physicalQuantity(), Direction::X);
```
nicolasaunai, line 257: ok so the code says Direction::X but it is really valid for all directions since moments are primal primal primal and thus have the same number of ghosts in all directions.
But that is not obvious for a new reader or maybe us in some time.
I therefore would recommend adding a comment explaining that or changing the code so that Direction::X disappears

src/core/data/ions/ions.h
```diff
@@ -244,52 +245,28 @@ namespace core
...
 
+                flux_comp[Mask{0, nGhosts - 1}] = NAN;
```
nicolasaunai, line 259: where is NAN defined?

src/core/data/ions/ions.h
```diff
@@ -244,52 +245,28 @@ namespace core
...
+                flux_comp[Mask{nGhosts}] >> flux_comp[Mask{nGhosts - 1}];
```
nicolasaunai, line 260: I would rather have physicalStartIndex (or phyStartIdx ?) rather than nbrGhost there.
I know it is the same value, but nbrGhost is a number of nodes (homogeneous to a distance) and physicalStartIndex is a node index (homogeneous to a position)... and Mask takes positions not distances.
Same remark for the nanification.
Reads better..

src/core/data/ndarray/ndarray_vector.h
```diff
@@ -64,10 +64,71 @@ struct NdArrayViewer
 };
 
 
+
+template<typename Array, typename Mask>
+class MaskedView
+{
+public:
+    static auto constexpr dimension = Array::dimension;
```
nicolasaunai, line 72: we should probably static_assert mask and array have compatible dimensions ?

src/core/data/ndarray/ndarray_vector.h
```diff
@@ -193,15 +261,175 @@ class NdArrayVector
...
+
+
+
+
+template<typename Array, typename Mask>
+void operator>>(MaskedView<Array, Mask>&& inner, MaskedView<Array, Mask>&& outer)
+{
+    using MasekdView_t = MaskedView<Array, Mask>;
+
+    if constexpr (MasekdView_t::dimension == 1)
```
nicolasaunai, line 370: Masked not Masekd (my bad)
same in the second and third constexpr and above using

tests/core/utilities/box/test_main.cpp
```diff
@@ -99,6 +99,76 @@ TEST(PointInBox, returnFalseIfIsNoBox)
     EXPECT_FALSE(isIn(point, boxes));
 }
 
+
```
nicolasaunai, line 102: transform into the Mask nD test?

tests/diagnostic/test_diagnostics.h
```diff
@@ -35,9 +36,73 @@ std::array<std::uint32_t, GridLayout::dimension> fieldIndices(FieldFilter ff, Fu
         return {get(core::Direction::X), get(core::Direction::Y), get(core::Direction::Z)};
 }
 
+
+template<std::size_t dim, typename T, typename GridLayout, typename Field>
+void validateFluidGhosts(std::vector<T> const& data, GridLayout const& layout, Field const& field)
+{
+    using Filter = FieldDomainPlusNFilter;
```
nicolasaunai, line 43: naive question : could that be now done with a MaskedView?

tests/diagnostic/test_diagnostics.h
```diff
@@ -35,9 +36,73 @@ std::array<std::uint32_t, GridLayout::dimension> fieldIndices(FieldFilter ff, Fu
...
+        ASSERT_TRUE(beg[d] > 0);
+        ASSERT_TRUE(end[d] > beg[d]);
+        ASSERT_TRUE(siz[d] > end[d]);
+    }
+
+    core::NdArrayView<dim, float> view{data, siz};
+
+    if constexpr (dim == 1)
+    {
+        std::size_t expected_nans = ((siz[0] - end[0]) * 2) - 2;
```
nicolasaunai, line 65: unused?
test should assert first nbrGhost-1 ghost nodes are indeed nans
Also : assumes nbrGhost == 5, which may not be true in future...

src/core/utilities/box/hollow_box.h
```diff
@@ -0,0 +1,209 @@
...
+template<typename GridLayout, typename Size = std::uint32_t>
+class GriddedHollowBox
+{
+    static constexpr auto dim = GridLayout::dimension;
+
+    template<typename HybridQuantity>
+    static auto ghostArray(GridLayout const& layout, HybridQuantity quantity)
+    {
+        auto ix0 = layout.physicalStartIndex(quantity, Direction::X);
+        return ConstArray<Size, dim>(ix0 - 1);
```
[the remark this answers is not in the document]
Dekken: this file doesn't even exist anymore

src/core/data/ndarray/ndarray_vector.h
Dekken: good catch!

tests/diagnostic/test_diagnostics.h
Dekken: the test needs updating like it was with the hollow_box so, maybe the Mask stuff can return the total number of cells it covers

not sure where the assumption of 5 is here

tests/diagnostic/test_diagnostics.h
Dekken: possibly

tests/core/utilities/box/test_main.cpp
Dekken: in the nDarray test file sure

tests/diagnostic/test_diagnostics.h
nicolasaunai: I misunderstood there is no further assumption that "we nan all but 1 ghost". that is fine.

src/core/data/ndarray/ndarray_vector.h
Dekken: I'm guessing there'd be a compile error, can confirm

tests/diagnostic/test_diagnostics.h
Dekken: can we consider this a separate issue?

tests/diagnostic/test_diagnostics.h
Dekken: tests added, unused variable removed.

tests/core/utilities/box/test_main.cpp
Dekken: done

src/core/data/ndarray/ndarray_vector.h
Dekken: MaskedView takes a mask as template, so the dim is implicit

src/core/data/ions/ions.h
Dekken: comment https://github.com/PHAREHUB/PHARE/pull/325/files#diff-3c9be39f547148798f7a961790c9d328R246

nicolasaunai (approved):

Issue #332, opened 2020-09-16
Deadlock in diagnostics if process lacks patch on lower level


Dekken:

Dekken:
https://github.com/PHAREHUB/PHARE/pull/337/commits/b3546e5bc154e7ab750d462c045855412459a306

Issue #331, opened 2020-09-16
Possible nullptr if coarser level has no patch


Dekken:
contrived test case where min/max patch size = domain size, with >= 2 mpi_cores 

https://github.com/PHAREHUB/PHARE/blob/bf5212d4ce7d3568d0165e720fe58b6d1281b86a/src/amr/messengers/hybrid_hybrid_messenger_strategy.h#L384

Dekken:
```assert(*std::begin(*hierarchy->getPatchLevel(levelNumber)));``` fails

nicolasaunai:
the patch here is used to get the times of ion resources, to know what is the "new" coarse time (after coarse push).
There may be another solution to get this time than using a patch, because using a patch assumes the current process indeed has at least one patch on the next coarser.....

nicolasaunai:
fixed by #337

PR #333, opened 2020-09-16
Fix possible deadlock when no coarsest patch is found locally under MPI

Dekken:
- start patch.size() check on "minlevel" in case current process lack lower level patch  
- add more MPI tests with N  = 3/4 for diagnostics

https://github.com/PHAREHUB/PHARE/issues/332

### DIFF
--- a/tests/simulator/CMakeLists.txt
+++ b/tests/simulator/CMakeLists.txt
@@ -10,9 +10,12 @@ endif()
 if(HighFive)
 
   ## These test use dump diagnostics so require HighFive!
-  add_python3_test(diagnostics     diagnostics.py         ${CMAKE_CURRENT_BINARY_DIR})
-  add_python3_test(init-particles  test_initialization.py ${CMAKE_CURRENT_BINARY_DIR})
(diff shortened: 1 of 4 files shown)


tests/simulator/CMakeLists.txt
```diff
@@ -10,8 +10,11 @@ endif()
 if(HighFive)
 
   ## These test use dump diagnostics so require HighFive!
-  add_python3_test(diagnostics     diagnostics.py          ${CMAKE_CURRENT_BINARY_DIR})
-  add_python3_test(init-particles  test_initialization.py ${CMAKE_CURRENT_BINARY_DIR})
+  add_python3_test(      diagnostics diagnostics.py  ${CMAKE_CURRENT_BINARY_DIR}) # serial or n = 2
+  add_mpi_python3_test(3 diagnostics diagnostics.py  ${CMAKE_CURRENT_BINARY_DIR})
```
Dekken, line 14: maybe these should only run if we're not doing a coverage build?

nicolasaunai (approved):

Dekken:
I think we're hitting this https://github.com/PHAREHUB/PHARE/issues/331

Dekken:
included as part of https://github.com/PHAREHUB/PHARE/pull/337

PR #336, opened 2020-09-18
Hierarchy init level ghost test

Dekken:
with WIP additions of nd hierarchy


tests/simulator/test_initialization.py
```diff
@@ -582,5 +585,58 @@ def test_overlapped_particledatas_have_identical_particles(self, interp_order, r
...
+              patchData.dataset.select(patchData.box) for patchData in patchDatas
+            ]) # including patch ghost particles means duplicates
+
+        for ilvl, particle_gaboxes in level_ghost_boxes(datahier).items():
+            assert ilvl > 0 # has no level 0
+
+            for particles_id, gaboxes_list in particle_gaboxes.items():
+                coarse_particles = domainParticles_for(ilvl - 1, particles_id)
+                coarse_split_particles = coarse_particles.split(sim)
+                assert coarse_split_particles.size() == coarse_particles.size() * sim.refined_particle_nbr
```
nicolasaunai, line 611: I think we need to assert sizes are not zero also?

Dekken: ok!

nicolasaunai:
should we close that one since now part of #349 ?

Dekken:
> should we close that one since now part of #349 ?

that's one option, this can be considered in isolation tho, so as you wish

nicolasaunai:
> > should we close that one since now part of #349 ?
> 
> that's one option, this can be considered in isolation tho, so as you wish

Let's do 349 only maybe, I have the impression some stuff here is lagging behind...

Dekken:
#349 might get a bit big

PR #337, opened 2020-09-18
set afterPushCoarseTime_ from coarser newTime

Dekken:
uses no mpi comms

includes #333 to show segfault is fixed

liberties taken to do some east-consting (putting "const" after the type)

### DIFF
--- a/tests/simulator/CMakeLists.txt
+++ b/tests/simulator/CMakeLists.txt
@@ -10,9 +10,12 @@ endif()
 if(HighFive)
 
   ## These test use dump diagnostics so require HighFive!
-  add_python3_test(diagnostics     diagnostics.py         ${CMAKE_CURRENT_BINARY_DIR})
-  add_python3_test(init-particles  test_initialization.py ${CMAKE_CURRENT_BINARY_DIR})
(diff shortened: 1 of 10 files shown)


tests/simulator/CMakeLists.txt
```diff
@@ -10,9 +10,12 @@ endif()
 if(HighFive)
 
   ## These test use dump diagnostics so require HighFive!
-  add_python3_test(diagnostics     diagnostics.py         ${CMAKE_CURRENT_BINARY_DIR})
-  add_python3_test(init-particles  test_initialization.py ${CMAKE_CURRENT_BINARY_DIR})
-  add_python3_test(sim-refineboxes refinement_boxes.py    ${CMAKE_CURRENT_BINARY_DIR})
+  add_python3_test(      diagnostics diagnostics.py  ${CMAKE_CURRENT_BINARY_DIR}) # serial or n = 2
```
nicolasaunai, line 13: I find it a bit msileading that there are _mpi_ tests that do not actually apply for n=2, now they are here you would expect 2 to fall in that category,no?

nicolasaunai (approved):
good to merge but conficts

PR #340, opened 2020-09-21
cmake: add google benchmark 

Dekken:


nicolasaunai (approved):
good for now I guess.
We should/could create issues to:
- add a priority level for each bench, so some can run on PRs while other not
- add bench not as test so outputs can be easily extracted and used to make plots on TC

Dekken:
1. I'll update this PR to use an exec_level in the new functions
2. https://github.com/PHAREHUB/PHARE/issues/346

PR #328, opened 2020-09-14
functional tests

nicolasaunai:

PR #353, opened 2020-10-07
test ohm is fine with a new fixture

rochSmets:
#353 : with a new fixture for 1, 2 & 3D test of Ohm

### DIFF
--- a/src/core/numerics/ohm/ohm.h
+++ b/src/core/numerics/ohm/ohm.h
@@ -312,9 +312,9 @@ namespace core
             for (auto ix = ix0; ix <= ix1; ++ix)
             {
                 Ex(ix) = ideal1D_(Ve, B, {ix}, ComponentTag<Component::X>{})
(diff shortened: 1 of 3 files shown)


tests/core/numerics/ohm/:w
```diff
@@ -0,0 +1,643 @@
+
```
Dekken, line 1: is the name of this file possibly because of vim? :)

Dekken: seems identical to https://github.com/rochSmets/PHARE/blob/ohm/tests/core/numerics/ohm/test_ohm.py

nicolasaunai (commented):
ok, agreed with @Dekken there is one file to remove.
Also, you have reintroduced the non ideal terms in Ohm's law and I would like to run functional tests in #328  with them before this gets merged maybe

nicolasaunai (approved):

PhilipDeegan (approved):

Issue #350, opened 2020-10-03
griddingAlgorithmDatabase uppercell discrepancy 


Dekken:
https://github.com/PHAREHUB/PHARE/blob/f221369215b0818c1f5a8ca803faac95972d208a/src/amr/wrappers/hierarchy.h#L78

Dekken:
assessed in https://github.com/PHAREHUB/PHARE/pull/349

PR #354, opened 2020-10-07
Small bug fixes

Dekken:
so we don't forget about them

src/amr/wrappers/hierarchy.h
   https://github.com/PHAREHUB/PHARE/issues/350

src/solver/multiphysics_integrator.h
   First step of subcycle need current newTime of coarser level (credit it @nicolasaunai)


nicolasaunai (approved):
thx

PR #361, opened 2020-10-12
Python simulation init dim==1 level ghost test

Dekken:


pyphare/pyphare/pharesee/geometry.py
```diff
@@ -245,28 +247,34 @@ def level_ghost_boxes(hierarchy):
...
-                    # not intersect other patch data interior
+                for patch in patches:
 
-                    if True:  # if periodic, always true for now
-                        refined_domain_box = hierarchy.refined_domain_box(ilvl)
-                        patches = get_periodic_list(lvl.patches, refined_domain_box)
+                    particles_ids = [key for key in patch.patch_datas.keys() if key.endswith("particles")]
 
-                    for patch in patches:
+                    for particles_id in particles_ids:
```
nicolasaunai, line 274: ok I think same as the above function, these level ghost boxes are going to be the same for all populations so I would say there is no need to loop over them all.

However I know see that the dict we return also gets the patch data which is going to be different for each pop (same applies to the function above actually...)

so... it seems stupid to get Npop times the same boxes, but it also seems stupid to get only 1 patchdata... (why that one ?)

tests/simulator/test_initialization.py
```diff
@@ -582,5 +583,79 @@ def test_overlapped_particledatas_have_identical_particles(self, interp_order, r
...
+
+
+        self.assertTrue(len(level_ghost_boxes(datahier).items()))
+        for ilvl, particle_gaboxes in level_ghost_boxes(datahier).items():
+            self.assertTrue(ilvl > 0) # has no level 0
+            coarse_particles = domainParticles_for(ilvl - 1)
+
+            self.assertTrue(all([particles.size() > 0 for k, particles in coarse_particles.items()]))
+
+            coarse_split_particles = {k: particles.split(sim) for k, particles in coarse_particles.items()}
```
nicolasaunai, line 618: rename k to pop_id ? (same above for particle_id, it's more pop_id no?)

tests/simulator/test_initialization.py
```diff
@@ -582,5 +583,79 @@ def test_overlapped_particledatas_have_identical_particles(self, interp_order, r
...
+
+            for k, particles in coarse_particles.items():
+                self.assertTrue(coarse_split_particles[k].size() > 0)
+                self.assertTrue(coarse_split_particles[k].size() == particles.size() * sim.refined_particle_nbr)
+
+            for gabox in particle_gaboxes:
+                gabox_patchData = gabox["pdata"]
+                particles_id = gabox_patchData.pop_name + "_particles"
```
nicolasaunai, line 626: I think this works only because you have all boxes for all populations, right? I mean that if we would stop duplicating the boxes per population that would fail ?

Dekken: ghost boxes imply data IMO, so I don't see it as duplication.

pyphare/pyphare/pharesee/geometry.py
Dekken: I don't think particles_ids is ever > 1 in size

tests/simulator/test_initialization.py
Dekken: done

tests/simulator/test_initialization.py
Dekken: there's probably a simpler way about it tho as I guess you're thinking of

Dekken: perhaps we can discuss and make a new issue with a strategy

pyphare/pyphare/pharesee/geometry.py
nicolasaunai: you mean in this test but in general yes ?

Dekken: I think in general, will confirm

Dekken: I don't see why it would be only size one, it should be equal to the total number of populations in the simulation.
We either need to, return all ghost boxes for all populations, or assume that the caller of the function knows that there's only one ghost box for all populations. I'm not sure which is simpler, because for the latter on the caller side you have to go through the patch_datas again and check if they are particles, but if you just give back a box per patch_data of type particles, you know its' the right one

Dekken: how it is here is, a ghost box per patch_data, so if that's the what's decided, I don't think anything needs to change

nicolasaunai: Technically the function returns the level ghost boxes, which a priori has nothing to do with particle populations since per definition particle ghost width is invariant with respect to populations. The caller should know that I think.
This being said, we can also keep it the way it is now and see if that's inconvenient at some point.

nicolasaunai (approved):

PR #382, opened 2020-11-11
Tagging

nicolasaunai:
#9 

Add Tagger for automatic refinement
`MultiphysicsIntegrator`  uses the Tagger, which is abstract and actually depends on which model is being solved at current level.
There are therefore concrete taggers, per model. Each concrete Tagger can then have multiple tagging strategies. Only 1 is now coded. (Other attempts may come later, and also may be taken from user supplied python tagging methods.)

Bug fixes

- first bug fix concerns EM_old and Jold in the hybridMessenger. These quantities are supposed to store the current state of those model variables at time n before the level advances to time n+1. They are then used by next finer level to get ghosts at subcycle times which are per definition between coarse steps n and n+1. before this PR, the "old" quantities had their patchdata timestamp fixed at the creation of the level but never updated afterwards. The current PR thus sets their timestamps in `prepareStep`, i.e. before every advance().

- fix bug concerning regriding: Regriding means taking a level out of the hierarchy and replacing it by a new one. This action invalidates all the schedules stored in the RefinerPool that deal with the level being replaced. This PR now re-create schedules for the level that's being regrided and the next finer as well. This is important because SAMRAI typically will do things in the following order:

- for a hierarchy with 2 levels:
- tag level 1
- create level 2 
- regrid level 1

thus at the time L2 is created, its schedules are made with the L1 taht is about to be taken out of the hierarchy. This PR now re-makes these schedules when regriding L1.

- the PR also adds particles to the maxwellian initializer test as well as prints, in an attempt to know more about latest (random?) failures observed of this tests...

tagging : 
The concrete tagging strategy that is implemented tags a cell (==1) when the local 3 points average of the magnetic field differs from the current value from more than 10%. It's a first attempt that seems to give consistent results in tha tdtagged functional test but will need further attention in upcoming PRs.

Notes : 
- The concrete tagging strategy ignores what samrai stuff is, and probably could be moved to a core lib section. That should be considered later as this matter possibly also concern other components.

### DIFF
--- a/src/solver/multiphysics_integrator.h
+++ b/src/solver/multiphysics_integrator.h
@@ -16,6 +16,7 @@
 
 
 #include "amr/messengers/messenger.h"
+#include "amr/tagging/tagger.h"
 
 #include "solver/physical_models/hybrid_model.h"
 #include "solver/physical_models/mhd_model.h"
@@ -37,10 +38,11 @@ namespace solver
 {
     struct LevelDescriptor
     {
-        static int const NOT_SET  = -1;
-        int modelIndex            = NOT_SET;
-        int solverIndex           = NOT_SET;
-        int resourcesManagerIndex = NOT_SET;
+        static int constexpr NOT_SET = -1;
+        int modelIndex               = NOT_SET;
+        int solverIndex              = NOT_SET;
+        int resourcesManagerIndex    = NOT_SET;
+        int taggerIndex              = NOT_SET;
         std::string messengerName;
     };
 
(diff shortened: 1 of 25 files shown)


src/solver/multiphysics_integrator.h
```diff
@@ -118,6 +120,21 @@ namespace solver
         auto nbrOfLevels() const { return nbrOfLevels_; }
 
 
+        void registerTagger(int coarsestLevel, int finestLevel,
+                            std::unique_ptr<PHARE::amr::Tagger> tagger)
```
Dekken, line 124: does "std::unique_ptr<PHARE::amr::Tagger> && tagger" work better?

the notion of a copy of a unique_ptr doesn't make much sense

Dekken:
LGTM otherwise

src/solver/multiphysics_integrator.h
PhilipDeegan: if we every tried to copy a unique_ptr it would error at compile so maybe it doesn't matter

Issue #363, opened 2020-10-13
Test Field refinement in 2D and 3D


nicolasaunai:
test/amr/data/field/refine/test_basic.hierarchy.cpp

the test "ALinearFieldRefine1DO1, conserveLinearFunction" is initializing an AMR hierarchy parametrized by the input/input_1d_ratio_xx.txt files where basically all PHARE quantity fields live. 

It initializes the root level with a linear function of the position x, and the finer levels with a refinement schedule.

The test checks that fine levels nodes, for all fields, equal the value given by the linear function at these nodes. This should be true exactly because the refinement itself is linear.

The goal of this issue is to update the test so that it runs in 2 and 3D.
Non exhaustive and rough lsit of things to consider : 

- [x] having an input txt file for a 2D and a 3D case, only for refinement ratio 2 is enough
- [x] modifying the tagging TagStrategy to intialize root level with a 2D and 3D linear function. These functions should use x AND y in 2D, and x and y and z in 3D so that each node value is unique.
 - [x] considering having a linear function that is different for each quantity is a plus.

rochSmets:
I'm starting to work on this one...

rochSmets:
ok

PR #383, opened 2020-11-12
the field data refinement typed tests now work for 1, 2 & 3d

rochSmets:
#363


pyphare/.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
```
nicolasaunai, line 1: add .idea to the gitignore

nicolasaunai (dismissed):
LGTM, can you add a comment, for instance at the location where input.txt files are read, or even better, in the input.txt files themselves if the samrai syntax allows for comments, that the test assumes the refinement boxes are such that the refined patches should NOT have any ghost beyond the coarest level domain boundaries, because periodicity would apply and ghost nodes would NOT have the value expected from the linear function.

nicolasaunai (commented):
LGTM, can you add a comment, for instance at the location where input.txt files are read, or even better, in the input.txt files themselves if the samrai syntax allows for comments, that the test assumes the refinement boxes are such that the refined patches should NOT have any ghost beyond the coarest level domain boundaries, because periodicity would apply and ghost nodes would NOT have the value expected from the linear function.

nicolasaunai:
actually @rochSmets #363 mentions that we could have a different affine function for each quantity that is refined, what do you think? doing that would allow us to be sure patchdatas are not mixed with each others while refining.

pyphare/.idea/inspectionProfiles/profiles_settings.xml
rochSmets: such comments have been added

rochSmets:
I added the the dataId value in the coeff for affineFill, so all components have different coefficients.
I guess I answered to all your comments guys,

nicolasaunai (commented):
remove .idea/* files and we merge

nicolasaunai (approved):

PhilipDeegan (approved):

PR #390, opened 2020-11-17
domain particles are correct from split particles tests

Dekken:


nicolasaunai (approved):
appart from Particles.__eq__ to do maybe, ok thx

PR #399, opened 2020-12-09
Add shock intial condition in functional tests

nicolasaunai:
add shock functional test.

Add Tagger for automatic refinement
MultiphysicsIntegrator uses the Tagger, which is abstract and actually depends on which model is being solved at current level.
There are therefore concrete taggers, per model. Each concrete Tagger can then have multiple tagging strategies. Only 1 is now coded. (Other attempts may come later, and also may be taken from user supplied python tagging methods.)

Bug fixes

    first bug fix concerns EM_old and Jold in the hybridMessenger. These quantities are supposed to store the current state of those model variables at time n before the level advances to time n+1. They are then used by next finer level to get ghosts at subcycle times which are per definition between coarse steps n and n+1. before this PR, the "old" quantities had their patchdata timestamp fixed at the creation of the level but never updated afterwards. The current PR thus sets their timestamps in prepareStep, i.e. before every advance().

    fix bug concerning regriding: Regriding means taking a level out of the hierarchy and replacing it by a new one. This action invalidates all the schedules stored in the RefinerPool that deal with the level being replaced. This PR now re-create schedules for the level that's being regrided and the next finer as well. This is important because SAMRAI typically will do things in the following order:

    for a hierarchy with 2 levels:

    tag level 1

    create level 2

    regrid level 1

thus at the time L2 is created, its schedules are made with the L1 taht is about to be taken out of the hierarchy. This PR now re-makes these schedules when regriding L1.

    the PR also adds particles to the maxwellian initializer test as well as prints, in an attempt to know more about latest (random?) failures observed of this tests...

tagging :
The concrete tagging strategy that is implemented tags a cell (==1) when the local 3 points average of the magnetic field differs from the current value from more than 10%. It's a first attempt that seems to give consistent results in tha tdtagged functional test but will need further attention in upcoming PRs.

Notes :

    The concrete tagging strategy ignores what samrai stuff is, and probably could be moved to a core lib section. That should be considered later as this matter possibly also concern other components.

### DIFF
--- a/pyphare/pyphare/core/gridlayout.py
+++ b/pyphare/pyphare/core/gridlayout.py
@@ -15,21 +15,23 @@ yee_centering = {
         'Ex':'dual', 'Ey':'primal', 'Ez':'primal',
         'Jx':'dual', 'Jy':'primal', 'Jz':'primal',
         'rho':'primal', 'Vx':'primal','Vy':'primal',
-        'Vz':'primal','P':'primal'
+        "Fx":"primal","Fy":"primal","Fz":"primal",
+        'Vz':'primal','P':'primal', "Fx":"primal","Fy":"primal","Fz":"primal"
     },
     'y' : {
         'Bx':'dual', 'By':'primal', 'Bz':'dual',
         'Ex':'primal', 'Ey':'dual', 'Ez':'primal',
         'Jx':'primal', 'Jy':'dual', 'Jz':'primal',
         'rho':'primal', 'Vx':'primal','Vy':'primal',
-        'Vz':'primal','P':'primal'
+        'Vz':'primal','P':'primal', "Fx":"primal","Fy":"primal","Fz":"primal"
     },
     'z' : {
         'Bx':'dual', 'By':'dual', 'Bz':'primal',
         'Ex':'primal', 'Ey':'primal', 'Ez':'dual',
         'Jx':'primal', 'Jy':'primal', 'Jz':'dual',
         'rho':'primal', 'Vx':'primal','Vy':'primal',
-        'Vz':'primal', 'P':'primal'
+        "Fx":"primal","Fy":"primal","Fz":"primal",
+        'Vz':'primal', 'P':'primal', "Fx":"primal","Fy":"primal","Fz":"primal"
     }
 }
 yee_centering_lower = {
@@ -71,7 +73,7 @@ class GridLayout(object):
                                  'Ex','Ey','Ez',
                                  'Jx','Jy','Jz',
                                  'rho','Vx','Vy',
-                                 'Vz','P']
+                                 'Vz','P', 'Fx', 'Fy', 'Fz']
 
 
         self.yeeCentering = YeeCentering()
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -179,6 +179,47 @@ class PatchLevel:
                 max([patch.patch_datas[name].x.max() for patch in self.patches])
 
 
+
+def finest_data(pdata, ilvl, hierarchy):
+    """
+    given a FieldData, a levelnumber and a hierarchy
+    this function returns the coordinates and values
+    of the field at locations where no value exist on the
+    next finer level in the hierachy.
+    """
+    # although a priori doable for particles
+    # this just works for fields for now
+
+    assert pdata.quantity == 'field'
+    assert hierarchy.dim == 1
+
+    x_ = pdata.x
+    v_ = pdata.dataset
+
+    if ilvl == hierarchy.finest_level():
+        return x_,v_
+
+    qtyname = pdata.field_name
+    inner = x_ != x_ #lgtm [py/comparison-of-identical-expressions]
+
+    # iteratively fill the mask with true where current patch coordinates
+    # are within limits of the next refined level patchdatas
+    for ipatch, finer_patch in enumerate(hierarchy.patch_levels[ilvl+1].patches):
+        pdat = finer_patch.patch_datas[qtyname]
+        xmin,xmax = [pdat.x[ix] for ix in (4,-4)]
+        inner  = inner | ((x_ > xmin) & (x_ < xmax))
+
+    # now take the complement of the mas
+    # i.e. data that has coordinates not existing on
+    # next finer level
+    x = x_[~inner]
+    v = v_[~inner]
+
+    return x, v
+
+
+
+
 class PatchHierarchy:
     """is a collection of patch levels """
 
@@ -196,6 +237,9 @@ class PatchHierarchy:
 
         if data_files is not None:
             self.data_files.update(data_files)
+
+    def finest_level(self):
+        return len(self.patch_levels)-1
 
 
     def levels(self, time=0.):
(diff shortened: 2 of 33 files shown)


nicolasaunai:
LGTM "checkout issue" : https://github.com/github/codeql/issues/4809

pyphare/pyphare/core/gridlayout.py
```diff
@@ -11,20 +11,22 @@
...
+        "Fx":"primal","Fy":"primal","Fz":"primal",
         'Vz':'primal','P':'primal'
     },
     'y' : {
         'Bx':'dual', 'By':'primal', 'Bz':'dual',
         'Ex':'primal', 'Ey':'dual', 'Ez':'primal',
         'Jx':'primal', 'Jy':'dual', 'Jz':'primal',
         'rho':'primal', 'Vx':'primal','Vy':'primal',
-        'Vz':'primal','P':'primal'
+        'Vz':'primal','P':'primal', "Fx":"primal","Fy":"primal","Fz":"primal"
```
PhilipDeegan, line 22: the odd one out

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -179,6 +179,47 @@ def level_range(self):
...
+    assert hierarchy.dim == 1
+
+    x_ = pdata.x
+    v_ = pdata.dataset
+
+    if ilvl == hierarchy.finest_level():
+        return x_,v_
+
+    qtyname = pdata.field_name
+    inner = x_ != x_ #lgtm [py/comparison-of-identical-expressions]
```
PhilipDeegan, line 203: whats' this doing?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -179,6 +179,47 @@ def level_range(self):
...
+
+    # iteratively fill the mask with true where current patch coordinates
+    # are within limits of the next refined level patchdatas
+    for ipatch, finer_patch in enumerate(hierarchy.patch_levels[ilvl+1].patches):
+        pdat = finer_patch.patch_datas[qtyname]
+        xmin,xmax = [pdat.x[ix] for ix in (4,-4)]
+        inner  = inner | ((x_ > xmin) & (x_ < xmax))
```
PhilipDeegan, line 210: not so obvious

pyphare/pyphare/simulator/simulator.py
```diff
@@ -96,7 +96,7 @@ def run(self):
         print("mean advance time = {}".format(np.mean(perf)))
         print("total advance time = {}".format(np.sum(perf)))
 
-        print(perf)
+        #print(perf)
```
PhilipDeegan, line 99: rm

src/amr/tagging/default_hybrid_tagger_strategy.h
```diff
@@ -0,0 +1,79 @@
...
+namespace PHARE::amr
+{
+template<typename HybridModel>
+class DefaultHybridTaggerStrategy : public HybridTaggerStrategy<HybridModel>
+{
+    using gridlayout_type           = typename HybridModel::gridlayout_type;
+    static auto constexpr dimension = HybridModel::dimension;
+
+public:
+    void tag(HybridModel& model, gridlayout_type const& layout, int* tags) const override;
```
PhilipDeegan, line 18: is this interacting with SAMRAI somehow that it has to be a raw "int"?

src/amr/tagging/default_hybrid_tagger_strategy.h
```diff
@@ -0,0 +1,79 @@
...
+    double threshold = 0.05;
+    int idx          = 4;
+    double epsilon   = 0.1;
+
+
+    if constexpr (dimension == 1)
+    {
+        for (auto iCell = 0u, ix = start; iCell <= endCell; ++ix, ++iCell)
+        {
+            auto Byavgm2 = 0.2 * (By(ix - 4) + By(ix - 3) + By(ix - 2) + By(ix - 1) + By(ix));
```
PhilipDeegan, line 46: is there some documentation to go along with this as a reference?

it's pretty complicated

src/solver/multiphysics_integrator.h
```diff
@@ -367,7 +391,7 @@ namespace solver
             // so dividing the mesh size by ratio means dt
             // needs to be divided by ratio^2.
             // we multiply that by a constant < 1 for safety.
-            return coarseDt / (ratio.max() * ratio.max()) * 0.4;
+            return coarseDt / (ratio.max() * ratio.max()) /** 0.4*/;
```
PhilipDeegan, line 394: TODO?

src/solver/multiphysics_integrator.h
```diff
@@ -525,6 +550,21 @@ namespace solver
...
+            bool hasTagger = true;
+
+            for (auto iLevel = coarsestLevel; iLevel <= finestLevel; ++iLevel)
+            {
+                if (levelDescriptors_[iLevel].taggerIndex != LevelDescriptor::NOT_SET)
+                {
+                    return hasTagger;
+                }
+            }
+            return !hasTagger;
```
PhilipDeegan, line 564: "hasTagger" seems a bit pointless

src/amr/tagging/default_hybrid_tagger_strategy.h
```diff
@@ -0,0 +1,79 @@
...
+            auto Navg   = 0.2 * (N(ix - 2) + N(ix - 1) + N(ix) + N(ix + 1) + N(ix + 2));
+
+            auto criter_by = std::abs(Byavgp1 - Byavg) / (1 + std::abs(Byavg));
+            auto criter_bz = std::abs(Bzavgp1 - Bzavg) / (1 + std::abs(Bzavg));
+            auto criter_b  = std::sqrt(criter_by * criter_by + criter_bz * criter_bz);
+            auto criter_n  = std::abs(Navgp1 - Navg) / (1 + std::abs(Navg));
+
+            auto criter = std::max(criter_b, criter_n);
+
+            if (criter > threshold)
```
PhilipDeegan, line 68:
```
 tags[iCell] = criter > threshold;
```

tests/amr/tagging/test_tagging.cpp
```diff
@@ -0,0 +1,222 @@
...
+        model.state.electromag.E.setBuffer("EM_E_y", &Ey);
+        model.state.electromag.E.setBuffer("EM_E_z", &Ez);
+        model.state.electromag.B.setBuffer("EM_B_x", &Bx);
+        model.state.electromag.B.setBuffer("EM_B_y", &By);
+        model.state.electromag.B.setBuffer("EM_B_z", &Bz);
+        model.state.electromag.initialize(layout);
+    }
+};
+
+/*
```
PhilipDeegan, line 178: TODO?

tests/functional/alfven_wave/alfven_wave1d.py
```diff
@@ -165,9 +163,9 @@ def main():
     simulator.run()
 
 
-    if cpp.mpi_rank() == 0:
-        b = hierarchy_from(h5_filename="phare_outputs/EM_B.h5")
-        plot(b)
-
+    #if cpp.mpi_rank() == 0:
```
PhilipDeegan, line 166: unnecessary, or was there an issue?

src/amr/tagging/default_hybrid_tagger_strategy.h
nicolasaunai: yes

tests/functional/alfven_wave/alfven_wave1d.py
nicolasaunai: tbd

PhilipDeegan (approved):

PR #401, opened 2020-12-16
fix some lgtm issues

nicolasaunai:


pyphare/pyphare/pharein/init.py
```diff
@@ -3,7 +3,7 @@ def get_user_inputs(jobname):
     import importlib
     from . import populateDict
     print(jobname)
-    jobmodule = importlib.import_module(jobname)
+    jobmodule = importlib.import_module(jobname) #lgtm [py/unused-local-variable]
```
PhilipDeegan, line 6: does 

```jobmodule = ```

need to be there?

PhilipDeegan (approved):

PR #409, opened 2021-01-06
Refactor box.py / simulation validation 

PhilipDeegan:
refactor box.cells() -> box.nCells()
change simulation validation to allow a finer box to be over a set of coarser boxes


nicolasaunai (approved):

nicolasaunai:
should this new check come with a test case that says it works ?

PhilipDeegan:
can add one here https://github.com/PHAREHUB/PHARE/blob/master/tests/simulator/refinement_boxes.py#L60

PR #426, opened 2021-01-08
refactor diagnostic timestamp checks 

PhilipDeegan:
such that the should compute/write checks are based on incremental windows rather than (> $number)


nicolasaunai (commented):
I think we still need to keep a python validation of the user time stamps. They should : 
- not be smaller than initial time (that is 0 or any restart time once we have the restart mechanism)
- not be larger than the final simulation time
- two consecutive time stamps should not be distant by less than the simulation time step (more thoughts is required for fine level dumps ?)

PhilipDeegan:
> * two consecutive time stamps should not be distant by less than the simulation time step (more thoughts is required for fine level dumps ?)

not sure it's relevant, we dump all levels at coarsest time unless doing subcycle dumps then it's just all times up to a defined level

nicolasaunai (approved):

PR #437, opened 2021-01-17
cast to float to drop double imprecision

PhilipDeegan:


src/diagnostic/diagnostic_manager.h
```diff
@@ -92,17 +92,17 @@ class DiagnosticsManager : public IDiagnosticsManager
     bool needsWrite(DiagnosticProperties& diag, double timeStamp, double timeStep)
     {
         auto nextWrite = nextWrite_[diag.type + diag.quantity];
-
         return nextWrite < diag.writeTimestamps.size()
-               and std::abs(diag.writeTimestamps[nextWrite] - timeStamp) < timeStep;
+               and static_cast<float>(std::abs(diag.writeTimestamps[nextWrite] - timeStamp))
```
nicolasaunai, line 96: can we add a sweet little comment explaining this not so sweet issue we had and pushed us to do this otherwise weird looking static casts?

tests/simulator/test_advance.py
```diff
@@ -244,7 +244,7 @@ def test_hierarchy_timestamp_cadence(self, refinement_boxes):
         dim = refinement_boxes["L0"]["B0"].ndim
 
         time_step     = .001
-        time_step_nbr = 10
+        time_step_nbr = 101
```
nicolasaunai, line 247: we should add a comment there too to explain why this weird looking 101

src/diagnostic/diagnostic_manager.h
PhilipDeegan: sure thing

nicolasaunai (approved):

Issue #346, opened 2020-09-28
cmake: benchmarking target


PhilipDeegan:
+ test harness to interface with the benchmarks statistics to allow test failures on slow down

PR #440, opened 2021-01-26
push benchmark targets per param

PhilipDeegan:
ref: https://github.com/PHAREHUB/PHARE/issues/346

possible extensions:
 test per entry from set of compile flag strings to see how things differ
 testing with various mallocs (jemalloc / tcmalloc)

### DIFF
--- a/tools/bench/functional/simulation_setup.py
+++ b/tools/bench/functional/simulation_setup.py
@@ -1,0 +1,103 @@
+def _density(*xyz):
+    return 1.0
+
+
+def _bx(*xyz):
+    return 1.0
+
+
+def _by(*xyz):
+    return 0.0
+
+
+def _bz(*xyz):
+    return 0.0
+
+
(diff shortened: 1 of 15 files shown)


tools/bench/functional/simulation_setup.py
```diff
@@ -0,0 +1,101 @@
...
+        smallest_patch_size=smallest_patch_size,
+        largest_patch_size=largest_patch_size,
+        time_step_nbr=time_step_nbr,
+        final_time=final_time,
+        boundary_types=boundary_types,
+        cells=cells,
+        dl=dl,
+        diag_options={
+            "format": "phareh5",
+            "options": {"dir": "dispersion", "mode": "overwrite"},
```
nicolasaunai, line 80: dir name? different for each so keyword?

nicolasaunai (approved):
I approve mostly because I'm ignorant

PR #443, opened 2021-02-10
Resistivities

rochSmets:
include resistivity and hyper_resistivity as coming from the dict filled via Python in `populateDict`

### DIFF
--- a/pyphare/pyphare/pharein/simulation.py
+++ b/pyphare/pyphare/pharein/simulation.py
@@ -406,6 +406,25 @@ def check_optional_keywords(**kwargs):
     return extra
 
 
+
+def check_resistivity(**kwargs):
+    resistivity = kwargs.get("resistivity", 0.0)
+    if resistivity < 0.0:
+            raise ValueError(f"Error: resistivity should not be negative")
+
+    return resistivity
+
+
+
+def check_hyper_resistivity(**kwargs):
+    hyper_resistivity = kwargs.get("hyper_resistivity", 0.0001)
+    if hyper_resistivity < 0.0:
+            raise ValueError(f"Error: hyper_resistivity should not be negative")
+
+    return hyper_resistivity
+
+
+
 # ------------------------------------------------------------------------------
 
 def checker(func):
@@ -414,7 +433,8 @@ def checker(func):
                              'time_step', 'time_step_nbr', 'layout', 'interp_order', 'origin',
                              'boundary_types', 'refined_particle_nbr', 'path', 'nesting_buffer',
                              'diag_export_format', 'refinement_boxes', 'refinement', 'init_time',
-                             'smallest_patch_size', 'largest_patch_size', "diag_options" ]
+                             'smallest_patch_size', 'largest_patch_size', "diag_options",
+                             'resistivity', 'hyper_resistivity' ]
 
         accepted_keywords += check_optional_keywords(**kwargs)
 
@@ -467,6 +487,10 @@ def checker(func):
             kwargs["max_nbr_levels"] = kwargs.get('max_nbr_levels', None)
             assert kwargs["max_nbr_levels"] != None # this needs setting otherwise
             kwargs["refinement_boxes"] = None
+
+        kwargs["resistivity"] = check_resistivity(**kwargs)
+
+        kwargs["hyper_resistivity"] = check_hyper_resistivity(**kwargs)
 
         return func(simulation_object, **kwargs)
 
@@ -551,6 +575,8 @@ class Simulation(object):
             # 1D case
             return extent[0] >= domain[0] and extent[1] <= domain[1]
         raise NotImplementedError("Error: 2D and 3D not implemented yet")
+
+
 
 
 
--- a/src/core/numerics/ohm/ohm.h
+++ b/src/core/numerics/ohm/ohm.h
@@ -10,6 +10,10 @@
 #include "core/data/vecfield/vecfield_component.h"
 #include "core/utilities/index/index.h"
 
+#include "phare_core.h"
+#include "initializer/data_provider.h"
+
+
 namespace PHARE
 {
 namespace core
@@ -17,7 +21,17 @@ namespace core
     template<typename GridLayout>
     class Ohm : public LayoutHolder<GridLayout>
     {
+    public:
+        explicit Ohm(PHARE::initializer::PHAREDict dict)
+            : eta_{dict["resistivity"].template to<double>()}
+            , nu_{dict["hyper_resistivity"].template to<double>()}
+        {
+        }
+
     private:
+        double eta_;
+        double nu_;
+
         template<typename VecField, typename ComponentTag>
         auto ideal1D_(VecField const& Ve, VecField const& B, MeshIndex<1> index, ComponentTag) const
         {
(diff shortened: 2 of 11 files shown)


src/solver/level_initializer/hybrid_level_initializer.h
```diff
@@ -35,7 +35,9 @@ namespace solver
         static constexpr auto interp_order = GridLayoutT::interp_order;
 
 
-        PHARE::core::Ohm<GridLayoutT> ohm_;
+        PHARE::core::Ohm<GridLayoutT>
+            ohm_(dict); // where the fuck should this dict come from ? should we make an explicit
```
PhilipDeegan, line 39: either 

```
explicit HybridLevelInitializer(PHARE::core::Ohm<GridLayoutT>&& ohm) : ohm_{ohm} {
```

or

```
explicit HybridLevelInitializer(PHARE::initializer::PHAREDict dict) : ohm_{dict} {
```

I suppose, the former is minimal, the latter is more extensible should "ampere_" need parameters later for example

pyphare/pyphare/pharein/simulation.py
```diff
@@ -536,6 +560,9 @@ def __init__(self, **kwargs):
           self.nSubcycles ** levelNumbers[ilvl] * self.time_step_nbr for ilvl in levelNumbers
         ]
 
+        print("resistivity : ", self.resistivity)
```
nicolasaunai, line 563: probably not needed now?

src/core/numerics/ohm/ohm.h
```diff
@@ -10,14 +10,29 @@
...
     template<typename GridLayout>
     class Ohm : public LayoutHolder<GridLayout>
     {
+    public:
+        explicit Ohm(PHARE::initializer::PHAREDict dict)
+            : eta_{dict["resistivity"].template to<double>()}
+            , nu_{dict["hyper_resistivity"].template to<double>()}
+        {
+        }
+        // Ohm() = default;
```
nicolasaunai, line 30: remove?

src/solver/level_initializer/level_initializer_factory.h
```diff
@@ -17,11 +18,12 @@ namespace solver
...
 
     public:
-        static std::unique_ptr<LevelInitializer<AMRTypes>> create(std::string modelName)
+        static std::unique_ptr<LevelInitializer<AMRTypes>>
+        create(std::string modelName, PHARE::initializer::PHAREDict dict)
         {
             if (modelName == "HybridModel")
             {
-                return std::make_unique<HybridLevelInitializer<HybridModel>>();
+                return std::make_unique<HybridLevelInitializer<HybridModel>>(dict);
```
nicolasaunai, line 26: should we do std::move(dict) here?

tests/core/numerics/ohm/test_main.cpp
```diff
@@ -303,19 +320,34 @@ TYPED_TEST_SUITE(OhmTest, OhmTupleInfos);
 
 
 
-TYPED_TEST(OhmTest, ThatOhmHasCtor)
+TYPED_TEST(OhmTest, ThatOhmHasDefaultCtor) // this test should be deprecated in a close future as
```
nicolasaunai, line 323: is it still needed?

tests/functional/alfven_wave/alfven_wave1d.py
```diff
@@ -21,15 +21,17 @@ def config():
...
+        smallest_patch_size = 20,
+        largest_patch_size = 20,
+        time_step_nbr = 2,           # number of time steps (not specified if time_step and final_time provided)
+        time_step = 0.001,           # simulation time step (not specified if final_time and time_step_nbr provided)
+        boundary_types = "periodic", # boundary condition, string or tuple, length == len(cell) == len(dl)
+        cells = 40,                  # integer or tuple length == dimension
+        dl = 0.2,                    # mesh size of the root level, float or tuple
+        refinement_boxes = {"L0": {"B0": [(10, ), (20, )]}},
+        diag_options = {"format": "phareh5", "options": {"dir": "alfven_norefine","mode":"overwrite"}},
+        resistivity = 0.01,
```
nicolasaunai, line 33: I'm not sure why the need to modify the alfven wave test in that PR?

src/core/numerics/ohm/ohm.h
```diff
@@ -239,7 +254,7 @@ namespace core
         template<typename VecField, typename ComponentTag>
         auto resistive_(VecField const& J, MeshIndex<VecField::dimension> index, ComponentTag) const
         {
-            auto const eta = 1.0; // TODO : eta should comme from input file
+            auto const eta = this->eta_;
```
PhilipDeegan, line 257: Possibly unnecessary copy?

src/solver/level_initializer/level_initializer_factory.h
PhilipDeegan: > should we do std::move(dict) here?

I think it's not needed as it's RVO, a copy shoud(would?) cause a compile error

PhilipDeegan: I think it's not needed as it's RVO, a copy shoud(would?) cause a compile error

PhilipDeegan: http://coliru.stacked-crooked.com/a/c5f1ed9aec4a50b4

src/core/numerics/ohm/ohm.h
rochSmets: not clear to why should we remove these 4 lines... init `nu_` is important no ?

src/solver/level_initializer/level_initializer_factory.h
rochSmets: so it should stays as is ?

src/core/numerics/ohm/ohm.h
rochSmets: just because `eta` is shorter than `this->eta_`

PhilipDeegan: sorry, I meant the lack of ref(&)

```
auto const& eta = this->eta_;
```

I think copying doubles is super fast anyway so maybe it's not all that important.

src/solver/level_initializer/level_initializer_factory.h
nicolasaunai: I meant 
```
 return std::make_unique<HybridLevelInitializer<HybridModel>>(std::move(dict)); 
```

PhilipDeegan: ohh right, no I don't think it's needed
maybe just make this[1] a const ref, I added a few const functions to dict.hpp so it should be ok

[1] https://github.com/rochSmets/PHARE/blob/resistivities/src/solver/level_initializer/hybrid_level_initializer.h#L47

```
explicit HybridLevelInitializer(PHARE::initializer::PHAREDict const& dict)
```

if const fails, let me know, we should assess that I think

rochSmets: done

nicolasaunai (approved):

PhilipDeegan (approved):

PR #444, opened 2021-03-01
Distplot

nicolasaunai:
- fix bugs in dist_plot and python Particles

### DIFF
--- a/pyphare/pyphare/pharesee/particles.py
+++ b/pyphare/pyphare/pharesee/particles.py
@@ -67,6 +67,7 @@ class Particles:
         self.v        = np.concatenate((self.v, particles.v))
         self.charges  = np.concatenate((self.charges, particles.charges))
         self.weights  = np.concatenate((self.weights, particles.weights))
+        self._x = None
 
 
     def shift_icell(self, offset):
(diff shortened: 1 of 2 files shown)


pyphare/pyphare/pharesee/particles.py
```diff
@@ -67,6 +67,7 @@ def add(self, particles):
         self.v        = np.concatenate((self.v, particles.v))
         self.charges  = np.concatenate((self.charges, particles.charges))
         self.weights  = np.concatenate((self.weights, particles.weights))
+        self._x = None
```
PhilipDeegan, line 70: does it make sense to move this to outside/before the if 
```
https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharesee/particles.py#L29
```

nicolasaunai: I don't see why not. Does it impact proposed changes or is it unrelated?

PhilipDeegan: kinda related, depends on whether this change is to make sure the attribute exists, or just to "None" it

either way, we'll probably end up with something like 

```
def _reset(self):
        self._x = None
        self._y = None
        self._z = None
```

given how it's set to "None" in more than one place

PhilipDeegan (approved):

PR #447, opened 2021-03-04
Diagnostics flush parameters to not open/close each dump

PhilipDeegan:


tests/diagnostic/test_diagnostics.ipp
```diff
@@ -81,6 +81,7 @@ void allFromPython_test(Simulator&& sim, std::string out_dir)
     using Hierarchy   = typename Simulator::Hierarchy;
 
     sim.dump(*sim.dMan);
+    sim.dMan.reset(); // flush h5files
```
PhilipDeegan, line 84: this is maybe not ideal

src/diagnostic/detail/h5typewriter.h
```diff
@@ -41,7 +42,15 @@ class H5TypeWriter : public PHARE::diagnostic::TypeWriter
         std::size_t maxLevel)
         = 0;
 
-    virtual void finalize(DiagnosticProperties& diagnostic) = 0;
+    void finalize(DiagnosticProperties& diagnostic)
+    {
+        ++dumpIdx_;
+        if (diagnostic.flushEvery > 0 and diagnostic.flushEvery % dumpIdx_ == 0)
```
nicolasaunai, line 48: I think the first condition is checked already in python thus redundant here?

pyphare/pyphare/pharein/__init__.py
```diff
@@ -195,6 +196,7 @@ def as_paths(rb):
         name_path = type_path + diag.name
         add(name_path + "/" + 'type' , diag.type)
         add(name_path + "/" + 'quantity' , diag.quantity)
+        add_size_t(name_path + "/" + "flush_every", diag.flush_every)
```
nicolasaunai, line 199: we have to be cautious that "flush every" could be interpreted as "every N time step" although it really means (if I understand correctly) "flush every N dump"

tests/diagnostic/test_diagnostics.h
```diff
@@ -207,10 +207,11 @@ struct Hi5Diagnostic
...
         PHARE::initializer::PHAREDict dict;
-        dict["name"]      = quantity;
-        dict["type"]      = type;
-        dict["quantity"]  = quantity;
-        dict["time_step"] = double{1};
+        dict["name"]        = quantity;
+        dict["type"]        = type;
+        dict["quantity"]    = quantity;
+        dict["time_step"]   = double{1};
+        dict["flush_every"] = std::size_t{0};
```
nicolasaunai, line 214: 0? what does it mean?

tests/diagnostic/test_diagnostics.ipp
nicolasaunai: can you elaborate a bit please?

tests/simulator/test_diagnostic_timestamps.py
```diff
@@ -107,9 +107,10 @@ def test_dump_diags_timestamps(self):
                 quantity=quantity,
                 write_timestamps=timestamps,
                 compute_timestamps=timestamps,
+                flush_every=0,
```
nicolasaunai, line 110: same, 0 ?

nicolasaunai (approved):
merge when you see fit

PR #448, opened 2021-03-04
energy conservation functional test

nicolasaunai:


tests/functional/conservation/conserv.py
```diff
@@ -0,0 +1,289 @@
...
+        for dl, nbrcell, nbrdt in zip(dls, nbrcells, nbrdts):
+            uniform(vth, dl, nbrcell, nbrdt)
+            simulator = Simulator(gv.sim)
+            simulator.initialize()
+            simulator.run()
+            gv.sim = None
+
+
+
+    paths = glob("*vth*")
```
PhilipDeegan, line 234: I haven't seen this before, what's happening here? 

Just
```
from glob import glob
glob("*vth*")
```

results in an empty list :\

PhilipDeegan, line 234: What's happening here? 

Just
```
from glob import glob
glob("*vth*")
```

results in an empty list :\

nicolasaunai: where? The runs diags are configured this way:
```
options": {"dir": "vth{}dx{}".format(vth,dl)
```
so that line should list all diag directories, no?

PhilipDeegan: ok it's directory listing with pattern matching, a bit different from what BASH considers "globbing" hence the confusion.

PhilipDeegan (approved):

tests/functional/conservation/conserv.py
PhilipDeegan: maybe "wrong level of abstraction"

```
def directories_from_wildcard(pattern): 
    from glob import glob
    return glob(pattern)
```

or something perhaps

tests/functional/conservation/conserv.py
```diff
@@ -0,0 +1,289 @@
...
+    def by(x):
+        return 0.
+
+    def bz(x):
+        return 0.
+
+    def bx(x):
+        return 1.
+
+    def vx(x):
```
rochSmets, line 56: we could also add a bulk velocity to check energy conservation no ?

tests/functional/conservation/conserv.py
```diff
@@ -0,0 +1,289 @@
...
+    for ilvl, lvl in B.patch_levels.items():
+        if lvlNbr == ilvl:
+            tot = 0.
+            for ip, patch in enumerate(lvl.patches):
+                pdata = patch.patch_datas["Bx"]
+
+                # Bx is primal, but By and Bz dual so
+                # we average By and Bz onto Bx to use them
+                # in the same formula
+                bx = patch.patch_datas["Bx"].dataset[5:-5]
```
rochSmets, line 174: there are no reason why we could change the 5 ghosts cells, but for clearness, may be these `5` and `4` should be a constant variable named `GHOST_NBR`

tests/functional/conservation/conserv.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+#############################################################
+#############################################################
+
+
+
+def main():
+
+    cases = [0.01,0.05,0.1,0.3,0.5,0.75,1,2]
```
rochSmets, line 218: could we also include anisotropic cases ?

tests/functional/conservation/conserv.py
```diff
@@ -0,0 +1,289 @@
...
+
+    dls = [0.2, 0.1]
```
rochSmets, line 220: we might also consider more cases... going from the lowest `dls` to the largest one ?

tests/functional/conservation/conserv.py
PhilipDeegan: should be accessible from https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharesee/hierarchy.py#L100

    patch.patch_datas["Bx"].ghost_nbr[0]

PR #458, opened 2021-03-12
field level ghost interpolation tests

PhilipDeegan:


tests/amr/data/field/refine/test_refine_field.py
```diff
@@ -0,0 +1,40 @@
...
+    data   = kwargs.get("data", field.dataset[:])
+    ghosts = field.ghosts_nbr
+
+    fine_box    = boxm.refine(field.box, refinement_ratio)
+    fine_layout = GridLayout(fine_box, field.origin, field.layout.dl/refinement_ratio, interp_order=field.layout.interp_order)
+
+    assert field.box.ndim == 1
+    fine_data   = np.zeros(fine_box.shape + primal_directions + (field.ghosts_nbr[0] * 2))
+
+    for i in range(field.box.shape[0] + int(primal_directions[0]) - 1):
```
nicolasaunai, line 27:
```python
ncells = 10
coarse_primal = np.arange(ncells+1)
fine_primal = np.zeros(2*ncells+1)
fine_primal[::2] = coarse_primal # coarse primal on top of fine
fine_primal[1::2] = 0.5*(coarse_primal[1:] + coarse_primal[:-1])
print(coarse_primal)
print(fine_primal)
```

and for dual

```python
ncells=10
coarse_dual = np.arange(ncells)
fine_dual = np.zeros(ncells*2)
fine_dual[2::2] = 0.75*coarse_dual[1:] + 0.25*coarse_dual[:-1]
fine_dual[3::2] = 0.25*coarse_dual[1:] + 0.75*coarse_dual[:-1]
print(coarse_dual)
print(fine_dual)
```

nicolasaunai: to adapt.... not stricly what should be done

tests/core/numerics/interpolator/interpolator_test.py
```diff
@@ -43,5 +43,16 @@ def main():
 
 
 
+def interpolate_datasets(beforeTime, afterTime, interp_time, beforeData, afterData):
+    alpha = (interp_time - beforeTime) / (afterTime - beforeTime)
+    assert beforeData.shape == afterData.shape
+    interpData = afterData.copy()
+    assert interpData.ndim == 1 # update for > 1d interpolation values
+    if interpData.ndim == 1:
+        for ix in range(interpData.shape[0]):
```
nicolasaunai, line 52: no loop needed

tests/simulator/test_advance.py
```diff
@@ -96,7 +99,8 @@ def vthz(x):
                                       "density": density,
                                       "vbulkx": vx, "vbulky": vy, "vbulkz": vz,
                                       "vthx": vthx, "vthy": vthy, "vthz": vthz,
-                                      "nbr_part_per_cell": nbr_part_per_cell})
+                                      "nbr_part_per_cell": nbr_part_per_cell,
+                                      "init": {"seed": 1337}})
```
nicolasaunai, line 103: do we want to keep the seed? I think it's better without no? (more chances to see a bug?)

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,106 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+              if it had no refined level to the level field ghost of L1 of L0L1_datahier
+        """
+
+        print("test_field_coarsening_via_subcycles for dim/interp : {}/{}".format(ndim, interp_order))
+
+        from tests.amr.data.field.refine.test_refine_field import refine
+        from tests.core.numerics.interpolator.interpolator_test import interpolate_datasets
+        from pyphare.pharein import global_vars
+
+        time_step_nbr = 1 # after first advance, simulations diverge
```
nicolasaunai, line 261: "diverge" has a meaning (in numerical math) that the simulation goes crazy (results diverge exponentially in time). I know this is not what you mean here where simulations literally diverge from one another. Maybe we should add "from one another" in the comment.

Maybe we should say a bit more in the docstring, like saying that these two simulations are expected to have identical L0 for the first substep only before any coarsening occurs on the one that's refined, hence the only step that's simulated.

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,106 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+        coarsest_time_after = coarsest_time_before + lvlSteps[0]
+        assert_time_in_hier(coarsest_time_before, coarsest_time_after)
+
+        coarseBeforePatches = L0_datahier.level(0, coarsest_time_before).patches
+        coarseAfterPatches = L0_datahier.level(0, coarsest_time_after).patches
+        assert len(coarseBeforePatches) == len(coarseAfterPatches)
+
+        for coarsePatch_idx in range(len(coarseBeforePatches)):
+            coarseBeforePatch = coarseBeforePatches[coarsePatch_idx]
+            coarseAfterPatch  = coarseAfterPatches[coarsePatch_idx]
```
nicolasaunai, line 292: are you sure the patches are ordered similarly in time? If the code is executed in parallel I think you necessarily want to sort them by box.lower before using the same index...

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,106 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+            assert coarseBeforePatch.box == coarseAfterPatch.box
+
+            for finer_subcycle in range(0, global_vars.sim.level_step_nbr[1] + 1):
+
+                finer_subcycle_time = coarsest_time_before + (lvlSteps[1] * finer_subcycle)
+                assert_time_in_hier(finer_subcycle_time)
+                finer_subcycle_patches = L0L1_datahier.level(1, finer_subcycle_time).patches
+
+                for finer_subcycle_patch in finer_subcycle_patches:
+                    if boxm.coarsen(finer_subcycle_patch.box, refinement_ratio) in coarseBeforePatch.box:
```
nicolasaunai, line 302: what happens if a L1 patch overlaps 2 L0 patches , i.e.  is really within none entirely

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,106 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+                        for EM in ["E", "B"]:
+                            for xyz in ["x", "y", "z"]:
+                                qty = f"{EM}{xyz}"
+                                assert qty in coarseBeforePatch.patch_datas
+
+                                coarseBefore_pd = coarseBeforePatch.patch_datas[qty]
```
nicolasaunai, line 308: you can also define coarseAfter_pd, to simplify the call to interpolate_datasets later

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,106 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+                                    coarsest_time_before, coarsest_time_after, finer_subcycle_time,
+                                    coarseBefore_pd.dataset[:], coarseAfterPatch.patch_datas[qty].dataset[:])
+                                )
+
+                                finer_subcycle_pd = finer_subcycle_patch.patch_datas[qty]
+                                ghosts_nbr = finer_subcycle_pd.ghosts_nbr
+
+                                for finer_ghost_box in finer_subcycle_pd.ghost_box - finer_subcycle_patch.box:
+                                    is_upper = np.array(finer_ghost_box.lower > finer_subcycle_patch.box.lower, dtype=int)
+                                    finer_ghost_box = boxm.shift(finer_ghost_box, is_upper and coarseBefore_pd.primal_directions())
```
nicolasaunai, line 321: I don't understand that shift

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,106 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+                                    lvlOverlap = refinedInterpolatedField.box * finer_ghost_box
+
+                                    if lvlOverlap is not None:
+                                        finer_ghostbox_data = finer_subcycle_pd[finer_ghost_box]
+                                        refinedInterpGhostBox_data = refinedInterpolatedField[boxm.shift(finer_ghost_box, ghosts_nbr)]
+                                        np.testing.assert_allclose(finer_ghostbox_data, refinedInterpGhostBox_data, atol=1e-6)
```
nicolasaunai, line 327: so, when I see this 1e-6 I think oh ok that's because we dump in floats... but really in this kind of test we expect binary equality (I think?) so shouldn't we check for that?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -48,20 +48,44 @@ def x(self):
...
 
+    def select(self, box):
+        """ return view of internal data based on overlap of input box """
+        assert isinstance(box, Box) and box.ndim == self.box.ndim
+
+        gbox = self.ghost_box.copy()
+        gbox.upper += self.primal_directions()
+
+        overlap = box * gbox
+        assert overlap is not None # or return empty array?
```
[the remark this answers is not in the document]
PhilipDeegan: ok

tests/simulator/test_advance.py
PhilipDeegan: will remove

tests/simulator/test_advance.py
PhilipDeegan: ok

tests/simulator/test_advance.py
PhilipDeegan: currently it's dropped,

tests/simulator/test_advance.py
PhilipDeegan: not sure, but it does sound safer so will do

tests/simulator/test_advance.py
PhilipDeegan: can change it to an intersection instead

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -47,20 +47,43 @@ def x(self):
...
+        gbox = self.ghost_box.copy()
+        gbox.upper += self.primal_directions()
+
+        overlap = box * gbox
+        if overlap is not None:
+            lower = self.layout.AMRIndexToLocal(dim=box.ndim - 1, index=overlap.lower)
+            upper  = self.layout.AMRIndexToLocal(dim=box.ndim - 1, index=overlap.upper)
+            assert box.ndim == 1 # this following line is only 1D
+            return self.dataset[lower[0]:upper[0] + 1]
+        return []
```
nicolasaunai, line 80: returning np.array([]) would be more consistent

tests/amr/data/field/refine/test_refine_field.py
```diff
@@ -0,0 +1,35 @@
...
+from pyphare.core.gridlayout import GridLayout
+from pyphare.pharesee.hierarchy import FieldData
+
+
+# only supports ndim == 1, updates needed for > 1
+# assumes refinement_ratio is 2
+
+def refine(field, **kwargs):
+    assert isinstance(field, FieldData)
+    kwarg_keys = ["data"] # doing if data == None for input data=None, fails if it's an np array
```
nicolasaunai, line 13: I don't understand the comment

tests/amr/data/field/refine/test_refine_field.py
```diff
@@ -0,0 +1,35 @@
...
+    assert any([key in kwarg_keys for key in list(kwargs.keys())])
+
+    refinement_ratio = 2
+    primal_directions = field.primal_directions()
+    data   = kwargs.get("data", field.dataset[:])
```
nicolasaunai, line 18: at this point, it's not obvious to guess what's the goal of "data", in particular since its default value is the field's data. The function refines "data", which may come from kwargs or from the field's dataset, isn't it a bit weird? Why would you need to pass a field but don't use its data

tests/amr/data/field/refine/test_refine_field.py
```diff
@@ -0,0 +1,35 @@
...
+
+    fine_box    = boxm.refine(field.box, refinement_ratio)
+    fine_layout = GridLayout(fine_box, field.origin, field.layout.dl/refinement_ratio, interp_order=field.layout.interp_order)
+
+    assert field.box.ndim == 1
+
+    fine_data = np.zeros(fine_box.shape + primal_directions + (field.ghosts_nbr[0] * 2))
+
+    if primal_directions[0]:
+        fine_data[5:-5:2] = data[5:-5] # coarse primal on top of fine
```
nicolasaunai, line 29: I think you want to use the actual ghost nbr variable since you have it

tests/core/numerics/interpolator/interpolator_test.py
```diff
@@ -43,5 +43,15 @@ def main():
 
 
 
+def interpolate_datasets(beforeTime, afterTime, interp_time, beforeData, afterData):
+    alpha = (interp_time - beforeTime) / (afterTime - beforeTime)
+    assert beforeData.shape == afterData.shape
+    interpData = afterData.copy()
+    assert interpData.ndim == 1 # update for > 1d interpolation values
+    if interpData.ndim == 1:
```
nicolasaunai, line 51: the formula is valid for any dimension so this assert is superfluous

tests/core/numerics/interpolator/interpolator_test.py
```diff
@@ -43,5 +43,15 @@ def main():
 
 
 
+def interpolate_datasets(beforeTime, afterTime, interp_time, beforeData, afterData):
+    alpha = (interp_time - beforeTime) / (afterTime - beforeTime)
```
nicolasaunai, line 47: you probably want to assert afterTime > beforeTime and that interp_time is between the two, otherwise you'll get invalid alpha (can't interpolate outside time limits)

tests/core/numerics/interpolator/interpolator_test.py
```diff
@@ -43,5 +43,15 @@ def main():
 
 
 
+def interpolate_datasets(beforeTime, afterTime, interp_time, beforeData, afterData):
```
nicolasaunai, line 46: I think `time_interpolate()` makes more sense (not to be confused with spatial interpolation)

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,120 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
 
+
+
+    def _test_field_level_ghosts_via_subcycles_and_coarser_interpolation(self, ndim, interp_order, refinement_boxes, refinement_ratio=2):
+        """
+          This test runs two virtually identical simulations for one step.
+            L0_datahier has no refined levels
+            L0L1_datahier has one refined level
+
+          This is done to compare what would be the uncoarsened L0 values in L0L1_datahier
```
nicolasaunai, line 251: "uncoarsened L0 values" is misleading, since it's L1 values that are coarsened onto L0, and L0 domain nodes overlapped by L1 that are updated via coarsening....

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,120 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+            if it had no refined level to the level field ghost of L1 of L0L1_datahier
+
+          The simulations are no longer comparable after the first advance, so this test cannot work beyond that.
```
nicolasaunai, line 254: 👍

pyphare/pyphare/pharesee/geometry.py
```diff
@@ -283,7 +283,7 @@ def borders_per(box):
 
 
 
-def ghost_area_boxes(hierarchy, quantities):
+def ghost_area_boxes(hierarchy, quantities, levels=[], time=0):
```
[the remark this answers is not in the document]
PhilipDeegan: sure will change

pyphare/pyphare/pharesee/geometry.py
```diff
@@ -297,9 +297,14 @@ def ghost_area_boxes(hierarchy, quantities):
...
     """
+    levels = listify(levels)
+    if len(levels) == 0:
+        levels = list(hierarchy.levels(time).keys())
+
     gaboxes = {}
 
-    for ilvl, lvl in hierarchy.levels().items():
+    for ilvl in levels:
+        lvl = hierarchy.level(ilvl, time)
```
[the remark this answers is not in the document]
PhilipDeegan: given the input list of levels, this is possibly more than levels than requested

tests/amr/data/field/refine/test_refine_field.py
PhilipDeegan:
```
def refine(field, data=None):
       if data is None: # this fails if data is an np array
```

tests/amr/data/field/refine/test_refine_field.py
PhilipDeegan: the field is for the other values, the "data" kwarg is an override in case you want to use another dataset, like in this PR, the time interpolated dataset

tests/simulator/test_advance.py
PhilipDeegan: "L0 values that haven't received the coarsened values of L1 because there is no L1"

tests/amr/data/field/refine/test_refine_field.py
PhilipDeegan: should the ":2" be refinement_factor or maybe "cadence = 2" or something?

tests/core/numerics/interpolator/interpolator_test.py
PhilipDeegan: and the "if" too

tests/core/numerics/interpolator/interpolator_test.py
PhilipDeegan:
```
    assert before_time < after_time
    assert before_time <= interp_time <= after_time
```

"before_time" must be less than "after_time"
but "interp_time" can be equal to either, or between

tests/amr/data/field/refine/test_refine_field.py
nicolasaunai: hmm yeah it's because of the refinement ratio, but if we were using a different ratio the formula would be different as well, no just that slicing...

pyphare/pyphare/pharesee/geometry.py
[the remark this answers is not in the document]
nicolasaunai: indeed ok.

tests/amr/data/field/refine/test_refine_field.py
nicolasaunai: and it's not possible to "wrap" the refined/time interpolated dataset into a FieldData based on the original, so to have a single data source?

tests/core/numerics/interpolator/interpolator_test.py
nicolasaunai: interp_time can be equal to before and after times, that means alpha 0 or 1, respectively, indeed. However checking equality with floats is tricky so you want to check if it's not < before and not > after, I guess

tests/amr/data/field/refine/test_refine_field.py
PhilipDeegan: it is possible, seems like overkill to make a temporary field just to refine different data

tests/core/numerics/interpolator/interpolator_test.py
PhilipDeegan: how bout something like this 
image

tests/amr/data/field/refine/test_refine_field.py
```diff
@@ -0,0 +1,82 @@
...
+    ghostX = field.ghosts_nbr[0]
+    assert ghostX > 0
+
+    cadence = 2
+
+    if primal_directions[0]:
+        fine_data[ghostX:-ghostX:cadence] = data[ghostX:-ghostX] # coarse primal on top of fine
+        fine_data[ghostX + 1:-(ghostX - 1):cadence] = 0.5*data[ghostX:-ghostX] + 0.5*data[ghostX + 1:-(ghostX - 1)]
+    else:
+        fine_data[ghostX:-(ghostX + 1):cadence] = 0.25*data[4:-(ghostX + 1)] + 0.75*data[ghostX:-ghostX]
```
nicolasaunai, line 42: you missed a 4 :-D

nicolasaunai: I'm sure you did that on purpose to see if I really read your code :p

PhilipDeegan: not this time :(

tests/core/numerics/interpolator/interpolator_test.py
```diff
@@ -43,5 +43,13 @@ def main():
 
 
 
+def time_interpolate(before_time, after_time, interp_time, before_data, after_data):
+    assert before_time < after_time
+    assert before_time <= interp_time <= after_time
```
nicolasaunai, line 48: assert after_time>interp_time and before_time<interp_time
?

tests/simulator/test_advance.py
```diff
@@ -159,13 +163,12 @@ def _test_field_coarsening_via_subcycles(self, dim, interp_order, refinement_box
                                       time_step_nbr=time_step_nbr, smallest_patch_size=5,
                                       largest_patch_size=30)
 
-        levelNumbers = list(range(global_vars.sim.max_nbr_levels))
-        lvlSteps = global_vars.sim.level_time_steps
-        print("LEVELSTEPS === ", lvlSteps)
-        assert len(lvlSteps) > 1  # this test makes no sense with only 1 level
+        lvl_steps = global_vars.sim.level_time_steps
+        print("LEVELSTEPS === ", lvl_steps)
+        assert len(lvl_steps) > 1  # this test makes no sense with only 1 level
```
nicolasaunai, line 168: you could put the comment as the assert message?

tests/simulator/test_advance.py
```diff
@@ -237,6 +240,94 @@ def test_field_coarsening_via_subcycles(self, refinement_boxes):
...
+        checks = 0
+        for fine_subcycle_time in fine_subcycle_times:
+            fine_level_qty_ghost_boxes = level_ghost_boxes(L0L1_datahier, quantities, fine_ilvl, fine_subcycle_time)
+            for qty in quantities:
+                for fine_level_ghost_box_data in fine_level_qty_ghost_boxes[qty]:
+                    fine_subcycle_pd = fine_level_ghost_box_data["pdata"]
+                    for fine_level_ghost_box in fine_level_ghost_box_data["boxes"]:
+                        for refinedInterpolatedField in interpolated_fields[qty][fine_subcycle_time]:
+                            lvlOverlap = refinedInterpolatedField.box * fine_level_ghost_box
+                            if lvlOverlap is not None:
```
nicolasaunai, line 310: can it be None? if not should we assert if it is? A priori if a box is a level ghost it *must* overlap the refined L0, no?

PhilipDeegan: it can be none if there's multiple L0 patches, when running under MPI for example

nicolasaunai: right....

pyphare/pyphare_tests/test_pharesee/test_geometry.py
```diff
@@ -371,5 +372,25 @@ def test_level_ghostboxes(self, refinement_boxes, expected):
...
+
+            pdata = hierarchy.level(0).patches[0].patch_datas[qty]
+            lower_gb, upper_gb = sorted(pdata.ghost_box - pdata.box, key=lambda box: box.lower.all())
+            lower_ds, upper_ds = pdata[lower_gb], pdata[upper_gb]
+
+            qty_is_primal = yee_element_is_primal(qty)
+            assert lower_ds.shape[0] == pdata.ghosts_nbr[0] + qty_is_primal
+            assert upper_ds.shape[0] == pdata.ghosts_nbr[0] + qty_is_primal
+
+            np.testing.assert_allclose(lower_ds, pdata.dataset[:lower_ds.shape[0]], atol=1e-12)
```
nicolasaunai, line 390: shouldn't we do 
```
np.all(array1 == array2)
```
if it's the same data it's supposed to be stricly equal no?

nicolasaunai (approved):
I think it's good now. to squash and merge?

Issue #468, opened 2021-03-17
Have dict arguments homogenized 


nicolasaunai:
various places of the code may use dict arguments as copies, or ref or const ref or rvalue (move).
This issue is about having it homogenized, a priori we do not need copying or moving.
We can subclass dict with a deleted move Ctor to be sure there are no branches cut from the tree

PR #477, opened 2021-03-18
const ref Dict passing rather than copy/non-const ref 

PhilipDeegan:
for less atomic shared_ptr counter ref increments ~~- tests is bit trickier - may need PR in dict lib~~


PhilipDeegan:
for https://github.com/PHAREHUB/PHARE/issues/468

nicolasaunai (approved):

PR #456, opened 2021-03-10
distribution plots with finest data

nicolasaunai:
#449

PR #478, opened 2021-03-18
refactor finest data

nicolasaunai:
(on top of #456 )

Previously finest_data was working per patch_data. This forced the client code to loop over the hierarchy etc.
This new version is more convenient, you just pass a field hierarchy and it 

example of usage 

```python
r007 = Run("../phare_jobs/tests/alfven_wave/run007/")
B = r007.GetB(0)
bx, xbx = finest_field(B, "Bx")
print(np.allclose(xbx, amr_grid(B, 0)))
by, xby = finest_field(B, "By")
fig,ax = plt.subplots(figsize=(10,5))
ax.plot(xb, bb, marker="x", markersize=5)
```

returns

```
True
```

image

this simplify functions used to plot stuff like (same simulation, hierarchy patches to be compared with marker density above)

image

where one had to loop over the patches etc. and call former finest_data() per patch as in : 

```python
for ilvl, lvl in self.patch_levels.items():
        for pidx, patch in enumerate(lvl.patches):
            pdata  = patch.patch_datas[qty]
            x,v = finest_data(pdata, ilvl, self)
            ax1.plot(x, v,label="lev{} - patch{}".format(ilvl, pidx),
                     marker='o', markersize=2, color=colors[ilvl])

```

moreover we now have a contiguous array for data, which is convenient.

### DIFF
--- a/pyphare/pyphare/pharein/diagnostics.py
+++ b/pyphare/pyphare/pharein/diagnostics.py
@@ -52,8 +52,7 @@ def validate_timestamps(clazz, **kwargs):
             raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
         if not np.all(np.diff(timestamps) >= 0):
             raise RuntimeError(f"Error: {clazz}.{key} not in ascending order)")
-        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):
-            print(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step)).max())
+        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):            
             raise RuntimeError(f"Error: {clazz}.{key} is inconsistent with simulation.time_step")
 
 
--- a/pyphare/pyphare/pharein/diagnostics.py
+++ b/pyphare/pyphare/pharein/diagnostics.py
@@ -52,8 +52,9 @@ def validate_timestamps(clazz, **kwargs):
             raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
         if not np.all(np.diff(timestamps) >= 0):
             raise RuntimeError(f"Error: {clazz}.{key} not in ascending order)")
-        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-10)):
-            raise RuntimeError(f"Error: {clazz}.{key} is inconsistent with simulation.time_step)")
+        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):
+            print(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step)).max())
+            raise RuntimeError(f"Error: {clazz}.{key} is inconsistent with simulation.time_step")
 
 
 
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -178,8 +178,8 @@ class PatchLevel:
 
 
 
-def are_adjacent(lower_x, upper_x, atol=1e-6):
-    return np.abs(upper_x[0]-lower_x[-1]) < atol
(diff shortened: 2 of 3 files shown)


pyphare/pyphare/pharein/diagnostics.py
```diff
@@ -52,8 +52,9 @@ def validate_timestamps(clazz, **kwargs):
                 raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
         if not np.all(np.diff(timestamps) >= 0):
             raise RuntimeError(f"Error: {clazz}.{key} not in ascending order)")
-        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-10)):
-            raise RuntimeError(f"Error: {clazz}.{key} is inconsistent with simulation.time_step)")
+        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):
+            print(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step)).max())
```
PhilipDeegan, line 56: needed? or maybe incorporated into the error string?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -179,42 +179,108 @@ def level_range(self):
 
 
 
-def finest_data(pdata, ilvl, hierarchy):
+def are_adjacent(lower_x, upper_x, atol=1e-6):
+    return np.abs(upper_x[0]-lower_x[-1]) < atol
```
PhilipDeegan, line 183: maybe? 

```
def are_adjacent(lower, upper, atol=1e-6):
    return np.abs(upper-lower) < atol
```

and client callers should do index access I think,

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -179,42 +179,108 @@ def level_range(self):
...
+
+
+
+def finest_field(hierarchy, qty, time=None):
+    """returns a non-uniform contiguous (value,grid)
+       associated to the given hierarchy
+       the quantity "qty" must be a field
+    """
+    if time is None:
+        time = hierarchy.times()[0] # to replace with default
```
PhilipDeegan, line 226: can probably just rm these lines, hierarchy handles "time=None" internally

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -179,42 +179,108 @@ def level_range(self):
...
+
+    lvl = hierarchy.levels(time)
+
+    for ilvl in range(hierarchy.finest_level(time)+1)[::-1]:
+        sorted_patches = sorted(lvl[ilvl].patches,
+                                key= lambda p:p.layout.box.lower[0])
```
PhilipDeegan, line 232: key= lambda p:p.box.lower[0])

pyphare/pyphare/pharein/diagnostics.py
nicolasaunai: it's a leftover from debugging... the 1e-10 was failing, had to put it to 1e-9... not sure why

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: ok, will change

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: you're probably right, I'll change it if that's true

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: right..

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: actually thinking about it, no. the point of that funciton in my opinion is actually that we do not have to think about that logic that makes two sets of coordinates adjacents. If I put upper[0] and lower[-1] at call site, then the function is almost worthless

PhilipDeegan: "math.isclose" could be replacing it

nicolasaunai: ok I removed the _x as discussed

PhilipDeegan (approved):

PR #485, opened 2021-03-22
patch ghost advance tests

PhilipDeegan:


PhilipDeegan:
remove the deprecated init tests now, or wait till these changes are reviewed?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -1076,7 +1076,7 @@ def hierarchy_from(simulator=None, qty= None, pop = "", h5_filename=None, time=N
 def merge_particles(hierarchy):
 
     for time, patch_levels in hierarchy.time_hier.items():
-        for ilvl, plvl in hierarchy.patch_levels.items():
+        for ilvl, plvl in patch_levels.items():
```
nicolasaunai, line 1079: maybe we should rather use the levels() method? (if we deprecate patch_levels soonish)

tests/simulator/test_advance.py
```diff
@@ -143,6 +168,258 @@ def vthz(x):
 
 
 
+    def _test_overlaped_fields_are_equal(self, time_step, time_step_nbr, datahier):
```
nicolasaunai, line 171: that's a copy/paste from initialization test, do we want that?

nicolasaunai (commented):
I would think that for those methods that exist in the initialization test that are repeated here, those here also test the initialization (t=0) no? thus I would remove them in the init test. If not I would like to understand why, and how to avoid possible duplication (of not so trivial code btw)

tests/simulator/test_advance.py
PhilipDeegan: copy/paste/change
this one has advances, and I would say deprecates the initialization version, which I haven't removed but can as an addendum to this PR

nicolasaunai: yes I would remove all initialization tests that are redundant with these advance tests

PhilipDeegan: commit/pushed/building

nicolasaunai (approved):
squash & merge?

PR #490, opened 2021-03-24
Alfven

nicolasaunai:
on top of #478 

This PR updates the Alfven wave functionnal test to the configuration we use in the paper.
We produce the following figure

image

calculate the phase speed of the wave as assert it is close enough from it's theoretical value V=1.

Note : the test runs 1 million time step and took many hours on 8 cores..... so we should discuss how that one gets run on nightly builds....

### DIFF
--- a/tests/functional/alfven_wave/alfven_wave1d.py
+++ b/tests/functional/alfven_wave/alfven_wave1d.py
@@ -16,20 +16,27 @@ mpl.use('Agg')
 
 
 
+####################################################################
+#
+#                     Simulation configuration
+#
+####################################################################
 def config():
 
     # configure the simulation
 
     Simulation(
-        smallest_patch_size=20,
-        largest_patch_size=20,
-        time_step_nbr=10000,        # number of time steps (not specified if time_step and final_time provided)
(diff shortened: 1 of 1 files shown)


nicolasaunai:
after this PR is merged, we need to add the "alfven_wave.png" file it produces to TC artifacts

tests/functional/alfven_wave/alfven_wave1d.py
```diff
@@ -118,54 +123,95 @@ def vthz(x):
...
+        wave_vec[it]  = k
 
+    vphi = medfilt(np.gradient(phase, time)/wave_vec, kernel_size=7)
+    return vphi, time, phase, amplitude, wave_vec
 
 
+def get_times(path):
+    import h5py
+    f = h5py.File(path, 'r')
+    times = np.array(sorted([float(s.strip("t")) for s in list(f.keys())]))
```
PhilipDeegan, line 158: https://github.com/PHAREHUB/PHARE/pull/494 will break this, maybe refactor to hierarchy.py as

"times_from_h5_file(path):" / "h5_file_times(path)"

can dedupe conserv.py too

Issue #464, opened 2021-03-15
Have Run.GetXXX return either hierarchies or ndarrays with amr coordinates


nicolasaunai:
For now Run::GetXXX() functions only return hierarchies. It would be convenient to have them return ndarrays of finest data with associated AMR grids. It could be chosen as a keyword option.

PR #492, opened 2021-03-24
Run

nicolasaunai:
#464 

On top of #490  and #478

### DIFF
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -409,6 +409,26 @@ class PatchHierarchy:
     def levelNbrs(self, time):
         return list(self.levels(time).keys())
 
+    def is_homogeneous(self):
+        """
+        return True if all patches of all levels at all times
+        have the same patch data quantities
+        """
+        qties = self._quantities()
+        it_is = True
+        for time, levels in self.time_hier.items():
+            for ilvl, lvl in levels.items():
+                for patch in lvl.patches:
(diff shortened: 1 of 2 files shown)


PhilipDeegan:
LGTM is giving a different reason for the "x != x" bit now + one unused var

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -409,6 +409,25 @@ def levelNbr(self, time):
...
+            for ilvl, lvl in levels.items():
+                for patch in lvl.patches:
+                    it_is &= qties == list(patch.patch_datas.keys())
+        return it_is
+
+    def _quantities(self):
+        return list(self.level(0).patches[0].patch_datas.keys())
+
+    def quantities(self):
+        assert self.is_homogeneous()
```
PhilipDeegan, line 429: this doesn't exist for "python -O" executions

maybe not an issue

nicolasaunai: I don't understand what you mean

PhilipDeegan: should it be a "raise RuntimeError" or an assert. as the assert is disabled when running with "python3 -O script.py"

PhilipDeegan (approved):

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: ok I didn't know that. I'll change it to a raise RunTimeError then

PR #494, opened 2021-03-24
Python3 diagnostic attributes

PhilipDeegan:


pyphare/pyphare/pharein/diagnostics.py
```diff
@@ -79,6 +86,15 @@ def __init__(self,name, **kwargs):
         self.write_timestamps = kwargs['write_timestamps'] #[0, 1, 2]
         self.compute_timestamps = kwargs['compute_timestamps']
 
+        self.attributes = kwargs.get("attributes", {})
+        self.attributes["git_hash"]  = phare_utilities.top_git_hash()
+
+        for dep, dep_ver in Diagnostics.cpp_dep_vers.items():
+            self.attributes["dep_" + dep] = dep_ver
+
+        for key in self.attributes: # avoid conflict with other root attributes
```
PhilipDeegan, line 95: these are no longer at the h5 root group

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -762,6 +760,9 @@ def patch_has_datasets(h5_patch_grp):
     return len(h5_patch_grp.keys())>0
 
 
+
+h5_time_grp_key = "t"
```
PhilipDeegan, line 764: maybe should be exposed via pybind for isolation

res/cmake/dep/highfive.cmake
```diff
@@ -25,7 +25,11 @@ if(HighFive)
 
   include_directories(
     ${HIGHFIVE_SRC}/include
+    ${CMAKE_BINARY_DIR}/subprojects/highfive/include # configured include for version info
   )
+  set(HIGHFIVE_USE_BOOST OFF)
+  set(HIGHFIVE_BUILD_DOCS OFF) # conflicts with phare doc target
+  add_subdirectory(${HIGHFIVE_SRC})
```
PhilipDeegan, line 32: highfive has a cmake configured file containing its version info, so we have to do this

src/core/utilities/types.h
```diff
@@ -18,6 +18,8 @@
 #define PHARE_DEBUG_DO(...)
 #endif
 
+#define _PHARE_TO_STR(x) #x // convert macro text to string
+#define PHARE_TO_STR(x) _PHARE_TO_STR(x)
```
PhilipDeegan, line 22: this functionality might exist somewhere we have access to if we don't want to manage it ourselves

src/diagnostic/diagnostic_manager.h
```diff
@@ -143,6 +143,15 @@ DiagnosticsManager<Writer>::addDiagDict(PHARE::initializer::PHAREDict const& dia
     diagProps.computeTimestamps
         = diagInputs["compute_timestamps"].template to<std::vector<double>>();
 
+    diagProps.nAttributes = diagInputs["n_attributes"].template to<std::size_t>();
+    for (std::size_t i = 0; i < diagProps.nAttributes; ++i)
+    {
+        std::string idx = std::to_string(i);
+        std::string key = diagInputs["attribute_" + idx + "_key"].template to<std::string>();
+        std::string val = diagInputs["attribute_" + idx + "_value"].template to<std::string>();
```
PhilipDeegan, line 151: I was wondering if there was a way to just set the key to a dict node without need to extract it, maybe it's possible, in that case we wouldn't need to enforce all values to be strings

nicolasaunai (approved):

pyphare/pyphare/core/phare_utilities.py
```diff
@@ -81,3 +81,27 @@ def check_mandatory_keywords(mandatory_kwd_list, **kwargs):
...
+
+def git_hashes(N=1):
+    return decode_bytes(run_cli_cmd(f"git log -{N} --pretty=format:%h").stdout).splitlines()
+
+
+def top_git_hash():
+    hashes = git_hashes(1)
+    if len(hashes) > 0:
+        return hashes[0]
+    return "master" # github actions fails?
```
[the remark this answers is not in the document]
PhilipDeegan: trying to get the git hash fails on github actions, or it can fail if there is no git info / git binary

PR #495, opened 2021-03-26
Caliper as cmake dep 

PhilipDeegan:
+ logging additions to various functions for active logging option


nicolasaunai (commented):
I think we should add a log in the multiphysics integrator regrid method
possibly also into field/particle Data copy/stream (the override from PatchData virtual) functions, so to discriminate SAMRAI internal from our stuff, during regrid times

nicolasaunai (approved):

nicolasaunai:
squash/merge

PR #508, opened 2021-03-31
use explicit dict population functions 

PhilipDeegan:
to force casts to the correct C++ usage types / +1 IWYU


PhilipDeegan:
seems without casts at during "add_$type", we can get pybind errors

proposal to change:

```
add_int = pp.add_int
```

to 

```
def add_int(path, val):
    pp.add_int(path, int(val))
```

strings can be cast to ints, but may not be valid so errors are throw in that case from python

nicolasaunai (approved):
looks good, merge if happy

PR #511, opened 2021-03-31
wip smoothing

nicolasaunai:
This is WIP

This PR adds the possibility to smooth the density. This helps reducing the noise mainly originating from the pressure gradient term in Ohm's law.
The way it is currently done is not ideal. Improvements can be

- let the Filter object declare a resource for a temporary density array, rather than allocating one on the fly, for each patch, each level, each pass. This would reduce the cadence of the memory allocation, basically to 1 per regrid.

- have configurable (via dict/python) number of filter passes. Not super important.

- The filter operates on a 3 points stencil, which means that border nodes need the first primal ghost. Before this, the first primal ghost was the copy of the border node, thus not equal to the overlaped neighbor domain primal node. This was not an issue since note used for domain computations (only for coarsening, which was OK). Now this ghost needs to be known. This means that a ghost exchange SAMRAI schedule is applied before filtering and at each filter pass.

- Currently, the filter passes are applied per level. This means that there is a potential "disconnection" at level boundaries. Ideally each pass of the filter should do the entire hierarchy, echange ghosts for each level, etc. for the level boundaries to stay in sync with other levels. Not doing that makes the level border mismatch only visible for high smoothing (large number of passes).

### DIFF
--- a/src/core/numerics/filter/filter.h
+++ b/src/core/numerics/filter/filter.h
@@ -14,19 +14,34 @@ public:
     {
         auto start = layout.physicalStartIndex(density, Direction::X);
         auto end = layout.physicalEndIndex(density, Direction::X);
-        auto alpha = 0.5;
+
+        // this is hard coded but probably will never change
+        auto constexpr alpha = 0.5;
+
+        // this is doing a memory allocation
+        // which is not optimal since that function will be called
+        // several time per patch. Another way would be that
+        // the filter is owned by the Solver and has an interface
+        // used by the solver to declare a resource to the resource manager
+        // for that temporary object. This way the allocation would be done
+        // for all patches only once per regrid. This is something that should
+        // be done eventually.
+        std::vector<typename Field::type> tmp{density.data(), density.data()+density.size()};
+
         if constexpr (Field::dimension == 1)
         {
             for (auto ix=start; ix <= end; ++ix)
             {
-                density(ix) = alpha*density(ix) + 0.5*(1-alpha)*(density(ix-1)+ density(ix+1));
+                // also the fact that 'tmp' is a vector (and not a field)
+                // forces us to use [] and not (), which is ok in 1D but much less
+                // extensible to nD than if we had one tmp Field registered.
+                density(ix) = alpha*tmp[ix] + 0.5*(1-alpha)*(tmp[ix-1]+ tmp[ix+1]);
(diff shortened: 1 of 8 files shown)


src/core/numerics/filter/filter.h
```diff
@@ -0,0 +1,52 @@
...
+        // for that temporary object. This way the allocation would be done
+        // for all patches only once per regrid. This is something that should
+        // be done eventually.
+        std::vector<typename Field::type> tmp{density.data(), density.data()+density.size()};
+
+        if constexpr (Field::dimension == 1)
+        {
+            for (auto ix=start; ix <= end; ++ix)
+            {
+                // also the fact that 'tmp' is a vector (and not a field)
```
PhilipDeegan, line 35: we have an NdArrayView for this 

```
https://github.com/PHAREHUB/PHARE/blob/master/tests/diagnostic/test_diagnostics.h#L61
```

src/solver/solvers/solver_ppc.h
```diff
@@ -474,6 +480,34 @@ void SolverPPC<HybridModel, AMR_Types>::average_(level_t& level, HybridModel& mo
...
+                                                       Messenger& messenger, double newTime)
+{
+    auto& ions             = model.state.ions;
+    auto& resourcesManager = model.resourcesManager;
+    PHARE::core::Filter filter;
+    auto levelNumber = level.getLevelNumber();
+    messenger.fillDensityGhosts(levelNumber, newTime);
+
+
+    for (auto i = 0; i < 2; ++i)
```
PhilipDeegan, line 494: should this be configurable?

or perhaps

```
constexpr std::size_t filter_passes = 2; 
```

src/solver/solvers/solver_ppc.h
```diff
@@ -474,6 +480,34 @@ void SolverPPC<HybridModel, AMR_Types>::average_(level_t& level, HybridModel& mo
...
+    {
+        for (auto& patch : level)
+        {
+            auto layout = PHARE::amr::layoutFromPatch<GridLayout>(*patch);
+            auto _      = resourcesManager->setOnPatch(*patch, ions);
+
+            auto& density = ions.density();
+            filter(density, layout);
```
PhilipDeegan, line 502: needs two lines?

PR #516, opened 2021-04-07
caliper / uniform patch_size bench suite

PhilipDeegan:


PhilipDeegan:
this currently puts all logs in the artifact output directory "data_out"
the logs are possibly quite large, so the working directory for these logs might want to be changed to "data_tmp" or something so that they are usable for graphs/etc but are not included in artifacts

tools/bench/functional/test_cases/uniform_w_patch_variance.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+### test permutation section - minimized is best ###
+dl = 0.25
+final_time = 1
+mpirun_Ns = [1, 10]
+patch_sizes = [5, 10, 50, 100, 200, 400, 800, 1600]
+time_step = .01
+cells = 1600
+ppc_list = [111, 1111, 2222, 3333]
```
PhilipDeegan, line 23: possibly overkill

PhilipDeegan:
assess python run_mp not causing test failure on CI
kwargs "check" appears to be ignored or dropped somehow

PhilipDeegan:
moved to https://github.com/PhilipDeegan/PHARE/pull/86

PR #523, opened 2021-04-10
fix conserv.py broken by #520

PhilipDeegan:


nicolasaunai (approved):
we need to add the figure that's created to the artifacts of nightly builds

PR #527, opened 2021-04-12
jupyter investigation to close any open h5 file handles so not to reuse old data

PhilipDeegan:


PhilipDeegan:
if you feel like testing this @nicolasaunai maybe it will resolve https://github.com/PHAREHUB/PHARE/issues/526

pyphare/pyphare/core/ipython.py
```diff
@@ -0,0 +1,55 @@
...
+        if h5_weak_ref["ref"]() is None:
+            del h5_weak_refs[h5_path]
+
+# https://stackoverflow.com/a/40222538/795574
+try:
+    def exit_register(fun, *args, **kwargs):
+        def callback():
+            fun()
+            ip.events.unregister('post_execute', callback)
+        ip.events.register('post_execute', callback)
```
nicolasaunai, line 19: probably not indented correclty?

nicolasaunai:
> after reading the PR I wonder: isn't the problem actually that we never close the files explicitly? when reading with hierarchy_fromh5, there is never a `data_file.close()`

ok probably not a good idea.
calling data_file.close() will make all our dataset become `<Closed HDF5 dataset>`  and the only way to keep our data would then be to really read it, i.e. not storing the dataset like [here](https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharesee/hierarchy.py#L1277) but storing the `dataset[:]` which may loose the nice if/when h5 offers us... (not sure I'm clear?)

pyphare/pyphare/core/ipython.py
PhilipDeegan: It's been some time since looking at this, and I'm pretty sure I had another later branch with modifications which I never added so it's quite likely

PhilipDeegan:
> after reading the PR I wonder: isn't the problem actually that we never close the files explicitly? when reading with hierarchy_fromh5, there is never a `data_file.close()`

instead of closing the object, we could try to just reset it with a new `handle = h5open(...)`

and if that doesn't work we should probably ask the h5py people

PR #517, opened 2021-04-08
change particle delta from float to double

nicolasaunai:
initialization test crashes with 
```
[pharemer:4042883] [ 0] /lib64/libc.so.6(+0x3ca70)[0x7f5e7caf5a70]
[pharemer:4042883] [ 1] /lib64/libc.so.6(gsignal+0x145)[0x7f5e7caf59e5]
[pharemer:4042883] [ 2] /lib64/libc.so.6(abort+0x127)[0x7f5e7cade895]
[pharemer:4042883] [ 3] /lib64/libc.so.6(+0x80857)[0x7f5e7cb39857]
[pharemer:4042883] [ 4] /lib64/libc.so.6(+0x87d7c)[0x7f5e7cb40d7c]
[pharemer:4042883] [ 5] /lib64/libc.so.6(+0x89452)[0x7f5e7cb42452]
[pharemer:4042883] [ 6] /usr/lib64/python3.8/site-packages/numpy/core/_multiarray_umath.cpython-38-x86_64-linux-gnu.so(+0x2d868)[0x7f5e6d624868]
[pharemer:4042883] [ 7] /usr/lib64/python3.8/site-packages/numpy/core/_multiarray_umath.cpython-38-x86_64-linux-gnu.so(+0x31513)[0x7f5e6d628513]
[pharemer:4042883] [ 8] /home/aunai/Documents/code/phare/build-PHARE-mpi-Release/pybindlibs/cpp.cpython-38-x86_64-linux-gnu.so(+0x489560)[0x7f5e6ed5f560]
[pharemer:4042883] [ 9] /home/aunai/Documents/code/phare/build-PHARE-mpi-Release/pybindlibs/cpp.cpython-38-x86_64-linux-gnu.so(+0x1de3f0)[0x7f5e6eab43f0]
[pharemer:4042883] [10] /lib64/libpython3.8.so.1.0(+0x12064a)[0x7f5e7c8a464a]
[pharemer:4042883] [11] /lib64/libpython3.8.so.1.0(_PyObject_MakeTpCall+0xe1)[0x7f5e7c894c41]
[pharemer:4042883] [12] /lib64/libpython3.8.so.1.0(_PyEval_EvalFrameDefault+0x5171)[0x7f5e7c891581]
[pharemer:4042883] [13] /lib64/libpython3.8.so.1.0(+0x116777)[0x7f5e7c89a777]
[...]

PR #529, opened 2021-04-12
dbg pybind cpp lib / double diags exposed by dbg pybind cpp lib

PhilipDeegan:
includes https://github.com/PHAREHUB/PHARE/pull/517

if cmake is configured via CMAKE_BUILD_TYPE=Debug, or not configured with CMAKE_BUILD_TYPE (which defaults to Debug), two pybind libraries are built, one with existing functionality, and one with extra debug info, in this PR that includes diagnostic quantities being written in double precision, for double values.

in python, we check to see if we're in \_\_debug\_\_ ,  if not (python -O), we return the normal cpp lib, otherwise we attempt to load the debug library, falling back on the regular library if debug is not present.

### DIFF
--- a/pyphare/pyphare/cpp/__init__.py
+++ b/pyphare/pyphare/cpp/__init__.py
@@ -1,17 +1,21 @@
 
 
+def cpp_lib():
+    import importlib
+    if not __debug__:
+        return importlib.import_module("pybindlibs.cpp")
+    try:
+        return importlib.import_module("pybindlibs.cpp_dbg")
+    except ImportError as err:
+        return importlib.import_module("pybindlibs.cpp")
+
+
 def splitter_type(dim, interp, n_particles):
-    from pybindlibs import cpp
-    return getattr(cpp,
-        "Splitter_" + str(dim) + "_" + str(interp)+ "_" + str(n_particles),
-    )
+    return getattr(cpp_lib(), f"Splitter_{dim}_{interp}_{n_particles}")
 
 def create_splitter(dim, interp, n_particles):
     return splitter_type(dim, interp, n_particles)()
 
 
 def split_pyarrays_fn(dim, interp, n_particles):
-    from pybindlibs import cpp
-    return getattr(cpp,
-        "split_pyarray_particles_" + str(dim) + "_" + str(interp)+ "_" + str(n_particles),
-    )
+    return getattr(cpp_lib(), f"split_pyarray_particles_{dim}_{interp}_{n_particles}")
--- a/src/python3/CMakeLists.txt
+++ b/src/python3/CMakeLists.txt
@@ -11,3 +11,14 @@ set_target_properties(cpp
 )
 # this is on by default "pybind11_add_module" but can interfere with coverage so we disable it if coverage is enabled
 set_property(TARGET cpp PROPERTY INTERPROCEDURAL_OPTIMIZATION ${PHARE_INTERPROCEDURAL_OPTIMIZATION})
+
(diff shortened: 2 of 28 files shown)


src/python3/CMakeLists.txt
```diff
@@ -11,3 +11,14 @@ set_target_properties(cpp
 )
 # this is on by default "pybind11_add_module" but can interfere with coverage so we disable it if coverage is enabled
 set_property(TARGET cpp PROPERTY INTERPROCEDURAL_OPTIMIZATION ${PHARE_INTERPROCEDURAL_OPTIMIZATION})
+
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  pybind11_add_module(cpp_dbg cpp_simulator.cpp)
+  target_link_libraries(cpp_dbg PUBLIC phare_simulator)
+  target_compile_options(cpp_dbg PRIVATE ${PHARE_FLAGS} -DPHARE_HAS_HIGHFIVE=${PHARE_HAS_HIGHFIVE} -DPHARE_DIAG_DOUBLES=1 -DPHARE_CPP_MOD_NAME=cpp_dbg)
```
PhilipDeegan, line 18: "-DPHARE_DIAG_DOUBLES=1" should probably be "-DPHARE_DEBUG_MODE=1"

and either "PHARE_DIAG_DOUBLES" is defined to true/1 if PHARE_DEBUG_MODE is active, or drop PHARE_DIAG_DOUBLES entirely and just use PHARE_DEBUG_MODE to pick between float/double in h5writer.h "using FloatType = std::conditional_t<PHARE_DIAG_DOUBLES, double, float>;"

pyphare/pyphare/cpp/__init__.py
```diff
@@ -1,17 +1,21 @@
 
 
+def cpp_lib():
+    import importlib
+    if not __debug__:
+        return importlib.import_module("pybindlibs.cpp")
+    try:
+        return importlib.import_module("pybindlibs.cpp_dbg")
+    except ImportError as err:
+        return importlib.import_module("pybindlibs.cpp")
```
nicolasaunai, line 10: if I understand correctly:, we could run in debug, but for some reason can't improt the debug cpp version then we'd "silently" fallback on the regular one? do we want that to be silent?

PhilipDeegan: python doesn't know what mode cmake was built with, this assumes if there's no dbg library, cmake wasn't built in debug

src/python3/CMakeLists.txt
nicolasaunai: I think we might, for some reason, want to have the option of using double diags beyond debug mode so I'd keep that option, and yes keep setting it to true when in debug also, is useful

pyphare/pyphare/cpp/__init__.py
nicolasaunai: what's this variable then __debug__

nicolasaunai: oh that's the python -O probably?

src/python3/CMakeLists.txt
PhilipDeegan: "__debug__" is false when exectuting with "python -O script.py"

pyphare/pyphare/cpp/__init__.py
PhilipDeegan: "\_\_debug\_\_" is false when exectuting with "python -O script.py"

PR #533, opened 2021-04-16
fix alfven wave imports/etc

PhilipDeegan:
maybe there's something we could do for functional tests to just run them with a single timestep or something so that we can verify syntax/correctness easily


nicolasaunai:
We'll need a "script validator" at some point that runs a fake simulation or something

nicolasaunai (approved):

PR #543, opened 2021-04-29
PGI C++ compiling/test passing

PhilipDeegan:


PhilipDeegan:
gtest define macro causes issues

https://github.com/google/googletest/issues/3390 

https://github.com/PHAREHUB/PHARE/pull/543/files#diff-aef27c28df873d118a95e44f080da9ef6175a83ca12da58df036c85d993e5724R25

PhilipDeegan:
highfive cmake config issue with PGI https://github.com/BlueBrain/HighFive/issues/446

res/cmake/dep/highfive.cmake
```diff
@@ -29,6 +29,8 @@ if(HighFive)
   )
   set(HIGHFIVE_USE_BOOST OFF)
   set(HIGHFIVE_BUILD_DOCS OFF) # conflicts with phare doc target
+  set(HIGHFIVE_EXAMPLES OFF)
+  set(HIGHFIVE_PARALLEL_HDF5 ON) # causes failure if cannot occure
```
nicolasaunai, line 33: I don't understand the comment

src/core/data/grid/gridlayout.h
```diff
@@ -774,6 +775,12 @@ namespace core
             return localPoint;
         }
 
+        template<typename T>
+        auto AMRToLocal(Point<T, dimension> AMRPoint) const
+        {
+            return AMRToLocal(AMRPoint());
```
nicolasaunai, line 781: I see this as a trick... but not the way I would program it in the first place if I didn't know it'd break on PGI.... So I think it will not be clear for future me this has to be done this way.... We should maybe comment why this is done this way (pgi...) so we don't break it later when we'll have forgotten the reason...

src/core/data/ions/particle_initializers/maxwellian_particle_initializer.h
```diff
@@ -197,9 +187,10 @@ void MaxwellianParticleInitializer<ParticleArray, GridLayout>::loadParticles(
             if (basis_ == Basis::Magnetic)
                 particleVelocity = basisTransform(basis, particleVelocity);
 
-            particles.emplace_back(Particle{cellWeight, particleCharge_,
-                                            AMRCellIndex.template toArray<int>(),
-                                            deltas(deltaDistrib, randGen), particleVelocity});
+            particles.emplace_back(
+                Particle{cellWeight, particleCharge_, AMRCellIndex.template toArray<int>(),
+                         core::ConstArrayFrom<dimension>([&] { return deltaDistrib(randGen); }),
```
nicolasaunai, line 192: From what I see what the code was before and based on what you said yesterday, I guess what PGI didn't like was the different return types in if constexprs for the lambda that was passed.

src/core/data/particles/particle_packer.h
```diff
@@ -10,6 +10,13 @@
 
 namespace PHARE::core
 {
+// PGI compiler doesn't (didn't at least) like static initializations of arrays, would result in an
```
nicolasaunai, line 13: maybe we could write the version in that comment

src/diagnostic/detail/types/particle.h
```diff
@@ -138,9 +141,11 @@ void ParticlesDiagnosticWriter<H5Writer>::initDataSets(
         std::string path{h5Writer_.getPatchPathAddTimestamp(lvl, patchID) + "/"};
         std::size_t part_idx = 0;
         core::apply(Packer::empty(), [&](auto const& arg) {
-            createDataSet(path + Packer::keys()[part_idx],
-                          null ? 0 : attr[Packer::keys()[part_idx]].template to<std::size_t>(),
-                          arg);
+            auto keys = core::packer_keys();
+            // PGI compiler doesn't like static array initializations
```
nicolasaunai, line 145: nvc++, version? maybe ?

src/amr/tagging/tagger.h
```diff
@@ -2,8 +2,8 @@
 #ifndef PHARE_TAGGER_H
 #define PHARE_TAGGER_H
 
-#include "physical_models/physical_model.h"
 #include "amr/types/amr_types.h"
+#include "solver/physical_models/physical_model.h"
```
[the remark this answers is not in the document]
PhilipDeegan: typically, solver depends on AMR

res/cmake/dep/highfive.cmake
PhilipDeegan: maybe unnecessary, if it's ON, and installed hdf5 is serial, an error is thrown

src/core/data/grid/gridlayout.h
PhilipDeegan: the issue is constructor resolution, something to try might just be to move the Point(std::array constructor up so it's first) in the list

src/amr/tagging/tagger.h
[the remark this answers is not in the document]
PhilipDeegan: also, adding in "#include "solver" in amr creates a bi-directional dependency

src/core/data/ions/particle_initializers/maxwellian_particle_initializer.h
PhilipDeegan: weirdly the other lambda was ok

src/core/data/grid/gridlayout.h
PhilipDeegan: I knew I had this issue before https://github.com/PhilipDeegan/PHARE/blob/gpu/src/core/numerics/interpolator/interpolator.h#L578

src/amr/tagging/tagger.h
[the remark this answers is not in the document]
nicolasaunai: yeah that's not good.
But since I created this solver lib, I have disliked it.... I created it with in mind that it would not depend on SAMRAI explicitly, but it does through the use of the PatchHierarchy, PatchLevels and Patches... I've almost reached the point where I seriously consider moving all this code back to amr...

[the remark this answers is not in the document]
nicolasaunai: on the other hand, there is code that's in amr, namely some tagging stuff, or some refinement and coarsening stuff, that do not depend on SAMRAI and that could be better located in core... TBC...

res/cmake/dep/highfive.cmake
nicolasaunai: an error is thrown at compile time (link time) ? because what? mpi-io stuff?

src/core/data/grid/gridlayout.h
nicolasaunai: ok let's try to move the Ctor first then and if it works add a comment so it stays there?

res/cmake/dep/highfive.cmake
PhilipDeegan: I've removed it, if we want to consider it another time we can

PhilipDeegan:
build passing on TC https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_GhPrPgccSamraisub

src/core/data/grid/gridlayout.h
PhilipDeegan: resolved via enable_if on the Point parameter pack constructor

src/core/utilities/point/point.h
```diff
@@ -129,17 +130,21 @@ namespace core
         auto begin() { return r.begin(); }
         auto end() { return r.end(); }
 
+        auto const& operator()() const { return r; }
```
nicolasaunai, line 133: is that one still needed?

res/cmake/dep/highfive.cmake
nicolasaunai: ok make an issue if you think it's worth it I have no strong opinion as I don't (yet) completely understand the point

src/core/utilities/point/point.h
PhilipDeegan: right you are, I thought I had removed it, one of the troubles on working on so many branches at once

nicolasaunai:
> gtest define macro causes issues
> 
> [google/googletest#3390](https://github.com/google/googletest/issues/3390)
> 
> ```
> sed -i 's/# define GTEST_CAN_STREAM_RESULTS_ 1/# define GTEST_CAN_STREAM_RESULTS_ 0/g' subprojects/googletest/googletest/include/gtest/internal/gtest-port.h
> ```
> 
> edit: resolved my cmake refactor, declare CXX_STANDARD after deps


why isn't the root cmakelists.txt modified then ?

PhilipDeegan:
> > gtest define macro causes issues
> > [google/googletest#3390](https://github.com/google/googletest/issues/3390)
> > ```
> > sed -i 's/# define GTEST_CAN_STREAM_RESULTS_ 1/# define GTEST_CAN_STREAM_RESULTS_ 0/g' subprojects/googletest/googletest/include/gtest/internal/gtest-port.h
> > ```
> > 
> > 
> > edit: resolved my cmake refactor, declare CXX_STANDARD after deps
> 
> why isn't the root cmakelists.txt modified then ?

because I haven't merged the updates from https://github.com/PhilipDeegan/PHARE/blob/openacc_pusher/CMakeLists.txt#L17

PhilipDeegan:
close @nicolasaunai ?

nicolasaunai:
why closing if that helped being compliant with PGI at reasonable cost?
On the other hand, if we have no PGI build on the CI, it might be difficult to assess whether this PR is enough and stays enough in the future

PhilipDeegan:
rebase/retry with pgcc on tc

PhilipDeegan:
possibly move to fork, thoughts @nicolasaunai ?

PR #545, opened 2021-05-06
merge solver and amr libs

nicolasaunai:


src/amr/multiphysics_integrator.h
```diff
@@ -18,14 +18,14 @@
 #include "amr/messengers/messenger.h"
 #include "amr/tagging/tagger.h"
 
-#include "solver/physical_models/hybrid_model.h"
-#include "solver/physical_models/mhd_model.h"
-#include "solver/physical_models/physical_model.h"
-#include "solver/solvers/solver.h"
-#include "solver/messenger_registration.h"
-#include "solver/level_initializer/level_initializer.h"
+#include "physical_models/hybrid_model.h"
```
PhilipDeegan, line 21: this compiles, but it probably shouldn't.

ie. there's a bunch of extra "-Isrc/$module" that are in cmake but I don't think should be

all includes should be from src

src/amr/multiphysics_integrator.h
```diff
@@ -18,14 +18,14 @@
...
+#include "physical_models/mhd_model.h"
+#include "physical_models/physical_model.h"
+#include "solvers/solver.h"
+#include "messenger_registration.h"
+#include "level_initializer/level_initializer.h"
```
PhilipDeegan, line 26: this compiles, but it probably shouldn't.

ie. there's a bunch of extra "-Isrc/$module" that are in cmake but I don't think should be

all includes should be from src

PhilipDeegan (approved):
just the includes otherwise LGTM

src/amr/multiphysics_integrator.h
nicolasaunai: done

PR #550, opened 2021-05-10
handle tuple values for smallest/largest patch size, validation tests

PhilipDeegan:


tests/simulator/test_validation.py
```diff
@@ -113,5 +111,63 @@ def test_1d_invalid(self, input):
         self._do_dim(1, input)
 
 
+
+
+    valid2D = [
```
nicolasaunai, line 116: should we add a case where smallest_patch_size (resp. largest) is set and not largest (res. smallest?
Same question with one given as a scalar and the other as a tuple/list?
and also add cases where they're given as a/ both list, b/ both tuples, c/ one list one tuple d/ one tuple one list?

PR #554, opened 2021-05-19
add a median filter and a cmap kwarg for pharsee/plotting.dist_plot

rochSmets:
older version of notebook in pharebook will still work as the kwarg "sigma" will be ignored

### DIFF
--- a/pyphare/pyphare/pharesee/plotting.py
+++ b/pyphare/pyphare/pharesee/plotting.py
@@ -17,7 +17,9 @@ def dist_plot(particles, **kwargs):
     * axis : ("x", "Vx"), ("x", "Vy"), ("x", "Vz"), ("Vx", "Vy") (default) --
        ("Vx", "Vz"), ("Vy", "Vz")
     * bins :  number of bins in each dimension, default is (50,50)
-    * sigma : width of the gaussian filter, default is (0,0)
+    * gaussian_filter_sigma : sigma of the gaussian filter, default is (0,0)
+    * median_filter_size : size of the median filter, default is (0,0)
+    * cmap : color table, default is "jet"
     * norm  : histogram will be normed to Normalize(0,norm)
     * kde : (default False) : adds contours of kernel density estimate
     * title : (str) title of the plot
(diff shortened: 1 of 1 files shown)


pyphare/pyphare/pharesee/plotting.py
```diff
@@ -60,7 +61,19 @@ def dist_plot(particles, **kwargs):
               bins=kwargs.get("bins", bins),
               weights=particles.weights)
 
-    sig = kwargs.get("sigma", (0,0))
+    if "gaussian_filter_sigma" in kwargs: 
+        from scipy.ndimage import gaussian_filter
+        sig = kwargs.get("gaussian_filter_sigma", (0,0))
+        image = gaussian_filter(h.T, sigma=sig)
+    elif "median_filter_size" in kwargs:
```
nicolasaunai, line 68: what if the user passes both gaussian_filter_sigma and median_filter_size?

rochSmets: good remark... now it is clear that the 2 filters can not be called at the same time

pyphare/pyphare/pharesee/plotting.py
```diff
@@ -60,7 +61,21 @@ def dist_plot(particles, **kwargs):
...
+        sig = kwargs.get("gaussian_filter_sigma", (0,0))
+        image = gaussian_filter(h.T, sigma=sig)
+    elif "median_filter_size" in kwargs and "gaussian_filter_sigma" not in kwargs: 
+        from scipy.ndimage import median_filter
+        siz = kwargs.get("median_filter_size", (0,0))
+        image = median_filter(h.T, size=siz)
+    elif "gaussian_filter_sigma" not in kwargs and "median_filter_size" not in kwargs:
+        image = h.T
+    else:
+        raise ValueError("gaussian and median filters can not be called at the same time")
```
nicolasaunai, line 75: not super important but : I would tend to think it is not this function's business to decide whether its caller is right or wrong to want to use both filters.
I would have simply done if gaussian apply gaussian, followed by if median apply median, and delegate the responsibility outside this function, which only aim is to plot what's askked.

nicolasaunai (approved):

nicolasaunai (approved):

Issue #462, opened 2021-03-13
add population mass to pharesee


PhilipDeegan:
once #459 is merged, h5 fluid/particle diagnostics will have a "pop_mass" attribute at the root of their files respectively.

This can be used to add a pop_mass attribute to the python PatchDataTypes for later access.
For the DataWrangler version of hierarchy_from, it's understood that a simulation exists, and thus the pop_mass(es) should be accessible, perhaps with some additional pybind function exports

something like 

```
std::unordered_map<std::string, double> pop_masses;
for(auto const& pop : simulator_.getHybridModel()->state.ions) pop_masses.emplace(pop.name(), pop.mass);
return pop_masses;
```

PR #555, opened 2021-05-19
#462 : add population mass in pharesee through a GetMass method

rochSmets:


pyphare/pyphare/pharesee/run.py
```diff
@@ -89,3 +89,19 @@ def filename(name):
...
             return hier
         return self._get_hierarchy(time, filename(pop_name), hier=hier)
+
+    def GetMass(self, pop_name):
+        list_of_qty = ['density', 'flux', 'domain', 'levelGhost', 'patchGhost']
+        list_of_mass = []
+
+        import h5py
+        for qty in list_of_qty:
+            file = os.path.join(self.path, "ions_pop_{}_{}.h5".format(pop_name, qty))
```
PhilipDeegan, line 99: slightly smaller maybe would be 

```
file = os.path.join(self.path, f"ions_pop_{pop_name}_{qty}.h5")
```

nicolasaunai (approved):
lgtm, no more to add than what phil said

PR #559, opened 2021-05-27
GhostBoxGeometryVariableFillPattern, to ignore the border node (overwrite_interior) for copy/stream operations for same level overlaps

PhilipDeegan:


src/amr/messengers/GhostBoxGeometryVariableFillPattern.cpp
```diff
@@ -0,0 +1,9 @@
+
+
+#include "GhostBoxGeometryVariableFillPattern.h"
```
nicolasaunai, line 3: maybe more explicit if renamed into `GhostOnlyVariablyFillPattern`

src/amr/messengers/GhostBoxGeometryVariableFillPattern.cpp
```diff
@@ -0,0 +1,9 @@
...
+
+namespace SAMRAI::xfer
+{
+const std::string GhostBoxGeometryVariableFillPattern::s_name_id = "BOX_GEOMETRY_FILL_PATTERN";
```
nicolasaunai, line 7: don't we need to change that string?

src/amr/messengers/GhostBoxGeometryVariableFillPattern.h
```diff
@@ -0,0 +1,178 @@
+
+#ifndef GhostBoxGeometryVariableFillPattern_
+#define GhostBoxGeometryVariableFillPattern_
+
+#include "SAMRAI/xfer/VariableFillPattern.h"
+
+namespace SAMRAI::xfer
```
nicolasaunai, line 7: should we keep these namespaces?

src/amr/messengers/GhostBoxGeometryVariableFillPattern.cpp
PhilipDeegan: should

PhilipDeegan: should, but can't

src/amr/messengers/GhostBoxGeometryVariableFillPattern.h
PhilipDeegan: dropped

nicolasaunai:
The code in this PR prevents schedules from exchanging primal shared border nodes.

PhilipDeegan:
obsoleted by #576

PR #560, opened 2021-05-27
save domain/patch ghost state to restore before second moveIons

PhilipDeegan:


src/core/numerics/ion_updater/ion_updater.h
```diff
@@ -168,9 +168,12 @@ void IonUpdater<Ions, Electromag, GridLayout>::updateMomentsOnly_(Ions& ions, El
             auto endInDomain = std::partition(firstGhostOut, std::end(outputArray), inDomainBox);
 
             interpolator_(firstGhostOut, endInDomain, pop.density(), pop.flux(), layout);
+
+            if (patchghost)
+                std::copy(firstGhostOut, endInDomain, std::back_inserter(domain));
         };
 
-        pushAndAccumulateGhosts(pop.patchGhostParticles(), tmpPatchGhost);
+        pushAndAccumulateGhosts(pop.patchGhostParticles(), tmpPatchGhost, true);
```
nicolasaunai, line 176: we should explain here why patch ghost particles have `copyInDomain=true` 

``` 
//  after this function is done domain particles overlaping ghost layers of neighbor patches
// are sent to these neighbor's patchghost particle array.
// After being pushed, some patch ghost particles may enter the domain. These
// thus need to be copied into the domain array so they are transfered to the neighbor patch
// ghost array and contribute to moments there too.
// on the contrary level ghost particles entering the domain here do not need to be copied
// since they contribute to nodes that are not shared with neighbor patches an since
// level border nodes will receive contributions from levelghost old and new particles 
```

src/core/numerics/ion_updater/ion_updater.h
```diff
@@ -80,7 +80,7 @@ void IonUpdater<Ions, Electromag, GridLayout>::updatePopulations(Ions& ions, Ele
     resetMoments(ions);
     pusher_->setMeshAndTimeStep(layout.meshSize(), dt);
 
-    if (mode == UpdaterMode::moments_only)
+    if (mode == UpdaterMode::domain_only)
```
nicolasaunai, line 83: I'm putting the comment here although it is for the line below.
Maybe  `updateMomentsOnly_` now needs to be renamed `updateAndDepositDomain()` and `updateAll` may become `updateAndDepositAll_` ? if you have better suggestions maybe?

PR #561, opened 2021-05-27
maxwell periodic test warning precision increase

PhilipDeegan:


nicolasaunai (commented):
Would be better to check for equality of user functions on all positions that PHARE will use, i.e. the 11 primal overlaped nodes or 10 dual overlaped nodes ?
Vth is same centering as the pressure.

pyphare/pyphare/core/gridlayout.py
```diff
@@ -228,9 +231,12 @@ def yeeCoords(self, knode, iStart, centering, direction, ds, origin, derivOrder)
         return x
 
 
-    def yeeCoordsFor(self, qty, direction):
+    def yeeCoordsFor(self, qty, direction, include_ghosts=True):
```
PhilipDeegan, line 234: TORM

pyphare/pyphare/pharein/simulation.py
```diff
@@ -504,6 +504,8 @@ def wrapper(simulation_object, **kwargs):
 
         kwargs["hyper_resistivity"] = check_hyper_resistivity(**kwargs)
 
+        kwargs["strict"] = kwargs.get('strict', True)
```
PhilipDeegan, line 507: TODO, change default to false, set as "True" in tests

nicolasaunai: TBD?

PhilipDeegan: done

nicolasaunai (approved):

PR #562, opened 2021-05-27
Geometry overlap changes for cases where patches overlap periodicaly with themselves

PhilipDeegan:


nicolasaunai (approved):
2 patch config to check later

PR #586, opened 2021-09-08
improve plotting

nicolasaunai:
- extracts yeeCoordsFor from the instance layout since we may want to use that without a proper instance
- move hierarchy.plot() to plot1d since it was only 1d and add a hierarchy.plot2d(). hierarchy.plot() now calls either one depending on ndim.

the 2d plot method plots plots ONE level (default root), in contrast to plotting.plot() which plots the finest data available.
Typical use case is for expert users who do not just want to see the "best" (finest) data available but data per level.

the mechanism is : loop on patches and pcolormesh each of them one after the other, on the same plot.

example : 

```python
r033 = Run("some_run_path")

qty="By"
t=0.8
vmin=-0.1
vmax=0.1

fig,(ax1,ax2) = plt.subplots(ncols=2)

# this returns a Hierarchy
Bh = r033.GetB(t)

# now we 2D plot 
Bh.plot(qty="By", level=0, ax=ax1, vmin=vmin, vmax=vmax)
Bh.plot(qty="By", level=1, ax=ax2, vmin=vmin, vmax=vmax)

ax1.set_xlim((13,28))
ax1.set_ylim((1,10))

# draw patch boundaries
for ax in (ax1,ax2):
    for patch in Bh.patch_levels[1].patches:
        pdat = patch.patch_datas["By"]
        dx,dy = pdat.layout.dl
        r = Rectangle((patch.layout.box.lower[0]*dx, patch.layout.box.lower[1]*dy),
                       patch.layout.box.shape[0]*dx, patch.layout.box.shape[1]*dy, 
                       fc="none", ec="b", alpha=0.4, lw=0.8)

        ax.add_patch(r)
``` 

this produces : 

image

### DIFF
--- a/pyphare/pyphare/core/gridlayout.py
+++ b/pyphare/pyphare/core/gridlayout.py
@@ -58,6 +58,34 @@ class YeeCentering(object):
         self.centerY = yee_centering["y"]
         self.centerZ = yee_centering["z"]
 
+
+def yeeCoordsFor(origin, nbrGhosts, dl, nbrCells, qty, direction, withGhosts=False):
+
+    assert direction in direction_to_dim, f"direction ({direction} not supported)"
+    assert qty in yee_centering[direction] or qty in yee_centering_lower[direction], f"qty ({qty} not supported)"
+    if qty in yee_centering_lower[direction] and qty not in yee_centering[direction]:
+        qty = qty[0].upper() + qty[1:]
+
+    centering = yee_centering[direction][qty]
+
+    offset = 0
+    dim = direction_to_dim[direction]
+    if withGhosts:
+        size = nbrCells[dim] + (nbrGhosts * 2)
+    else:
+        size = nbrCells[dim]
+
+    if centering == 'dual':
+        offset = 0.5*dl[dim]
+    else:
+        size += 1
+
+    if withGhosts:
+        return origin[dim] - nbrGhosts * dl[dim] + np.arange(size) * dl[dim] + offset
+    else:
+        return origin[dim] + np.arange(size) * dl[dim] + offset
+
+
 class GridLayout(object):
 
     def __init__(self, box=Box(0,0), origin=0, dl=0.1, interp_order=1):
@@ -240,24 +268,12 @@ class GridLayout(object):
         :param direction: can only be a single one
         """
 
-        assert direction in direction_to_dim, f"direction ({direction} not supported)"
-        assert qty in yee_centering[direction] or qty in yee_centering_lower[direction], f"qty ({qty} not supported)"
         if qty in yee_centering_lower[direction] and qty not in yee_centering[direction]:
             qty = qty[0].upper() + qty[1:]
 
-        centering = yee_centering[direction][qty]
-        nbrGhosts = self.nbrGhosts(self.interp_order, centering)
+        return yeeCoordsFor(self.origin, self.nbrGhosts(self.interp_order, yee_centering[direction][qty]),
+                            self.dl, self.box.shape, qty, direction, withGhosts=True)
 
-        offset = 0
-        dim = direction_to_dim[direction]
-        size = self.box.shape[dim] + (nbrGhosts * 2)
-
-        if centering == 'dual':
-            offset = 0.5*self.dl[dim]
-        else:
-            size += 1
-
-        return self.origin[dim] - nbrGhosts * self.dl[dim] + np.arange(size) * self.dl[dim] + offset
 
 
 
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -644,7 +644,7 @@ class PatchHierarchy:
 
         return fig
 
-    def plot(self, **kwargs):
+    def plot1d(self, **kwargs):
         """
         plot
         """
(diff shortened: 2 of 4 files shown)


pyphare/pyphare/core/gridlayout.py
```diff
@@ -58,6 +58,34 @@ def __init__(self):
         self.centerY = yee_centering["y"]
         self.centerZ = yee_centering["z"]
 
+
```
rochSmets, line 61: line 43 and 44, there is a nasty "prumal"...

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -690,6 +690,68 @@ def plot(self, **kwargs):
...
+            data = pdat.dataset[:]
+            nbrGhosts = 5 # CLEAN THIS
+            x = pdat.x
+            y = pdat.y
+            nx,ny =x.size, y.size
+            data = data.reshape((nx,ny))
+            data = data[nbrGhosts:-nbrGhosts, nbrGhosts:-nbrGhosts]
+            x = x[nbrGhosts:-nbrGhosts]
+            y = y[nbrGhosts:-nbrGhosts]
+            dx = x[1] - x[0]
```
rochSmets, line 716: i found this not very nice... and we often need to do this. could we decide to have the grid mesh as an attribute of the `patch_data` object or decorate this class ?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -690,6 +690,68 @@ def plot(self, **kwargs):
...
+            dy = y[1] - y[0]
+            x -= dx/2
```
rochSmets, line 718: better use `x -= 0.5*dx` as multiplying is faster

pyphare/pyphare/pharesee/plotting.py
```diff
@@ -277,6 +278,17 @@ def finest_field_plot(run_path, qty, **kwargs):
         vmax = kwargs.get("vmax", np.nanmax(DATA))
         cmap = kwargs.get("cmap", 'Spectral_r')
 
+        dx = finest_coords[0][1] - finest_coords[0][0]
```
rochSmets, line 281: same comment as above : this would deserve to be in the object. also because we could have \x[1]-x[0]=0.2` while `x[2]-x[1]=0.1999999999999`

pyphare/pyphare/pharesee/run.py
```diff
@@ -57,8 +57,8 @@ def make_interpolator(data, coords, interp, domain, dl):
                                 assume_sorted=False)
 
         nx = 1+int(domain[0]/dl[0])
-        x = dl[0]*np.arange(0, nx)
 
+        x = yeeCoordsFor([0]*dim, [5]*dim, dl, [nx],  qty, "x")
```
rochSmets, line 61: for coming reformating, we'd better have `[nbr_of_ghost]*dim` as second arg of `YeeCoordsFor`

rochSmets (commented):
few comments... but the main is to include the grid size in the `patch_datas` instead of recalculate it when needed

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: replace by pdat.layout.dl

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: not sure it changes anything in python (or in cpp actually)
changed anyway

pyphare/pyphare/pharesee/run.py
nicolasaunai: changer

pyphare/pyphare/pharesee/run.py
```diff
@@ -72,8 +72,11 @@ def make_interpolator(data, coords, interp, domain, dl):
         else:
             raise ValueError("interp can only be 'nearest' or 'bilinear'")
 
-        x = np.arange(0, domain[0]+dl[0], dl[0])
-        y = np.arange(0, domain[1]+dl[1], dl[1])
+        nCells = [1+int(d/dl) for d,dl in zip(domain, dl)]
+        x = yeeCoordsFor([0]*dim, [5]*dim, dl, nCells,  qty, "x")
+        y = yeeCoordsFor([0]*dim, [5]*dim, dl, nCells,  qty, "y")
+        #x = np.arange(0, domain[0]+dl[0], dl[0])
```
PhilipDeegan, line 78: torm?

nicolasaunai: I'd prefer to keep it for now

PhilipDeegan (approved):
I think there's some edits that don't belong in this PR that will hopefully be gone by rebasing

nicolasaunai:
Now can calculate the current density in 2D.
Can also plot the finest field by the superposition of pcolormesh per patch. This actually looks nicer than plotting a single ndarray with the finest solution since in most places the values won't be plotted where defined, and "nearest" interpolation produces artifacts.

```python
r = Run("path_to_run")
r.GetB(40)
fig,ax = B.plot(qty="Bz", vmin=-0.3, vmax=0.3, cmap="coolwarm", plot_patches=True)
ax.set_xlim((25,55))
```
produces : 

Bz41

PR #589, opened 2021-09-23
SAMRAI: Algorithm/schedule per quantity component

PhilipDeegan:


nicolasaunai (approved):

PhilipDeegan:
It has been observed that in at least for the case of level ghost refinement that there is a discrepancy which appears to be related to SAMRAI imposing the first geometry registered with a schedule to be used in all cases for generating overlaps.
It was seen that in 2d, for some patch layouts that there was mismatch between patches on pure level ghost regions. This mismatch seemed to be related to the fact that the level ghost area covered two coarser patch domains, with both patches belonging to a different MPI domain.

See for further information https://github.com/LLNL/SAMRAI/issues/170

nicolasaunai (approved):

PR #609, opened 2021-11-05
Less printing to screen, cerr not redirected

PhilipDeegan:
we probably don't want any prints in a super comupting facilty, so this isn't the "final" config, but it's nicer for dev


res/cmake/test.cmake
```diff
@@ -39,7 +39,6 @@ if (test AND ${PHARE_EXEC_LEVEL_MIN} GREATER 0) # 0 = no tests
 
   function(set_exe_paths_ binary)
     set_property(TEST ${binary}        PROPERTY ENVIRONMENT "PYTHONPATH=${PHARE_PYTHONPATH}")
-    set_property(TEST ${binary} APPEND PROPERTY ENVIRONMENT "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}")
```
nicolasaunai, line 42: why is that deleted?

PhilipDeegan: it's unnecessary and blocks manual exports from command line

PR #626, opened 2021-11-22
Parallel decomposition 

PhilipDeegan:


mkn.yaml
```diff
@@ -0,0 +1,147 @@
+#! clean build -x mpi -dtKOp py
```
PhilipDeegan, line 1: ignore this file

PhilipDeegan:
todo - readd pusher factory

PR #629, opened 2021-11-26
phare::amr::box for samrai interop / particle data copy benchmark

PhilipDeegan:


src/core/utilities/point/point.h
```diff
@@ -126,6 +126,42 @@ namespace core
         }
 
 
+
+        auto operator+(Type value) const
```
nicolasaunai, line 130: why not Type const& ?

src/core/utilities/point/point.h
```diff
@@ -126,6 +126,42 @@ namespace core
...
+        {
+            auto copy = *this;
+            for (auto iDim = 0u; iDim < dim; ++iDim)
+                copy[iDim] += value;
+            return copy;
+        }
+        auto operator+(std::array<Type, dim> value) const
```
nicolasaunai, line 137: same

nicolasaunai (approved):

src/core/utilities/point/point.h
PhilipDeegan: box.h is similar, but both could be changed

nicolasaunai (approved):

PR #641, opened 2022-01-19
Range maker/replacer/refiller

PhilipDeegan:
Allows splitting of containers into parts (ranges) so each part can be operated on independently.
Range replacer/refiller, currently not used, but added with tests for if/when it will be used. if not ever used. rm


PhilipDeegan:
moved to https://github.com/PhilipDeegan/PHARE/pull/61

PR #642, opened 2022-01-19
fix the finest_part_data to work (at least in 1d)

rochSmets:
This PR makes the `finest_part_data` working with the new format
of particles data (like `iCells` had a size `(num_of_part)` while now,
it is of size `(nom_of_part, ndim)`. It can be used with something like :

```
allParts = Run(run_path).GetParticles(time=time, pop_name=pop_name)
particles = finest_part_data(allParts)

# ..... scatter plot in x-vx plane
plt.scatter(particles["proton"].x,
            particles["proton"].v[:,0],
            s=particles["proton"].weights*100)
```

While used in 1D, not all the functions have been tested, neither their 2D
form.

### DIFF
--- a/pyphare/pyphare/core/box.py
+++ b/pyphare/pyphare/core/box.py
@@ -131,7 +131,7 @@ def shrink(box, size):
 
 def remove(box, to_remove):
     """
-    removes "remove" from the box
+    removes "to_remove" from the box
     this operation returns the list of the boxes
     that are not part of the intersection box*remove
     """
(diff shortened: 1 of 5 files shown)


nicolasaunai:
if I had to bet I would say this PR breaks `select()` function, which is used in the test : 
https://github.com/PHAREHUB/PHARE/blob/master/tests/simulator/test_initialization.py#L559-L560
that fails

PR #645, opened 2022-01-25
add mpi barrier to synchronize before deleting test hdf5 files

PhilipDeegan:
and move startMPI to not interfere with python subprocesses


PhilipDeegan:
the reason for moving "startMPI" is so that we only start MPI when we actually need it.
It can happen if automating tests in python3 via subprocesses, if (open) mpi is already loaded in the outer process, all subprocess calls to mpirun fail, see: https://github.com/open-mpi/ompi/issues/9729

nicolasaunai (approved):

PR #646, opened 2022-01-26
fix the ionIonBeam functional test

rochSmets:
This is a new version of the ionIonBeam test...

The former one used some unneeded latex command for producing the figure that bother the CI. They are now removed.

In the test, we now also calculate the real angular frequency of the mode with a decolvolution of the time series with the a `exp(gamma t)` function, the `gamma` being the interpolated value. But then the resulting expected cosine function is damped, with about 3 periods until `T=50`. We thus perform a `np.fft` to get the largest mode, but the result being too fragile, it does not appear a good idea to have an associated `assert`. The time evoluation of this cosine is displayed in the lower panel, so we can at least `visualy` check it on the CI

### DIFF
--- a/tests/functional/ionIonBeam/ion_ion_beam1d.py
+++ b/tests/functional/ionIonBeam/ion_ion_beam1d.py
@@ -30,8 +30,8 @@ def config():
     Simulation(
         smallest_patch_size=20,
         largest_patch_size=60,
-        time_step=0.0005, 
-        final_time=40,
+        final_time=50,
+        time_step=0.0005,
         boundary_types="periodic",
         cells=165,
         dl=0.2,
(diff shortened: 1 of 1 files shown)


tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -102,68 +102,91 @@ def vth(x):
...
+        ax2.text(0.40, 0.20, "omega (real) = {:5.3f}... expected 0.19".format(omega), transform=ax2.transAxes)
 
-        cov = np.fabs(np.array(pcov).flatten())
+        fig.savefig("ion_ion_beam1d.png")
 
-        assert np.fabs(popt[1]-0.09) < 1e-2
-        assert cov.max() < 1e-2
+        # compare with the values given gary et al. 1985
+        assert np.fabs(gamma-0.09) < 2e-2
+        #assert np.fabs(omega-0.19) < 8e-2
```
PhilipDeegan, line 189: rm comment?

PhilipDeegan (commented):
This test doesn't run on pull requests but I've verified it https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Phil_Build/73668?buildTab=log&focusLine=4907&linesState=18.4955

PhilipDeegan (approved):

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -102,68 +102,91 @@ def vth(x):
...
+        by = by_interpolator(x)
+        bz = by_interpolator(x)
 
-        byz = np.append(byz, mm)
+        # get the mode 1, as it is the most unstable in a box of length 33
+        mode1 = np.absolute(np.fft.fft(by-1j*bz)[1])
+        first_mode = np.append(first_mode, mode1)
 
-    popt, pcov = curve_fit(yaebx, times, byz, p0=[0.1, 0.1]) # last args are initial guess
+    popt, pcov = curve_fit(yaebx, times, first_mode, p0=[0.08, 0.09])
```
nicolasaunai, line 134: do we need to comment on those initial guesses which are quite close to the solution no ?

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -102,68 +102,91 @@ def vth(x):
...
+    damped_mode=first_mode*yaebx(times, 1/popt[0], -popt[1])
+
+    # find the omega for which "damped_mode" is the largest :
+    # this term is twice the one it should be because "mode1" resulting from
+    # an absolute value, this (cosx)^2 = cos(2x) then appears at the 2nd
+    # harmonoic (hence the factor 0.5 to get "omega")
+    # the factor "+1" is because we remove the DC component, so the value
+    # given by argmax has also to miss this value
+    omegas = np.fabs(np.fft.fft(damped_mode).real)
+    omega = 0.5*(omegas[1:omegas.size//2].argmax()+1)*2*np.pi/times[-1]
```
nicolasaunai, line 146: there is "//2"

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -102,68 +102,91 @@ def vth(x):
...
+
+    print(omegas[0:6])
+    print(omegas[1:omegas.size//2].argmax())
```
nicolasaunai, line 149: same //2

nicolasaunai (commented):
few questions/comments

PR #650, opened 2022-01-29
updater test with simple mocked out pusher

PhilipDeegan:


nicolasaunai:
was your new test failing before changing the ion updater?

PhilipDeegan:
this PR is more suggestive than functional, there is no test yet, just a setup for a test

nicolasaunai (commented):
I think we can merge the parts of this PR that revert the ion updater as it was, and suggest the test for later as an issue

PR #652, opened 2022-02-07
Cellmap

nicolasaunai:
This PR proposes to map particles per cell so that operations that are intrinsically "cell based" can avoid looping over all particles if only few are concerned. This typically concerns operations such as SAMRAI PatchData::copy() or streampack(), which needs to export particles from a particular group of cells (box). This also concerns copying ghost particles entering the domain into domain particle array. 

The idea of the CellMap is to store cells in a map, which values are particle array indexes.
The "list" of particle array indexes per cell is implemented in a so-called `BucketList`, which is a vector of arrays.

BucketList has certain advantages : 

- allow the contiguous storage of particle indexes (up to a given number)
- prevents/limits uncontrolled dynamic allocation

inconvenients:

- needs to estimate at compile time the most frequent number of particles per cell, to set the bucket size and be optimal.

THIS PR: 
- implements cellmap & buckelists. Note these classes know nothing about particles.
- removes the use (but keep the code) of boundary conditions for pushers, to be reconsidered when needed
- update ParticleData to now take advantage of cell mapping and export required particles without looping over the source particle array. An important change in the way these functions work is also that this PR removes the "PatchGhost" particle array as a possible source for copy and packStream. This choice has been made after noticing empirically that there was never a single particle coming from these arrays, and also that schedules are defined to take from source box to fill destination ghost box. Therefore having ghost particle arrays as source makes no sense.
- defines the `IndexRange`  as a range of indexes, convenient to use for cell mapping and particle users
- a Particle can now be printed as `std::cout << particle;`  for debug convenience
- `ContiguousParticles` is a particle array so is moved into the particle array file rather than in the particle header
- `ParticleArray` class is changed to use the cell mapping. A `ParticleArray` is now basically a wrapper of a particle container and its associated `CellMap`. The CellMap is not accessible from outside the ParticleArray which thus defines the needed interface.
- `Pusher` and subclasses are changed. The two overloads of `move()` are now a single one with 2 selectors, one per push step. This is not entirely satisfying since having 2 push steps is intrinsic to Boris and not general to all pushers but this can be addressed later. The particle Selector has also evolved. It was a lambda working on a particle. It is now a lambda processing a range of indexes and returns a partiionned range according to some predicate. This internally uses the `CellMap::Partition` method. Being a cell-based partition, it reduces the number of particles to loop over and focuses on those which are in targeted cells.
- `IonUpdater` is another big modif of this PR. Given the above changes to Pusher, it now defines selection lambdas consistently. Note also that now the ghosts that need to be copied into the domain array are copied via the CellMap export mechanism, that allows to avoid the copy_if. It is possible still that some partitions in some cases are unnecessary, this is a thinking beyond the scope of the current PR.
- the `Box` has been improved a bit; The biggest change here is the addition of a box iterator, that allows to return all cells (as Points) within a multidimensional box, assuming a given ordering.
- Ion Updater tests are changed because they initialized the particles in a way that is now incompatible with cell mapping. Particles are only mapped if added to a ParticleArray via push_back or emplace_back. Any default initialization and then modification of particles iCell will result in inconsistencies between the cellmap and the actual location of particles unless `change_icell` is explicitly called, as it is in the pusher move() method.
- same changes have been done in test pusher.


src/core/data/particles/particle_packer.hpp
```diff
@@ -19,7 +19,8 @@ class ParticlePacker
     {
     }
 
-    static auto get(Particle<dim> const& particle)
+    template<template<std::size_t> typename particle_t>
+    static auto get(particle_t<dim> const& particle)
```
PhilipDeegan, line 23: is this still needed?

src/core/numerics/boundary_condition/boundary_condition.hpp
```diff
@@ -24,14 +24,15 @@ namespace core
         ParticleIterator applyOutgoingParticleBC(ParticleIterator begin, ParticleIterator end)
         {
             // TODO loop while last partition not equel to begin.
-            auto partitions = partitionner(begin, end, boundaryBoxes_);
+            //            auto partitions = partitionner(begin, end, boundaryBoxes_);
```
PhilipDeegan, line 27: rm?

src/core/numerics/boundary_condition/boundary_condition.hpp
```diff
@@ -24,14 +24,15 @@ namespace core
...
 
             // applyBC for each box.
             // end while loop
 
             // particles to delete are between te last iterator
             // of 'partitions' and the end.
-            return partitions[boundaryBoxes_.size()];
+            //          return partitions[boundaryBoxes_.size()];
```
PhilipDeegan, line 34: rm?

src/core/numerics/moments/moments.hpp
```diff
@@ -46,17 +46,17 @@ namespace core
             if constexpr (std::is_same_v<DepositTag, DomainDeposit>)
             {
                 auto& partArray = pop.domainParticles();
-                interpolate(std::begin(partArray), std::end(partArray), density, flux, layout);
+                interpolate(makeRange(partArray), density, flux, layout);
```
PhilipDeegan, line 49: given the changes to interpolator.hpp

this is much the same as just `interpolate(partArray, density, flux, layout);`

containers and ranges should be interchangeable I think

src/core/utilities/bucketlist.hpp
```diff
@@ -0,0 +1,314 @@
...
+#include <type_traits>
+#include <vector>
+#include <cassert>
+#include <unordered_map>
+
+namespace PHARE::core
+{
+struct BucketListIndex
+{
+    using Bindex                          = int; // std::uint_fast16_t;
```
PhilipDeegan, line 22: commented needed?

src/core/utilities/bucketlist.hpp
```diff
@@ -0,0 +1,314 @@
...
+        return {bucket_idx - 1, bucket_size - 1};
+    else
+        return {bucket_idx, curr - 1};
+}
+
+
+template<std::size_t bucket_size>
+void BucketList<bucket_size>::empty()
+{
+    if (capacity() > 3 * size())
```
PhilipDeegan, line 196: replace "3" with named constexpr variable?

tests/simulator/test_advance.py
```diff
@@ -154,7 +154,7 @@ def vthz(*xyz):
                                     write_timestamps=timestamps,
```
PhilipDeegan, line 154: probably don't want to keep these edits?

tests/core/utilities/box/test_box.cpp
```diff
@@ -0,0 +1,367 @@
...
+    auto dummy3 = Point<int, 3>{};
+    for (auto const& point : b3)
+    {
+        dummy3 = point;
+    }
+    expected = Point{10, 12, 24};
+    EXPECT_EQ(expected, dummy3);
+}
+
+#if 0 // TODO TORM?
```
PhilipDeegan, line 256: ?

tests/amr/data/particles/stream_pack/test_main.cpp
```diff
@@ -160,36 +160,36 @@ TYPED_TEST(StreamPackTest, ShiftTheiCellWhenPackStreamWithPeriodics)
...
-                                                    particlesWriteStream.getBufferStart()};
-
-    destData.unpackStream(particlesReadStream, *cellOverlap);
-
-    auto expectediCell = ConstArray<int, dim>(0);
-
-    ASSERT_THAT(destData.domainParticles.size(), Eq(1));
-    ASSERT_THAT(destData.domainParticles[0].iCell, Eq(expectediCell));
-}
+// TYPED_TEST(StreamPackTest, PackInTheCorrectBufferWithPeriodics)
```
PhilipDeegan, line 163: forgot to update or we don't need it?

res/cmake/def.cmake
```diff
@@ -16,7 +16,7 @@ endif(withIPO)
 
 set (PHARE_WITH_CCACHE FALSE)
 if(devMode) # -DdevMode=ON
-  set(testDuringBuild ON)
+     #set(testDuringBuild ON)
```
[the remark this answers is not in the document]
nicolasaunai: done

src/amr/data/particles/particles_data.hpp
```diff
@@ -1,27 +1,48 @@
...
+#include "amr/utilities/box/amr_box.hpp"
+#include "core/utilities/point/point.hpp"
 
 #include "core/logger.hpp"
 
 namespace PHARE
 {
 namespace amr
 {
+    inline bool operator==(SAMRAI::hier::Box const& b1, SAMRAI::hier::Box const& b2)
```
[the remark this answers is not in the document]
nicolasaunai: done

src/amr/data/particles/particles_data.hpp
```diff
@@ -371,17 +323,17 @@ namespace amr
                         // interior array or ghost array
                         auto const intersect = getGhostBox() * overlapBox;
 
-                        for (auto const& particle : particleArray)
+                        for (auto& particle : particleArray)
```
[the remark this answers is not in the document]
nicolasaunai: just had it back, apparently it builds... so done

src/core/data/particles/particle_array.hpp
```diff
@@ -3,77 +3,211 @@
...
+        // But in principle that cell will never be accessed because it is outside the patch.
+        // The only thing "bad" if these indexes are not deleted is that the
+        // size of the cellmap becomes unequal to the size of the particleArray.
+        // but  ¯\_(ツ)_/¯
+        return particles_.erase(first, last);
+    }
+
+
+
+    Particle_t& emplace_back(bool mapping = true)
```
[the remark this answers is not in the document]
nicolasaunai: I don't see an emplace_back(Args... args) in the particleArray

src/core/data/particles/particle_packer.hpp
nicolasaunai: rmed

src/core/data/particles/particle_array.hpp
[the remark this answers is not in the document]
PhilipDeegan: `https://github.com/PhilipDeegan/PHARE/blob/soa/src/core/data/particles/particle_array_aos.hpp#L132` :)

src/core/numerics/moments/moments.hpp
nicolasaunai: seems ok

tests/core/utilities/box/test_box.cpp
nicolasaunai: TORMED

tests/amr/data/particles/stream_pack/test_main.cpp
nicolasaunai: it's back in, modified slightly. Was indeed removed because it puts a particle in a ghost cell and expects it in the neighbor domain.
Changed it to put the particle in the last domain cell of source, and expect it in ghost of destination in cell -1 (periodicity).
Seems to work.

nicolasaunai:
All comments have been addressed, I'm closing this as #656 is the one to consider

PR #552, opened 2021-05-14
shaping C++ diagostics from flat to expected shape when reading back

PhilipDeegan:

PR #355, opened 2020-10-08
add data getters and ploting routines

nicolasaunai:
Typically allows one to do stuff like 

```python
path = "phare_jobs/initialization_tests/phare_outputs_1000"
r = Run(path)

# getting data as hierarchies

B = r.GetB(0)
E = r.GetE(0)
Ni = r.GetNi(0)
Vi = r.GetVi(0)
Fp = r.GetFlux(0, "protons")
J = r.GetJ(0)

# plotting scalars

Ni.plot()
```

image

```python

# plotting components of vectors 

Vi.plot(qty="bulkVelocity_z")

``` 

image

```python

# distribution functions

parts = r.GetParticles(0, "protons")
parts.dist_plot(select=lambda particles: particles.select(Box(2.,12.5), box_type="pos"),
                axis=("x","Vz"),dual
               bins=(100,200))

``` 
image

or in full velocity space

``` python
parts = r.GetParticles(0, "protons")
parts.dist_plot(select=lambda particles: particles.select(Box(2.,2.5), box_type="pos"),
                kde=True,
                axis=("Vx","Vy"),
               bulk=True)
``` 
image

PR #653, opened 2022-02-09
fix the finest_part_data to work (at least in 1d)

rochSmets:
This PR makes the `finest_part_data` working with the new format
of particles data (like `iCells` had a size `(num_of_part)` while now,
it is of size `(nom_of_part, ndim)`. It can be used with something like :

```
allParts = Run(run_path).GetParticles(time=time, pop_name=pop_name)
particles = finest_part_data(allParts)

plt.scatter(particles["proton"].x,
            particles["proton"].v[:,0],
            s=particles["proton"].weights*100)
```

### DIFF
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -510,7 +510,8 @@ def finest_part_data(hierarchy, time=None):
                         else:
                             toRemove = np.concatenate((toRemove, within))
 
-                    parts = remove(parts, toRemove)
+                    if toRemove.size != 0:
+                        parts = remove(parts, toRemove)
                     if parts is not None:
                         particles[popname].add(parts)
     return particles
--- a/pyphare/pyphare/pharesee/particles.py
+++ b/pyphare/pyphare/pharesee/particles.py
@@ -39,8 +39,6 @@ class Particles:
 
 
     def xyz(self, i=0):
-        if self.ndim == 1:
-            return self.dl[:,0]*(self.iCells[:] + self.deltas[:])
         return self.dl[:,i]*(self.iCells[:,i] + self.deltas[:,i])
 
 
@@ -241,8 +239,8 @@ def remove(particles, idx):
     if len(idx) == particles.size():
         return None
 
-    icells = np.delete(particles.iCells, idx)
-    deltas = np.delete(particles.deltas, idx)
+    icells = np.delete(particles.iCells, idx, axis=0)
+    deltas = np.delete(particles.deltas, idx, axis=0)
     vx = np.delete(particles.v[:,0], idx)
     vy = np.delete(particles.v[:,1], idx)
     vz = np.delete(particles.v[:,2], idx)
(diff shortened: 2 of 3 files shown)


pyphare/pyphare/pharesee/plotting.py
```diff
@@ -103,7 +103,7 @@ def dist_plot(particles, **kwargs):
         ax.set_xlim(kwargs["xlim"])
     if "ylim" in kwargs:
         ax.set_ylim(kwargs["ylim"])
-    ax.legend()
+    #ax.legend()
```
nicolasaunai, line 106: no legend ?

nicolasaunai (approved):

pyphare/pyphare/pharesee/particles.py
```diff
@@ -39,8 +39,6 @@ def __init__(self, **kwargs):
 
 
     def xyz(self, i=0):
-        if self.ndim == 1:
```
PhilipDeegan, line 42: is this no longer necessary since #552  ?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -498,7 +498,8 @@ def finest_part_data(hierarchy, time=None):
                         else:
                             toRemove = np.concatenate((toRemove, within))
 
-                    parts = remove(parts, toRemove)
+                    if toRemove.size != 0:
```
PhilipDeegan, line 501: in this case of L0 only and/or no finer data for a spatial area?

PhilipDeegan (approved):

PhilipDeegan:
is "finest_part_data" called anywhere in functional tests, even indirectly?

pyphare/pyphare/pharesee/plotting.py
rochSmets: well, as is, i do not see any labels to write in the legend... and this legend gives me a warning. as i do not see the interest of this, i commented it.

pyphare/pyphare/pharesee/particles.py
rochSmets: I dont think so... and ih kept, it gives me problems... since now, quantities like `deltas`, `iCells` of a particles are of shape `(numOfPart, ndim)`

pyphare/pyphare/pharesee/plotting.py
rochSmets: because i'm not comfortable with removing something I did not wrote, since I'm not totally sure it has to...

nicolasaunai: that's what the review is for. Go for it.
I think this ax.legend() is what drew the little gray square on the top right part of the figure in : https://github.com/PHAREHUB/PHARE/pull/355

PhilipDeegan:
> ... Would you prefer the code to be called somewhere before being merged? - it's a small update...

no it's ok, I'll try to remember to run the future ionbeam PR through the nightly build to trigger this

PR #654, opened 2022-02-11
SOA ParticleArray

PhilipDeegan:
since introducing cellmap this needs reconsidering


PhilipDeegan:
issues with std::mem_fn on older compilers. I saw this with gcc 10.2.0, but it no longer appears in gcc 10.2.1

this isn't the only way it can be achieved, it was just the best way given benchmark analysis

PR #657, opened 2022-02-14
python3 tests concurrency by test_case across all included files

PhilipDeegan:
using this method python tests (with 3d from #591) was taking 16 hours on TC.
whereas the normal method didn't finish after 24 hours

### DIFF
--- a/tests/simulator/test_python_concurrent.py
+++ b/tests/simulator/test_python_concurrent.py
@@ -4,54 +4,80 @@
 """
 
 import os
+import time
 import unittest
-import multiprocessing
+from multiprocessing import cpu_count
 
 from tests.simulator.test_validation import SimulatorValidation
+from tests.simulator.test_diagnostics import DiagnosticsTest  # mpirun -n 1/2/3/4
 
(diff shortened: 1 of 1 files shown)


tests/simulator/test_python_concurrent.py
```diff
@@ -4,54 +4,130 @@
...
+    queue.put(runnable())
+
+def print_tests(batches):
+    for batch in batches:
+        for test in batch.tests:
             print(test)
+
+if __name__ == "__main__":
+    batches = build_batches()
+    # batches = [TestBatch([batches[0].tests[0]])]
```
PhilipDeegan, line 91: rm

tools/ctest_exec_mp.py
```diff
@@ -1,47 +0,0 @@
-
```
PhilipDeegan, line 1: keep this, caliper tc build uses it to isolate logs per test

PhilipDeegan:
I've pretty much refactored this to a custom library https://github.com/philipdeegan/phlop
I do think we should probably use this for less total test runtime

PR #672, opened 2022-03-08
cmake: set default optimization level 

PhilipDeegan:
if cmake build type is not set, and CMAKE_CXX_FLAGS is not set


nicolasaunai (approved):

PhilipDeegan:
gotta make sure it (CMAKE_CXX_FLAGS_RELWITHDEBINFO) doesn't include "-DNDEBUG"

PhilipDeegan:
I got the info from here `https://cmake.org/pipermail/cmake/2007-April/013544.html`

it looks like CMAKE_CXX_FLAGS_RELWITHDEBINFO does contain "-DNDEBUG", so we have asserts off by default.

Issue #667, opened 2022-03-04
Restarts save from simulation elapsed time


PhilipDeegan:
Rather than just having coarsest timestamps to check to see if restart files should be written, it would be good to also include a mechanism to have timestamps which are from when the simulation started, such that restart files would be made after X hours and/or minutes

PhilipDeegan:
I would guess the only check that is really feasible is if the elapsed time and the next restart time are the same, only make one, but don't attempt to do anything smart like checking if there was a recent restart written if the elapsed time is triggered?

nicolasaunai:
Not sure I understand you last comment/question. IMO the "elapsed time" is a mode of operation for the checkpointing. If selected, the code dumps a restart file every XX "real time units", rather than every "simulated time units".
So actually, what we could do is ask the user the checkpointing time interval in one variable, and the unit (real time VS simulated time) in another?

PhilipDeegan:
is it feasible to give input parameters for both "modes" such that they can both be active?
in that case, they could possibly try to write restart files at the same timestep

nicolasaunai:
I see now. I had not imagined they could be both actives at once. I agree that (although probably quite unfortunate) they could dump at the same time, in which case we'd want only one. I don't think in that case we can do something with the "next" dump times, because if you know when you're going to dump in the "sim unit" mode, you can't really know how when in terms of sim units the next real time dump will occur. 
I think in case both modes are active, the sim unit must be prioritized and the time at which it was done must be kept in memory so that next real time based dump checks if last "sim unit" dump has been done "recently", in which case that next real time dump is skipped?

PR #676, opened 2022-03-19
Restarts : elapsed time file writes #667

PhilipDeegan:
#667 

### Summary by CodeRabbit

- New Feature: Added support for elapsed timestamps in the restart options, allowing users to specify restart points based on elapsed time.
- New Feature: Introduced a new method to retrieve all available quantities in a simulation run, enhancing data accessibility.
- Improvement: Enhanced validation for restart options, ensuring data integrity and user input accuracy.
- Improvement: Updated the simulation run method to use the current time instead of the start time, providing more accurate time tracking.
- Test: Added new test cases to verify the functionality of elapsed time restarts and validate elapsed timestamps.

### DIFF
--- a/pyphare/pyphare/pharein/__init__.py
+++ b/pyphare/pyphare/pharein/__init__.py
@@ -288,8 +288,11 @@ def populateDict():
 
         add_string(restarts_path + "filePath", restart_file_path)
 
+        if "elapsed_timestamps" in restart_options:
+            pp.add_array_as_vector(restarts_path + "elapsed_timestamps", restart_options["elapsed_timestamps"])
 
-        pp.add_array_as_vector(restarts_path + "write_timestamps", restart_options["timestamps"])
+        if "timestamps" in restart_options:
+            pp.add_array_as_vector(restarts_path + "write_timestamps", restart_options["timestamps"])
 
         add_string(restarts_path + "serialized_simulation", serialize_sim(simulation))
     #### restarts added
--- a/pyphare/pyphare/pharein/restarts.py
+++ b/pyphare/pyphare/pharein/restarts.py
@@ -9,20 +9,39 @@ from pyphare.core import phare_utilities
 
 def validate(sim):
 
-    sim.restart_options["timestamps"] = phare_utilities.np_array_ify(sim.restart_options["timestamps"])
-    init_time = sim.start_time()
+    restart_options = sim.restart_options
 
-    timestamps = sim.restart_options["timestamps"]
-    if np.any(timestamps < init_time):
-        raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be less than simulation.init_time({init_time}))")
-    if np.any(timestamps > sim.final_time):
-        raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
-    if not np.all(np.diff(timestamps) >= 0):
-        raise RuntimeError("Error: restart timestamps not in ascending order)")
-    if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):
-        raise RuntimeError("Error: restart timestamps is inconsistent with simulation.time_step")
+    if "elapsed_timestamps" in restart_options:
+        import datetime
+        restart_options["elapsed_timestamps"] = [
+            int(ts.total_seconds()) if isinstance(ts, datetime.timedelta) else ts
+            for ts in phare_utilities.np_array_ify(restart_options["elapsed_timestamps"])]
 
-    sim.restart_options["timestamps"] = conserve_existing(sim)
+        if not np.all(np.diff(restart_options["elapsed_timestamps"]) >= 0):
+            raise RuntimeError(f"Error: restart_options elapsed_timestamps not in ascending order)")
+
+        # seconds_in_an_hour = 60 ** 2
+        # for cmp_idx, ref_ts in enumerate(restart_options["elapsed_timestamps"][1:]):
+        #     cmp_ts = restart_options["elapsed_timestamps"][cmp_idx]
+        #     if ref_ts - cmp_ts < seconds_in_an_hour:
+        #         raise RuntimeError(f"Error: time betweeen restart_options elapsed_timestamps must be at least one hour)")
+
+    if "timestamps" in restart_options:
+        restart_options["timestamps"] = phare_utilities.np_array_ify(restart_options["timestamps"])
+        init_time = sim.start_time()
+
+        timestamps = restart_options["timestamps"]
+        if np.any(timestamps < init_time):
+            raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be less than simulation.init_time({init_time}))")
+        if np.any(timestamps > sim.final_time):
+            raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
+        if not np.all(np.diff(timestamps) >= 0):
+            raise RuntimeError(f"Error: restart_options timestamps not in ascending order)")
+        if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):
+            raise RuntimeError(f"Error: restart_options timestamps is inconsistent with simulation.time_step")
+
+        sim.restart_options["timestamps"] = conserve_existing(sim)
+
 
 # ------------------------------------------------------------------------------
 
--- a/pyphare/pyphare/pharein/simulation.py
+++ b/pyphare/pyphare/pharein/simulation.py
@@ -395,12 +395,14 @@ def check_diag_options(**kwargs):
 
 def check_restart_options(**kwargs):
 
+    valid_keys = ["dir", "elapsed_timestamps", "timestamps", "mode", "restart_time"]
     restart_options = kwargs.get("restart_options", None)
 
     if restart_options is not None:
 
-        if "timestamps" not in restart_options:
-            raise ValueError (f"restart_options expects a list of timestamps")
+        for key in restart_options.keys():
+            if key not in valid_keys:
+                raise ValueError(f"invalid option ({key}), valid options are {valid_keys}")
 
         valid_modes = ["conserve", "overwrite"]
         if "mode" not in restart_options:
@@ -602,6 +604,7 @@ class Simulation(object):
                    restart_options={"dir": restart_outputs,
                                    "mode": "overwrite" or "conserve",
                                    "timestamps" : [.009, 99999]
+                                   "elapsed_timestamps" : [datetime.timedelta(hours=1)],
                                    "restart_time" : 99999.99999 },
                    strict=True (turns warnings to errors, false by default),
                   )
--- a/src/core/utilities/mpi_utils.cpp
+++ b/src/core/utilities/mpi_utils.cpp
@@ -55,7 +55,7 @@ std::string date_time(std::string format)
 
 std::int64_t unix_timestamp_now()
 { // static cast as std::time_t typedef can vary across operating systems
-    return collect(static_cast<std::int64_t>(std::time(NULL)))[0];
+    return all_get_from_rank_0([]() { return static_cast<std::int64_t>(std::time(NULL)); });
 }
 
 } // namespace PHARE::core::mpi
--- a/src/core/utilities/mpi_utils.cpp
+++ b/src/core/utilities/mpi_utils.cpp
@@ -53,5 +53,9 @@ std::string date_time(std::string format)
     return collect(date_time)[0];
 }
 
+std::int64_t unix_timestamp_now()
+{ // static cast as std::time_t typedef can vary across operating systems
+    return collect(static_cast<std::int64_t>(std::time(NULL)))[0];
+}
 
 } // namespace PHARE::core::mpi
--- a/src/core/utilities/mpi_utils.hpp
+++ b/src/core/utilities/mpi_utils.hpp
@@ -36,6 +36,8 @@ void barrier();
 
 std::string date_time(std::string format = "%Y-%m-%d-%H:%M:%S");
 
+std::int64_t unix_timestamp_now();
+
 template<typename Data>
 auto mpi_type_for()
 {
@@ -47,6 +49,8 @@ auto mpi_type_for()
         return MPI_INT;
     else if constexpr (std::is_same_v<std::uint32_t, Data>)
         return MPI_UNSIGNED;
+    else if constexpr (std::is_same_v<std::int64_t, Data>)
+        return MPI_INT64_T;
     else if constexpr (std::is_same_v<std::uint8_t, Data>)
         return MPI_UNSIGNED_SHORT;
     else if constexpr (std::is_same_v<std::size_t, Data>)
--- a/src/core/utilities/mpi_utils.hpp
+++ b/src/core/utilities/mpi_utils.hpp
@@ -61,6 +61,49 @@ auto mpi_type_for()
     // don't return anything = compile failure if tried to use this function
 }
 
+
+template<typename Fn, typename... Args>
+auto all_get_from(int const& rank_, Fn&& fn, Args&&... args)
+{
+    using Data = std::decay_t<std::result_of_t<Fn&(Args & ...)>>;
+
+    Data var;
+    auto local_rank = rank();
+    if (local_rank == rank_)
+        var = fn(args...);
+    void* data = &var;
+
+    int count = 1; // default
+    MPI_Datatype sendtype;
+    if constexpr (std::is_same_v<std::string, Data> or core::is_std_vector_v<Data>)
+    {
+        sendtype = mpi_type_for<typename Data::value_type>();
+        count    = all_get_from(rank_, [&]() { return var.size(); });
+        if (local_rank != rank_)
+            var.reserve(count);
+        data = var.data();
+    }
+    else
+        sendtype = mpi_type_for<Data>();
+
+    MPI_Bcast(         // MPI_Bcast
+        data,          //   void *buffer
+        count,         //   int count
+        sendtype,      //   MPI_Datatype sendtype
+        0,             //   int root == 0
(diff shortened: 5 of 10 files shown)


pyphare/pyphare/pharein/__init__.py
```diff
@@ -283,8 +283,11 @@ def as_paths(rb):
 
         add_string(restarts_path + "filePath", restart_file_path)
 
+        if "elapsed_timestamps" in restart_options:
+            pp.add_array_as_vector(restarts_path + "elapsed_timestamps", restart_options["elapsed_timestamps"])
 
-        pp.add_array_as_vector(restarts_path + "write_timestamps", restart_options["timestamps"])
+        if "timestamps" in restart_options:
+            pp.add_array_as_vector(restarts_path + "write_timestamps", restart_options["timestamps"])
```
PhilipDeegan, line 290: `"timestamps"` should maybe be renamed

pyphare/pyphare/pharein/restarts.py
```diff
@@ -9,20 +9,24 @@
...
-        raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
-    if not np.all(np.diff(timestamps) >= 0):
-        raise RuntimeError(f"Error: {clazz}.{key} not in ascending order)")
-    if not np.all(np.abs(timestamps / sim.time_step - np.rint(timestamps/sim.time_step) < 1e-9)):
-        raise RuntimeError(f"Error: {clazz}.{key} is inconsistent with simulation.time_step")
-
-    sim.restart_options["timestamps"] = conserve_existing(sim)
+    if "elapsed_timestamps" in sim.restart_options:
+        sim.restart_options["elapsed_timestamps"] = phare_utilities.np_array_ify(sim.restart_options["elapsed_timestamps"])
+
```
nicolasaunai, line 14: there are checks for timestamps that I guess don't make sense for elapsed ones. Maybe there are few (others) that make sense to do here though.
I think we should probably not allow elapsed timestamps that are closer than XX time units (to determine). First, close elapsed time stamps may stall the simulation.. worse timestamps too close to each other may be closer than the time it takes to dump, and the simulation would end up dumping two checkpoints almost simultaneously.

It's hard to define a minimum time interval... I'd think 1 hour is a good start, we may want to have a minimum time that depends on how much data is to write (roughly estimated from the init I guess) in future versions of the feature.

src/restarts/restarts_manager.hpp
```diff
@@ -69,25 +74,48 @@ class RestartsManager : public IRestartsManager
...
     }
+    bool needsElapsedAction_(double const nextTime, const double /*timeStamp*/,
+                             const double /*timeStep*/) const
+    {
+        auto const current_time = linux_time_now();
+        return current_time > nextTime;
+    }
+
+
+    auto static linux_time_now() { return std::time(NULL); }
```
nicolasaunai, line 91: curious: why "linux"?

src/restarts/restarts_manager.hpp
```diff
@@ -100,8 +128,18 @@ RestartsManager<Writer>::addRestartDict(initializer::PHAREDict const& params)
...
+    if (params.contains("write_timestamps"))
+        restarts_properties_->writeTimestamps
+            = params["write_timestamps"].template to<std::vector<double>>();
+
+    if (params.contains("elapsed_timestamps"))
+    {
+        restarts_properties_->elapsedTimestamps
+            = params["elapsed_timestamps"].template to<std::vector<double>>();
+        for (auto& time : restarts_properties_->elapsedTimestamps)
+            time += start_time_; // expected for comparison later
```
nicolasaunai, line 141: :small_red_triangle: :small_red_triangle: :small_red_triangle: :small_red_triangle:
hmmm just a question : when the program runs in parallel, don't you expect to have issues ? The clock time has no reason to be the same on all processes. I think we need to sync that somehow.
:small_red_triangle: :small_red_triangle: :small_red_triangle: :small_red_triangle:

pyphare/pyphare/pharein/restarts.py
```diff
@@ -9,20 +9,24 @@
...
+    if "timestamps" in sim.restart_options:
+        sim.restart_options["timestamps"] = phare_utilities.np_array_ify(sim.restart_options["timestamps"])
```
nicolasaunai, line 16: we should also probably decide in which time units the users provides their timestamps. Is that seconds, minutes, hours, days?
Also do they provide time stamps (i.e. dates) or real time intervals (seconds, hours, etc.)? 
I would think users provide timestamps **relative** to the start time, in some unit that we need to decide what it is.... I'd say hours?

tests/simulator/test_restarts.py
```diff
@@ -97,6 +97,59 @@ def ddt_test_id(self):
...
+
+
+        def count_levels_and_patches(qty):
+            n_levels = len(qty.patch_levels)
+            n_patches = 0
+            for ilvl, lvl in qty.patch_levels.items():
+                n_patches += len(qty.patch_levels[ilvl].patches)
+            return n_levels, n_patches
+
+        n_quantities_per_patch = 20
```
nicolasaunai, line 129: do you think we can get that magic number from somewhere so it does not become inconsistent later?
Could be a `run.GetAllAvailableQties(time=0)` that returns the list of names of all quantities (exploding vectors into components) and `n_quantities_per_patch=20` would become 
`n_quantities_per_patch = len(run.GetAllAvailableQties())`
(just an example there might be a better idea)

nicolasaunai: the following checks have to stay consistent if that total changes though...

tests/simulator/test_restarts.py
```diff
@@ -97,6 +97,59 @@ def ddt_test_id(self):
...
+
+        def check_particles(qty0, qty1):
+            return check(qty0, qty1, lambda pd0, pd1: self.assertEqual(pd0.dataset, pd1.dataset))
+
+        def check_field(qty0, qty1):
+            return  check(qty0, qty1, lambda pd0, pd1: np.testing.assert_equal(pd0.dataset[:], pd1.dataset[:]))
+
+
+        def count_levels_and_patches(qty):
+            n_levels = len(qty.patch_levels)
```
nicolasaunai, line 123: generally I think it may be safer to use len(qty.levels())

tests/simulator/test_restarts.py
```diff
@@ -161,58 +215,81 @@ def test_restarts(self, dim, interp, simInput):
...
+        timestamps = [time_step * restart_idx, time_step * time_step_nbr]
 
-            n_levels, n_patches = count_levels_and_patches(run0.GetB(time))
-            self.assertEqual(n_levels, 2) # at least 2 levels
-            self.assertGreaterEqual(n_patches, n_levels) # at least one patch per level
-            self.assertEqual(checks, n_quantities_per_patch * n_patches)
+        # first simulation
+        local_out = f"{out}/elapsed_test/{dim}/{interp}/mpi_n/{cpp.mpi_size()}/id{self.ddt_test_id()}"
+        simput["restart_options"]["dir"] = local_out
+        simput["restart_options"]["elapsed_timestamps"] = [9]
```
nicolasaunai, line 262: what is the unit of this 9?

nicolasaunai (commented):
ok thx for the PR.
some questions/comments + one possible issue with MPI sync of the start time and dump times

src/restarts/restarts_manager.hpp
PhilipDeegan: should be "unix" https://www.unixtimestamp.com/

tests/simulator/test_restarts.py
PhilipDeegan: currently seconds

tests/simulator/test_restarts.py
PhilipDeegan: number of quantities is configured in the diagnostic setup for this test, it's not for "restarts" per say

PhilipDeegan: it has a name, so it's not really a magic number, but I see you point, maybe it should be in `tests/diagnostics/__init__.py` as the number of quantities in the diagnostic output is a product of the "dump_all_diags" function

pyphare/pyphare/pharein/restarts.py
PhilipDeegan: for the sake of testing it should probably support seconds

PhilipDeegan: maybe we can do something with "timedelta" / datetime https://docs.python.org/3/library/datetime.html ?

nicolasaunai: yes I was considering this option too... @jeandet worked quite a bit on python/C++ time stuff, in case there are unexpected issues.

pyphare/pyphare/pharein/restarts.py
PhilipDeegan: this is done

src/restarts/restarts_manager.hpp
```diff
@@ -49,7 +50,10 @@ class RestartsManager : public IRestartsManager
                                                         initializer::PHAREDict const& dict)
     {
         auto rMan = std::make_unique<RestartsManager>(Writer::make_unique(hier, model, dict));
-        if (dict.contains("write_timestamps")) // else is only loading not saving restarts
+        auto restarts_are_written = core::any(
+            core::generate([&](auto const& v) { return dict.contains(v); },
+                           std::vector<std::string>{"write_timestamps", "elapsed_timestamps"}));
+        if (restarts_are_written) // else is only loading not saving restarts
```
nicolasaunai, line 56: I don't understand the comment.

src/core/utilities/mpi_utils.cpp
```diff
@@ -53,5 +53,9 @@ std::string date_time(std::string format)
     return collect(date_time)[0];
 }
 
+std::int64_t unix_timestamp_now()
+{ // static cast as std::time_t typedef can vary across operating systems
+    return collect(static_cast<std::int64_t>(std::time(NULL)))[0];
```
nicolasaunai, line 58: If I understand correctly, this gather the local unix time per process on all processes and then each process take the time of rank 0?
I wonder if that is what should be done. When using a very large number of MPI processes we will be transferring lots of data for just using 1 value

There is [MPI Broadcast](https://mpitutorial.com/tutorials/mpi-broadcast-and-collective-communication/) that can/should preferably used for that. It allows one rank, here rank 0, to send its value to all other processes.

tests/simulator/test_restarts.py
```diff
@@ -161,58 +212,85 @@ def test_restarts(self, dim, interp, simInput):
...
+            simput[key] = [simput[key]] * dim
 
-        def count_levels_and_patches(qty):
-            n_levels = len(qty.patch_levels)
-            n_patches = 0
-            for ilvl, lvl in qty.patch_levels.items():
-                n_patches += len(qty.patch_levels[ilvl].patches)
-            return n_levels, n_patches
+        if "refinement" not in simput:
+            b0 = [[10 for i in range(dim)], [19 for i in range(dim)]]
```
nicolasaunai, line 236: suggestion : 

```python
lo_cell = 10
hi_cell = 19
b0 = [ [lo_cell]*dim, [hi_cell]*dim ]
```

pyphare/pyphare/pharein/restarts.py
```diff
@@ -9,20 +9,41 @@
...
-    sim.restart_options["timestamps"] = conserve_existing(sim)
+    restart_options = sim.restart_options
+
+    if "elapsed_timestamps" in restart_options:
+        import datetime
+        restart_options["elapsed_timestamps"] = [
+            int(ts.total_seconds()) if isinstance(ts, datetime.timedelta) else ts
+            for ts in phare_utilities.np_array_ify(restart_options["elapsed_timestamps"])]
+
+        if not np.all(np.diff(restart_options["elapsed_timestamps"]) >= 0):
```
nicolasaunai, line 20: I don't think we should accept 0 seconds as a valid interval, do you?

pyphare/pyphare/pharein/restarts.py
```diff
@@ -9,20 +9,41 @@
...
+            raise RuntimeError(f"Error: restart_options elapsed_timestamps not in ascending order)")
+
+        seconds_in_an_hour = 60 ** 2
+        for cmp_idx, ref_ts in enumerate(restart_options["elapsed_timestamps"][1:]):
+            cmp_ts = restart_options["elapsed_timestamps"][cmp_idx]
+            if ref_ts - cmp_ts < seconds_in_an_hour:
+                raise RuntimeError(f"Error: time betweeen restart_options elapsed_timestamps must be at least one hour)")
```
nicolasaunai, line 27: why ? Not saying we should not put limits between two checkpoints, but why 1 hour?
I guess this has to do with the fact that we want a checkpoint to be dumped **before** another one is triggered? Could that be something that breaks the code if elaped_timestamps[i+1] was already passed while the last dump is not yet even finished?

nicolasaunai (commented):
Very sorry for the delay in addressing this PR.... 
Some progress now...
In the test, should we add the test of possibly several elapsed timestamps instead of just one?
I think indeed the term "timestamps" should be reserved for the "simulation times" and here maybe clock_timestamps (or other?) should be used for being clear we're talking about true wall clock time.

pyphare/pyphare/pharein/restarts.py
PhilipDeegan: I think I put one hour assuming it was a mistake on the users part, but it could be configurable too
Pretty sure restart dumps are synchronous so it finishes before the next simulation step starts

pyphare/pyphare/pharein/restarts.py
PhilipDeegan: if you only configure elapsed timestamps. it's essentially asking for the init dump, not sure it's bad

src/restarts/restarts_manager.hpp
PhilipDeegan: it means it's only setup for loading from restart, not creating any new restart dumps

pyphare/pyphare/pharein/restarts.py
PhilipDeegan: left as commented out until we figure it out

PhilipDeegan:
`gh_pr_gcc_samraisub_F38 (nico) ` failure is samrai std iterator thing

src/restarts/restarts_manager.hpp
nicolasaunai: could it mean it's neither saving nor loading (i.e. start from scratch without checkpointing)?

src/core/utilities/mpi_utils.hpp
```diff
@@ -57,6 +61,46 @@ auto mpi_type_for()
     // don't return anything = compile failure if tried to use this function
 }
 
+
+template<typename Fn, typename... Args>
+auto get_from_rank_0_only(Fn&& fn, Args&&... args)
```
nicolasaunai, line 66: `all_get_from(Fn&& fn, Args&&... args, int rank = 0)` ?

PhilipDeegan: I don't think you can do that, maybe 

```
all_get_from(int rank, Fn&& fn, Args&&... args) {}

all_get_from_rank_0(Fn&& fn, Args&&... args) {
 return all_get_from(0, ...)
}
```

src/restarts/restarts_manager.hpp
PhilipDeegan: generally yes, but we only go in this function if there's any restarts [key in the simulation dict](https://github.com/PHAREHUB/PHARE/blob/3b4644dcdb33e5dff0a7164dcd98d7e7a8753065/src/simulator/simulator.hpp#L266C68-L266C68)

essentially we don't configure the C++ side of restart writing if there is no timestamps, and loading is handled in the hierarchy

pyphare/pyphare/pharein/simulation.py
```diff
@@ -602,6 +604,7 @@ class Simulation(object):
                    restart_options={"dir": restart_outputs,
                                    "mode": "overwrite" or "conserve",
                                    "timestamps" : [.009, 99999]
+                                   "elapsed_timestamps" : [datetime.timedelta(hours=1)],
                                    "restart_time" : 99999.99999 },
```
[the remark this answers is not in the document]
PhilipDeegan: this is a docstring and those are the only acceptable values

src/core/utilities/mpi_utils.hpp
nicolasaunai: yeah ok, as you wish, the rank think is not essential. I prefer the `all_get_from....` however, it conveys better the collective aspect of the com.

PhilipDeegan: to me, that it's in the MPI namespace, kinda implies it's collective, otherwise why would it be via "mpi"
but anyway, it is now `all_get_from`

tests/simulator/test_restarts.py
```diff
@@ -162,58 +213,87 @@ def test_restarts(self, dim, interp, simInput, expected_num_levels):
...
+        self.assertEqual([seconds], ph.global_vars.sim.restart_options["elapsed_timestamps"])
+
+        assert "restart_time" not in ph.global_vars.sim.restart_options
+        model = setup_model()
+        dump_all_diags(model.populations, timestamps=np.array(timestamps))
+        simulator = Simulator(ph.global_vars.sim).initialize()
+        for i in range(restart_idx - 1):
+            simulator.advance()
+        import time
+        time.sleep(10)
```
nicolasaunai, line 277: to avoid future mistakes I think it would probably be safer to write `time.sleep(seconds+1)`

nicolasaunai (approved):

pyphare/pyphare/pharein/__init__.py
PhilipDeegan: @nicolasaunai thoughts on updating the original (non-elapsed time based) variable from `timestamps` to something clearer?

PhilipDeegan: oh you answered already below 

> I think indeed the term "timestamps" should be reserved for the "simulation times" and here maybe clock_timestamps (or other?) should be used for being clear we're talking about true wall clock time.

nicolasaunai (approved):

PR #678, opened 2022-04-19
add cellmaps and bucketlists and 2D tagging

nicolasaunai:
bug fix : add patch ghost fillling on regrid


res/cmake/def.cmake
```diff
@@ -16,7 +16,7 @@ endif(withIPO)
 
 set (PHARE_WITH_CCACHE FALSE)
 if(devMode) # -DdevMode=ON
-  set(testDuringBuild ON)
+    #set(testDuringBuild ON)
```
PhilipDeegan, line 19: this is enabled on CI even without this, so rm or uncomment?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -752,11 +763,15 @@ def plot2d(self, **kwargs):
...
-                data = data[nbrGhosts[0]:-nbrGhosts[0], nbrGhosts[1]:-nbrGhosts[1]]
-                x = np.copy(x[nbrGhosts[0]:-nbrGhosts[0]])
-                y = np.copy(y[nbrGhosts[1]:-nbrGhosts[1]])
+
+                # if nbrGhosts is 0, we cannot do array[0,-0]
+                if np.all(nbrGhosts==np.zeros_like(nbrGhosts)):
+                    x = np.copy(x)
+                    y = np.copy(y)
+                else:
+                    data = data[nbrGhosts[0]:-nbrGhosts[0], nbrGhosts[1]:-nbrGhosts[1]]
```
PhilipDeegan, line 772: maybe something like [this](https://github.com/PHAREHUB/PHARE/pull/591/files#diff-884f2783892817b4ae5e349e6ddb2741883861e14f5526ecf73854f10a6f99d1R307) could make this a little cleaner?

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -701,10 +705,10 @@ def plot1d(self, **kwargs):
                 if qty is None:
                     qty = list(patch.patch_datas.keys())[0]
 
-                layout  = patch.patch_data[qty].layout
-                nbrGhosts = layout.nbrGhostsFor(qty)
-                val = patch.patch_datas[qty].dataset[nbrGhosts:-nbrGhosts]
-                x   = patch.patch_datas[qty].x[nbrGhosts:-nbrGhosts]
+                layout  = patch.patch_datas[qty].layout
+                nbrGhosts = layout.nbrGhostFor(qty)
+                val = patch.patch_datas[qty].dataset[nbrGhosts[0]:-nbrGhosts[0]]
```
PhilipDeegan, line 710: if this is 1d . this could be 

```
val = patch.patch_datas[qty][patch.box]
```

pretty sure it handles primals as expected

pyphare/pyphare/pharesee/hierarchy.py
PhilipDeegan: or even

```
data = pdat[patch.box]
```

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -752,11 +763,15 @@ def plot2d(self, **kwargs):
...
                 x = pdat.x
                 y = pdat.y
-                nx,ny =x.size, y.size
-                data = data.reshape((nx,ny))
-                data = data[nbrGhosts[0]:-nbrGhosts[0], nbrGhosts[1]:-nbrGhosts[1]]
-                x = np.copy(x[nbrGhosts[0]:-nbrGhosts[0]])
-                y = np.copy(y[nbrGhosts[1]:-nbrGhosts[1]])
+
+                # if nbrGhosts is 0, we cannot do array[0,-0]
+                if np.all(nbrGhosts==np.zeros_like(nbrGhosts)):
```
PhilipDeegan, line 768: this is true only for tags?

src/amr/data/particles/particles_data.hpp
```diff
@@ -97,10 +105,20 @@ namespace amr
...
+        // without access to interp_order we cannot do this:
+        // static constexpr int ghostSafeMapLayer = ghostWidthForParticles<interp_order>() + 1;
+        // so we do that :
+        static constexpr int ghostSafeMapLayer = 4;
+        // which results in mapping more cells than strictly necessary but should never crash
 
     public:
         ParticlesData(SAMRAI::hier::Box const& box, SAMRAI::hier::IntVector const& ghost)
             : SAMRAI::hier::PatchData::PatchData(box, ghost)
+            , domainParticles{grow(phare_box_from<dim>(getGhostBox()), ghostSafeMapLayer)}
```
PhilipDeegan, line 117: is this like the ghostbox of the ghostbox? thinking

src/amr/data/particles/particles_data.hpp
```diff
@@ -185,9 +205,9 @@ namespace amr
                 }
 
                 assert(particles.size() == 0);
-                particles.resize(n_particles);
+                particles.reserve(n_particles);
                 for (std::size_t i = 0; i < n_particles; ++i)
-                    particles[i] = soa.copy(i);
+                    particles.push_back(soa.copy(i));
```
PhilipDeegan, line 210: this might want to be instrumented at some point to check if we're constructing more objects than we really need to

src/amr/solvers/solver_ppc.hpp
```diff
@@ -200,8 +200,15 @@ void SolverPPC<HybridModel, AMR_Types>::saveState_(level_t& level, Ions& ions, R
         auto _ = rm.setOnPatch(*patch, ions);
         for (auto& pop : ions)
         {
-            tmpDomain[ss.str() + "_" + pop.name()]  = pop.domainParticles();
-            patchGhost[ss.str() + "_" + pop.name()] = pop.patchGhostParticles();
+            auto key = ss.str() + "_" + pop.name();
+            if (!tmpDomain.count(key))
```
PhilipDeegan, line 204: is it less allocations?
seems about the same

PhilipDeegan, line 204: is it less allocations?
seems about the same

src/core/data/particles/particle_array.hpp
```diff
@@ -15,69 +22,220 @@ class ParticleArray
...
+    }
+
+    template<typename Cell>
+    void change_icell(Cell const& newCell, std::size_t particleIndex)
+    {
+        auto oldCell                    = particles_[particleIndex].iCell;
+        particles_[particleIndex].iCell = newCell;
+        if (!box_.isEmpty())
+        {
+            // std::cout << "ChangeIcell : " << particleIndex << " from " << Point{oldCell} << " to
```
PhilipDeegan, line 188: rm?

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+        {
+            auto& itemIndexes = cellIndexes_(local_(cell));
+            for (auto const& index : itemIndexes)
+            {
+                if (counts.count(index) == 0)
+                    counts[index] = 0;
+                counts[index]++;
+                if (counts[index] > 1)
+                {
+                    std::cout << "OUPS " << cell << " " << index << "\n";
```
PhilipDeegan, line 63: ?

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+                }
+            }
+        }
+        bool unique = true;
+        for (auto const& [index, count] : counts)
+            if (count > 1)
+            {
+                std::cout << "CellMap CHECKUNIQUE : " << index << "appears " << count << " times\n";
```
PhilipDeegan, line 71: ¿

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+
+    void empty();
+
+    bool is_empty() const { return size() == 0; }
+
+
+    float used_mem_ratio() const { return static_cast<float>(size()) / capacity(); }
+
+    // trim all bucketlists
+    // void trim(std::size_t max_empty);
```
PhilipDeegan, line 139: rm

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+}
+
+
+template<std::size_t dim, typename cell_index_t>
+template<typename Cells>
+inline void CellMap<dim, cell_index_t>::reserve(Cells const& cells)
+{
+    for (auto const& cell : cells)
+    {
+        // cellIndexes_.emplace(cell, Indexer{});
```
PhilipDeegan, line 294: function doesn't need to exist anymore?

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+{
+    for (auto const& cell : box_)
+    {
+        cellIndexes_(local_(cell)).empty();
+        assert(cellIndexes_(local_(cell)).is_empty());
+    }
+}
+
+
+#if 0
```
PhilipDeegan, line 411: rm?

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+
+
+template<std::size_t dim, typename cell_index_t>
+template<typename Array, typename CellIndex, typename CellExtractor>
+inline void CellMap<dim, cell_index_t>::erase(CellIndex cell, Array& items, CellExtractor extract)
+{
+    auto lastIndex     = items.size() - 1;
+    auto& itemsToErase = cellIndexes_(local_(cell));
+    assert(false);
+    //  for (auto itemIndex : itemsToErase)
```
PhilipDeegan, line 514: is this function supposed to exist?

src/core/utilities/cellmap.hpp
```diff
@@ -0,0 +1,565 @@
...
+{
+    for (auto const& cell : box_)
+    {
+        cellIndexes_(local_(cell)).sort();
+    }
+}
+
+
+#if 0
+// we don't need it now but may become handy in the future so leave it here
```
PhilipDeegan, line 544: :ok_hand:

tests/core/numerics/interpolator/test_main.cpp
```diff
@@ -249,7 +252,7 @@ class A1DInterpolator : public ::testing::Test
 
     A1DInterpolator()
         : em{"EM"}
-        , particles(1)
+        , particles{grow(layout.AMRBox(), safeLayer), 1}
```
PhilipDeegan, line 255: seems inconsistent with the use of "grow" in `particles_data.hpp`

tests/core/numerics/pusher/test_pusher.cpp
```diff
@@ -221,20 +226,22 @@ TEST_F(APusher2D, trajectoryIsOk)
...
         actual[0][i]
             = (particlesOut[0].iCell[0] + particlesOut[0].delta[0]) * static_cast<float>(dxyz[0]);
 
-        pusher->move(rangeIn, rangeOut, em, mass, interpolator, selector, layout);
+        pusher->move(rangeIn, rangeOut, em, mass, interpolator, layout, selector, selector);
 
         std::copy(rangeOut.begin(), rangeOut.end(), rangeIn.begin());
     }
 
+    std::cout << actual[0][985] << " " << expectedTrajectory.x[985] << " " << actual[0][986] << " "
```
PhilipDeegan, line 243: ?

res/cmake/def.cmake
nicolasaunai: ok deleted

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: ok replaced by 
```python
                    data = data[patch.box]
```

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: can we really throw a box at a h5 dataset? :/

nicolasaunai: this is in plot1d so I think it's ok.
btw it's nice to be able to do pd[box] but we still have to handle .x[ ] manually

pyphare/pyphare/pharesee/hierarchy.py
nicolasaunai: quite possibly yes

src/amr/data/particles/particles_data.hpp
nicolasaunai: not sure why I did this exactly.... 
so, the fact the ParticleData now needs a box is because of the cellmap not being a map anymore but a vector, so has to be preallocated to a size.... a size that is large enough for all cells possibly reached by particles...
I think this was added as a safety so that particles leaving the ghost box cells outward can still be updated in the cell map, 

I think the exact minimum value should be
        // static constexpr int ghostSafeMapLayer = ghostWidthForParticles<interp_order>() + 1;
as commented
but I don't have interporder here.... so hard coded 4 for all interporder........

src/amr/data/particles/particles_data.hpp
nicolasaunai: specific issue? (curious: what is related with your particle counter you showed me)

src/core/utilities/cellmap.hpp
nicolasaunai: yeah.... that's just for debug...

src/core/utilities/cellmap.hpp
nicolasaunai: I cautiously answer :  no and it should not...
reason is that this function asserts false, cellmap is only visible through the particleArray...
 the other overload `     auto erase(IndexRange_&& range)` in ParticleArray calls `inline void CellMap<dim, cell_index_t>::erase(Range&& range)`, which seems to remove all indexes in the range from the map, and all corresponding items from the container associated with the range....

removed... :crossed_fingers:

tests/core/numerics/pusher/test_pusher.cpp
nicolasaunai: you don't like these particles 985 and 986?
removed:)

src/amr/data/particles/particles_data.hpp
nicolasaunai: ah you mean it the safe layer could/should be only 1?

PhilipDeegan: yeah

src/amr/data/particles/particles_data.hpp
PhilipDeegan: pointless copies aren't costless,  (curious: yes)

pyphare/pyphare/pharesee/hierarchy.py
PhilipDeegan: > handle .x[ ] manually

yeah, you can do something like [this](https://github.com/PhilipDeegan/PHARE/blob/3d_base/tests/simulator/test_initialization.py#L307)

so it might be `patch.patch_datas[qty].x[select[0]]`

tests/core/numerics/interpolator/test_main.cpp
PhilipDeegan: if there is no ghosts layer then this makes sense

PhilipDeegan (approved):

PR #680, opened 2022-06-04
Get split box per ParticleArray

PhilipDeegan:


src/amr/data/particles/refine/particles_data_split.hpp
```diff
@@ -244,27 +245,23 @@ namespace amr
...
         {
-            SAMRAI::hier::Box splitBox{destinationBox};
-            SAMRAI::tbox::Dimension dimension{dim};
-            auto growingVec = SAMRAI::hier::IntVector::getZero(dimension);
+            auto splitBox{destinationBox};
+            auto growingVec = SAMRAI::hier::IntVector::getZero(SAMRAI::tbox::Dimension{dim});
 
             for (auto iDim = 0u; iDim < dim; ++iDim)
-            {
                 growingVec[iDim] = Splitter::maxCellDistanceFromSplit();
```
PhilipDeegan, line 254: could be `std::fill_n` if samrai supports it

src/amr/data/particles/refine/particles_data_split.hpp
```diff
@@ -244,27 +245,23 @@ namespace amr
...
-            splitBox.grow(growingVec);
 
+            splitBox.grow(growingVec);
             return splitBox;
         }
 
         template<typename Particle>
-        bool isCandidateForSplit_(Particle const& particle,
-                                  SAMRAI::hier::Box const& toFillBox) const
+        auto static isCandidateForSplit_(Particle const& particle,
```
PhilipDeegan, line 261: this function may no longer need to exist

nicolasaunai:
If I understand correctly, now you make the "splitBox" per destination box rather than per particle, which is probably much better.
However, we still loop over all particles of the source particle arrays to check if they are in the split box.
Do you think now we have the cellmap it is possible to use one of the overloads of `ParticleArray::export_particles` to directly get particles that are in the splitBox? If yes, that could be even better.

PhilipDeegan:
> If I understand correctly...

I don't see it as exporting particles so much as selecting particles to use in some function

nicolasaunai:
AFAIR currently what we do is : 
- loop over **all** particles of the source arrays
- check if the current particle is in the split area
- if yes split it into an array of N refined particles
- copy these refined particles in the appropriate array


what I suggested is to replace this by:
- export all particles from source arrays that fall into split box, using the cellmap export (into a pre-allocated buffer ideally)
- for each of the bufferized particles: split it into N refined particles
- copy these refined particles in the appropriate array

does it make sense?

PR #682, opened 2022-06-16
return distribution interpolator

nicolasaunai:
enable doing cuts in distribution functions by now returning an interpolator (with bin centered coordinates for convenience)

### DIFF
--- a/pyphare/pyphare/pharesee/plotting.py
+++ b/pyphare/pyphare/pharesee/plotting.py
@@ -35,6 +35,7 @@ def dist_plot(particles, **kwargs):
     return value : fig,ax
     """
     from pyphare.pharesee.particles import Particles, aggregate
+    from scipy.interpolate import interp2d
 
     if isinstance(particles, list):
         particles = aggregate(particles)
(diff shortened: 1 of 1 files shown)


pyphare/pyphare/pharesee/plotting.py
```diff
@@ -35,6 +35,7 @@ def dist_plot(particles, **kwargs):
     return value : fig,ax
     """
     from pyphare.pharesee.particles import Particles, aggregate
+    from scipy.interpolate import interp2d
```
rochSmets, line 38: I have no religion... but for the other interpolator in `run.py`
https://github.com/PHAREHUB/PHARE/blob/91e0830b0131ebf31677d8a8b733fee542600d74/pyphare/pyphare/pharesee/run.py#L119-L120
we use different ones.
`interp2d` can only be linear, cubic or quintic... while we can have nearest. but the format to use for the coordinate is not the same. we can keep it as is, but then think about it later one  if any associated problem is arising.

rochSmets (approved):

PR #688, opened 2022-06-29
fix function pointer argument (gcc12 bug)

nicolasaunai:


kolegram (approved):
I thought that a ref did not need the `&` char to be specified... but if it goes well this way, and it is the proper way to do it, that's fine to me.

PhilipDeegan (approved):

PR #696, opened 2022-07-22
diagnostics timestamps validation improvements

PhilipDeegan:
a few tests to add to test_validation.py

### DIFF
--- a/pyphare/pyphare/pharein/diagnostics.py
+++ b/pyphare/pyphare/pharein/diagnostics.py
@@ -40,24 +40,23 @@ def diagnostics_checker(func):
 
 import numpy as np
 # ------------------------------------------------------------------------------
-def validate_timestamps(clazz, **kwargs):
+def validate_timestamps(clazz, key, **kwargs):
     sim = global_vars.sim
 
     init_time = sim.start_time()
 
-    for key in ["write_timestamps", "compute_timestamps"]:
-        timestamps = kwargs[key]
+    timestamps = phare_utilities.np_array_ify(kwargs.get(key, []))
 
-        if np.any(timestamps < init_time):
-            raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be less than simulation.init_time({init_time}))")
-        if np.any(timestamps > sim.final_time):
-            raise RuntimeError(f"Error: timestamp({sim.time_step_nbr}) cannot be greater than simulation.final_time({sim.final_time}))")
-        if not np.all(np.diff(timestamps) >= 0):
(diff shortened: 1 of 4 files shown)


PhilipDeegan:
need to handle the diags_dir trailing / messing things up

pyphare/pyphare/pharein/simulation.py
```diff
@@ -395,6 +395,10 @@ def check_restart_options(**kwargs):
         if "mode" not in restart_options:
             raise ValueError (f"Restart mode not set, valid modes are {valid_modes}")
 
+        if "dir" in restart_options:
+            while restart_options["dir"][-1] == "/": # samrai can't handle this
```
PhilipDeegan, line 399: maybe should be 

```
while len(restart_options["dir"]) > 1 restart_options["dir"][-1] == "/
```

?

PhilipDeegan: actually 

```
while "//" in restart_options["dir"]: restart_options["dir"] = restart_options["dir"].replace("//", "/")
```

PhilipDeegan: or maybe just raise and say it's invalid

nicolasaunai: "aaa/bbb///".rstrip("/")

nicolasaunai: IMO we should also do other checks such as permissions (https://www.tutorialspoint.com/How-to-check-the-permissions-of-a-directory-using-Python), other things?

PhilipDeegan: check there's some space left on the disk maybe

nicolasaunai: probably not straightforward with the quotas

pyphare/pyphare/pharein/diagnostics.py
```diff
@@ -145,7 +143,7 @@ def to_dict(self):
 # ------------------------------------------------------------------------------
 
 def population_in_model(population):
-    return population in [p for p in global_vars.sim.model.populations] + ["all","ions"]
+    return population in [p for p in global_vars.sim.model.populations]
```
nicolasaunai, line 146: don't remember why we dropped those. I think it was causing a bug somewhere?

nicolasaunai (approved):

pyphare/pyphare/pharein/diagnostics.py
PhilipDeegan: yeah, I think there was something

Issue #693, opened 2022-07-16
coarsen moments too


nicolasaunai:
At the moment only E, B and particles are regrided (copied+refined).
Moments (n,V, fluxes) are recomputed from regrided particles on the new level once regridding is done. 
But regriding occurs after level synchronisation. Thus the coarsened density and bulk velocity are replaced by the local self-consistent particle deposit when regriding occurs... This potentially is a source of density/V mismatch between a fine level and its next coarser density/V.

Possible fix:
- n and V are needed for next time step to compute Ohm so we need to get them at the end of regriding
- the idea is to regrid them as we do for E and B, and not recompute them form just obtained particles.
-  this could be even faster since at the moment, computing n and V needs a (heavy) deposition of all densities, all fluxes and then compute the total n and total bulk vel.

notes :
- moments are computed from particles, even on border nodes. Order or operations (domain then ghosts) may result in slight mismatch in moments on shared nodes. This mismatch was seen before to result in E mismatch and then B mimatch and was solved by sharing border node values of E and B for one value to take precedence, before sharing overlapped values
- without sync moments, any mismatch in moments could thus not propagate errors
- with n and V coarsened, mismatch values may leak to next coarser domain nodes and possibly result in coarsening test failing

nicolasaunai:
done in #697

PR #697, opened 2022-07-26
sync n and v

nicolasaunai:
see #693 

This PR enables the syncing of the total ion density and total ion bulk velocity in the coarsening ops.
Coarsening just E and B is not enough. If moments are not synced, E gets re-computed on next step with unsynced moments, any possible divergence in particle dynamics across levels will result in diverging moments and thus diverging E which enhances the particle dynamic divergence etc.
The PR also adds rho and V in the coarsening test.

Because of the 3-nodes coarsening stencil, the first primal ghost node is needed for n and V. Before this PR, n&V were NOT complete on any ghost node from particle deposition, because the particle ghost area is not wide enough.
The first ghost node was assigned a copy of the shared border primal node (for no good reason) and the other nodes were assigned NaN as they should be unused. This causes a coarsening issue since two adjacent procs would not share the same value, i.e. first primal ghost on proc1 would be a copy of border node (of proc 1 and 2), and not equal to the overlaped domain node of proc2.
This PR thus now fills ghosts for n and V with a regular ghost field schedule. Not this schedule does NOT touch the shared border node (through one mode of the FieldVariableFillPattern/FieldGeometry). The shared border node was and still is complete anyway, from the deposit of domain particles and the patch or level ghost (old and new) particles.

Fluid tests are changed so not to expect NaNs and not to expect the copy of the shared border node

### DIFF
--- a/src/amr/data/field/field_geometry.hpp
+++ b/src/amr/data/field/field_geometry.hpp
@@ -2,8 +2,11 @@
 #define PHARE_SRC_AMR_FIELD_FIELD_GEOMETRY_HPP
 
 #include <cassert>
+#include <iostream>
 
+#include "SAMRAI/hier/IntVector.h"
 #include "core/data/grid/gridlayoutdefs.hpp"
+#include "core/hybrid/hybrid_quantities.hpp"
 #include "core/utilities/types.hpp"
 #include "core/data/grid/gridlayout.hpp"
 #include "core/data/grid/gridlayout_impl.hpp"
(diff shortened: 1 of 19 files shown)


src/amr/data/field/field_geometry.hpp
```diff
@@ -14,68 +17,76 @@
 #include <SAMRAI/hier/BoxGeometry.h>
 
 
-namespace PHARE::amr
+namespace PHARE
```
PhilipDeegan, line 20: `namespace PHARE::amr`

for less indentation?

src/amr/data/field/field_geometry.hpp
```diff
@@ -14,68 +17,76 @@
...
+    private:
+        static SAMRAI::hier::Box
+        pureInteriorBox_(SAMRAI::hier::Box const& interiorFieldBox,
+                         std::array<core::QtyCentering, dimension> const& centerings)
+        {
+            auto noSharedNodeBox{interiorFieldBox};
+            SAMRAI::hier::IntVector growth(SAMRAI::tbox::Dimension{dimension});
+            for (auto dir = 0u; dir < dimension; ++dir)
+            {
+                std::cout << static_cast<int>(centerings[dir]) << " and "
```
PhilipDeegan, line 68: you can probably go back to your `? 1 : 0` version

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
```diff
@@ -34,6 +34,35 @@ namespace PHARE
...
+        {
+        }
+        std::string name;
+        FieldT* f;
+        typename HybridQuantity::Scalar quantity;
+        using field_type = FieldT;
+
+        std::vector<Property> getFieldNamesAndQuantities() const { return {{name, quantity}}; }
+
+        void setBuffer([[maybe_unused]] std::string const& bufferName, FieldT* field) { f = field; }
```
PhilipDeegan, line 61:
```
void setBuffer(std::string const& /*bufferName*/, FieldT* field) { f = field; }
```

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
```diff
@@ -501,14 +556,23 @@ namespace amr
...
             {
                 auto dataOnPatch = resourcesManager_->setOnPatch(
-                    *patch, hybridModel.state.electromag, hybridModel.state.J, EM_old_, Jold_);
+                    *patch, hybridModel.state.electromag, hybridModel.state.J,
+                    hybridModel.state.ions, EM_old_, Jold_, NiOldUser_, ViOld_);
+
                 resourcesManager_->setTime(EM_old_, *patch, currentTime);
                 resourcesManager_->setTime(Jold_, *patch, currentTime);
+                resourcesManager_->setTime(NiOldUser_, *patch, currentTime);
+                resourcesManager_->setTime(ViOld_, *patch, currentTime);
```
PhilipDeegan, line 565: maybe overkill but this could be a tuple apply

something like

```
core::apply(std::forward_as_tuple(EM_old_, Jold_,NiOldUser_, ViOld_), [&](auto& item){
  resourcesManager_->setTime(item, *patch, currentTime);
});
```

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
```diff
@@ -501,14 +556,23 @@ namespace amr
...
 
                 auto& EM = hybridModel.state.electromag;
                 auto& J  = hybridModel.state.J;
+                auto& Vi = hybridModel.state.ions.velocity();
+                auto& Ni = hybridModel.state.ions.density();
+
                 EM_old_.copyData(EM);
                 Jold_.copyData(J);
+                ViOld_.copyData(Vi);
+                NiOldUser_.f->copyData(Ni);
```
PhilipDeegan, line 575: maybe implement `copyData` on FieldUser?

tests/simulator/test_advance.py
```diff
@@ -380,31 +403,32 @@ def _test_field_coarsening_via_subcycles(self, dim, interp_order, refinement_box
...
+                                if dim == 1:
+                                    afterCoarse[dataBox.lower[0] : dataBox.upper[0] + 1] = -144123
+                                if dim == 2:
+                                    afterCoarse[dataBox.lower[0] : dataBox.upper[0] + 1,
+                                                dataBox.lower[1] : dataBox.upper[1] + 1] = -144123
+
+                                coarsen(qty, coarse_pd, fine_pd, coarseBox, fine_pdDataset, afterCoarse)
+                                # 1e-16 seems to fail for some reason. Discrepency appears to be
+                                # on refined adjacent patches. Was way worse (1e-4 or so) when not
+                                # filling pure ghosts for Ni and Vi. 1e-15 seems food enough
```
PhilipDeegan, line 430: 🍔

src/amr/data/field/field_geometry.hpp
nicolasaunai: sure, I'm always a bit reluctant to do so because it makes a big diff for nothing

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
nicolasaunai: curious : what's your reason for that VS maybe_unused? (in this case we know it's not maybe?)

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
nicolasaunai: appealing but I think I prefer to keep it "simple"

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
nicolasaunai: yes

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
PhilipDeegan: > in this case we know it's not maybe?

yeah and it's slightly easier on the eyes

src/amr/data/field/field_geometry.hpp
PhilipDeegan: in this case you're making a diff by adding indentation ;)

PhilipDeegan (approved):

nicolasaunai:
at this point there are two errors only visible on the clang build and only in parallel:

1/  `Read -1, expected 1432704, errno = 1`  shows up on all `test_fields_advance_2d.py AdvanceTest.test_field_coarsening_via_subcycles_xxx` tests, for some reason. I can't reproduce that locally on pharemer

2/ the absolute 3.5e-15 error for `test_overlaped_fields_are_equal_09`  is too small for 1 node apparently :

```
 Mismatched elements: 1 / 261 (0.383%)
  Max absolute difference: 3.55271368e-15
  ```

considering the test says :+1: 
```
                        # empirical max absolute observed 3.33e-15
                        # seems correct considering ghosts are filled with schedules
                        # involving linear/spatial interpolations and so on where
                        # rounding errors may occur.... setting atol to 3.5e-15
```

I am raising the threshold to 4e-15.

PR #701, opened 2022-09-10
monitor system resources during simulation

PhilipDeegan:
currently it stores data throughout the entire simulation and writes the files at the end,
this might not be ideal, depends on the total simulation time and monitoring interval.
also currently it will run per process, where we can probably do something purely on rank 0 to monitor all ranks


PhilipDeegan:
impl moved to https://github.com/PhilipDeegan/phlop/blob/master/phlop/app/stats_man.py
client code to be considered in phare later

PR #702, opened 2022-09-20
splitbox per box not particle

PhilipDeegan:
it's possible the optimized binaries are the same before/after, but the intention is clearer with the change

### DIFF
--- a/src/amr/data/particles/refine/particles_data_split.hpp
+++ b/src/amr/data/particles/refine/particles_data_split.hpp
@@ -171,6 +171,7 @@ namespace amr
             for (auto const& destinationBox : destBoxes)
             {
                 std::array particlesArrays{&srcInteriorParticles, &srcGhostParticles};
+                auto splitBox = getSplitBox(destinationBox);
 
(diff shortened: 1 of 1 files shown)


nicolasaunai (commented):
given that now the `isCandidateForSplit_` private method is 1 line, what do you think to even remove it and change the call site into : 

```cpp
if (isInBox(candidateRefinedPart, candidateSplitBox))
                        {
                            split(candidateRefinedPart, refinedParticles);
```

PhilipDeegan:
> given that now the `isCandidateForSplit_` private method is 1 line, what do you think to even remove it and change the call site into :
> 
> ```c++
> if (isInBox(candidateRefinedPart, candidateSplitBox))
>                         {
>                             split(candidateRefinedPart, refinedParticles);
> ```

sure

PhilipDeegan:
@nicolasaunai [how bout](https://github.com/PHAREHUB/PHARE/compare/master...PhilipDeegan:PHARE:box_grow) ?

nicolasaunai (approved):

Issue #658, opened 2022-02-14
wrong time interpolation of levelghosts


nicolasaunai:
https://github.com/PHAREHUB/PHARE/blob/d9cbcdf0e17125d42eda72dfdcb7cd9e8101765a/src/amr/solvers/solver_ppc.hpp#L575

times given here results in the interpolation coef calculated here https://github.com/PHAREHUB/PHARE/blob/master/src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp#L711

to always be dt_fine/dt_coarse == 1/4.

it should rather be (newTime - coarserLastTime)/dt_coarse  to represent the fraction of the time elapsed on current level since last sync time with next coarse.

As is (since may 2020), levelghostParticleOld always see their contribution to be 0.75 and levelGhostNew 0.25.

nicolasaunai:
done in #704

PR #704, opened 2022-09-29
fix levelghost particle interpolation time #658

nicolasaunai:
issue #658 
time interpolation coef does not need to know about current time, just newTime.
The coef is given as the ratio between the time elapsed in the current subcycle (since last sync) and the total dt of the next coarser.


PhilipDeegan (approved):

rochSmets (approved):
that's fine for me,

Issue #673, opened 2022-03-09
Ion Beam test criterion


nicolasaunai:
The ion streaming instability test asserts the growth rate is close to the expected theoretical value.

```python
 assert np.fabs(gamma-0.09) < 2e-2
 ```

Most of the time, it works, like [this build](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_GhMasterGccSamraisubNightly/74318?showRootCauses=false&expandBuildChangesSection=true)

And produces [this figure](https://hephaistos.lpp.polytechnique.fr/teamcity/repository/download/Phare_Phare_GhMasterGccSamraisubNightly/74318:id/functional.png.zip!/ionIonBeam/ion_ion_beam1d.png) : 

![image](https://user-images.githubusercontent.com/3200931/157396372-dbbdab15-443a-4b0f-adf6-7f3e48107237.png)

Sometime it fails : 

![image](https://user-images.githubusercontent.com/3200931/157395980-ccc4e752-8ea7-4f30-ac4e-4c8cd0f66d04.png)

[figure here](https://hephaistos.lpp.polytechnique.fr/teamcity/repository/download/Phare_Phare_GhMasterGccSamraisubNightly/74372:id/functional.png.zip!/ionIonBeam/ion_ion_beam1d.png)

[from build here](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_GhMasterGccSamraisubNightly/74372?showRootCauses=false&expandBuildChangesSection=true&expandBuildProblemsSection=true)

another fail example ([from this build](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_GhMasterGccSamraisubNightly/74388?showRootCauses=false&expandBuildChangesSection=true&expandBuildProblemsSection=true)): 

![image](https://user-images.githubusercontent.com/3200931/157396777-45fb20b3-9b8c-4969-a2a5-c32a169a66d5.png)

([figure here](https://hephaistos.lpp.polytechnique.fr/teamcity/repository/download/Phare_Phare_GhMasterGccSamraisubNightly/74388:id/functional.png.zip!/ionIonBeam/ion_ion_beam1d.png)

[...]

PR #705, opened 2022-09-30
Fix ion_ion_Beam1d test

loic-darrieu:
Issue [#673](https://github.com/PHAREHUB/PHARE/issues/673)
The fit interval is now set dynamically to exclude the non-linear phase.
The output figure is a plot of the fit and a plot of the phase space and Ey at key times.

### DIFF
--- a/tests/functional/ionIonBeam/ion_ion_beam1d.py
+++ b/tests/functional/ionIonBeam/ion_ion_beam1d.py
@@ -1,9 +1,8 @@
-#!/usr/bin/env python3
 
 import pyphare.pharein as ph #lgtm [py/import-and-import-from]
 from pyphare.pharein import Simulation
 from pyphare.pharein import MaxwellianFluidModel
-from pyphare.pharein import ElectromagDiagnostics, FluidDiagnostics
+from pyphare.pharein import ElectromagDiagnostics, FluidDiagnostics, ParticleDiagnostics
 from pyphare.pharein import ElectronModel
 from pyphare.simulator.simulator import Simulator
 from pyphare.pharein import global_vars as gv
@@ -18,6 +17,10 @@ from pyphare.pharein.global_vars import sim
 import matplotlib.pyplot as plt
 import matplotlib as mpl
 import numpy as np
+from scipy.optimize import curve_fit
+from scipy.signal import find_peaks
+from pyphare.pharesee import plotting
+from pathlib import Path
 mpl.use('Agg')
 
 
@@ -30,8 +33,8 @@ def config():
     Simulation(
         smallest_patch_size=20,
         largest_patch_size=60,
-        final_time=50,
-        time_step=0.0005,
+        final_time=100,
+        time_step=0.001,
         boundary_types="periodic",
         cells=165,
         dl=0.2,
@@ -52,10 +55,10 @@ def config():
         return 1.
 
     def by(x):
-        return 0.
+        return 0
 
     def bz(x):
-        return 0.
+        return 0
 
     def vB(x):
         return 5.
@@ -69,7 +72,7 @@ def config():
 
     vMain = {
         "vbulkx": v0, "vbulky": v0, "vbulkz": v0,
-        "vthx": vth, "vthy": vth, "vthz": vth
+        "vthx": vth, "vthy": vth, "vthz": vth,
     }
 
 
@@ -89,14 +92,16 @@ def config():
 
     sim = ph.global_vars.sim
 
-    timestamps = np.arange(0, sim.final_time, 1.)
+    timestamps = np.arange(0, sim.final_time, 0.1)
 
-    for quantity in ["B"]:
+    for quantity in ["B","E"]:
         ElectromagDiagnostics(
             quantity=quantity,
             write_timestamps=timestamps,
             compute_timestamps=timestamps,
         )
+    ParticleDiagnostics(quantity = "domain", population_name = "main" , write_timestamps=timestamps, compute_timestamps=timestamps)
+    ParticleDiagnostics(quantity = "domain", population_name = "beam" , write_timestamps=timestamps, compute_timestamps=timestamps)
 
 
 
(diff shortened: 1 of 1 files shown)


tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -52,10 +55,10 @@ def bx(x):
         return 1.
 
     def by(x):
-        return 0.
+        return 0
```
rochSmets, line 58: not sure why the return should be `0` and not `0.`

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -106,14 +111,14 @@ def yaebx(x, a, b):
     return a*np.exp(np.multiply(b, x))
 
 
-def growth_b_right_hand(run_path):
+def growth_b_right_hand(run_path,d):
+
     file = os.path.join(run_path, "EM_B.h5")
     times = get_times_from_h5(file)
-
+    dt = times[1]-times[0]
```
rochSmets, line 118: this only works because `times` is coming from an `arange`. I would rather prefer `dt_stamps = 0.1` and use it at line 95 and 118

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -125,16 +130,19 @@ def growth_b_right_hand(run_path):
...
 
         # get the mode 1, as it is the most unstable in a box of length 33
         mode1 = np.absolute(np.fft.fft(by-1j*bz)[1])
         first_mode = np.append(first_mode, mode1)
+    
+    imax = find_peaks(first_mode,width = int(d/dt))[0][0]
+    
 
-    popt, pcov = curve_fit(yaebx, times, first_mode, p0=[0.08, 0.09])
+    popt, pcov = curve_fit(yaebx, times[:imax-int(d/dt)], first_mode[:imax-int(d/dt)], p0=[0.08, 0.09])
```
rochSmets, line 142: the `int(d/dt)` value is multiply used : i suggest `ioffset = int(d/dt)`

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -143,52 +151,85 @@ def growth_b_right_hand(run_path):
...
-    print(0.5*(omegas[1:omegas.size//2].argmax()+1)*2*np.pi/times[-1])
 
     return times, first_mode, popt[0], popt[1], damped_mode, omega
 
-
 def main():
-    from pybindlibs.cpp import mpi_rank
 
+    from pybindlibs.cpp import mpi_rank
+    d = 10 # The exponential fit is performed on [0, times[imax] - d]
```
rochSmets, line 162: instead of `d`, could it be `time_offset`... and say why we have such an offset (I dunno)

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -125,16 +130,19 @@ def growth_b_right_hand(run_path):
         x = xyz_finest[0][:-1]
 
         by = by_interpolator(x)
-        bz = by_interpolator(x)
+        bz = bz_interpolator(x)
 
         # get the mode 1, as it is the most unstable in a box of length 33
         mode1 = np.absolute(np.fft.fft(by-1j*bz)[1])
         first_mode = np.append(first_mode, mode1)
+
```
rochSmets, line 138: add a comment :
the find_peaks method is used to identify the max value of the time serie of the Fourier mode m=1. this max being an overshoot, the width of this overshoot is delta_t=10, which is an empirical value

tests/functional/ionIonBeam/ion_ion_beam1d.py
rochSmets, line 142: add comment
the curve_fit is performed from time index 0 to imax-ioffset as this offset prevent to use the final part of the curve which is no more exponential as this is the end of the linear mode

rochSmets (commented):
any other comments before I do what I suggested ?

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -143,52 +151,85 @@ def growth_b_right_hand(run_path):
...
-
-        ax2.plot(times, damped_mode, color='g', linestyle='', marker='o')
-        ax2.set_xlabel("Time")
-        ax2.set_ylabel("Real mode")
-        ax2.text(0.48, 0.30, "~ 3 periods until t=50", transform=ax2.transAxes)
-        ax2.text(0.40, 0.20, "omega (real) = {:5.3f}... expected 0.19".format(omega), transform=ax2.transAxes)
-
+                = growth_b_right_hand(os.path.join(os.curdir, "ion_ion_beam1d"),d)
+
+        dt = times[1]-times[0]
```
loic-darrieu, line 176: Should be replaced by dt_stamps = 0.1 to be consistant with line 118's proposition

tests/functional/ionIonBeam/ion_ion_beam1d.py
```diff
@@ -143,52 +151,85 @@ def growth_b_right_hand(run_path):
...
+        ax1.set_title("Time evolution of the first right-hand circular mode amplitude")
+        ax1.plot(times, first_mode, color = 'k',label ="|$\hat{B}_y (m=1,t)-i\hat{B}_z (m=1,t)$|")
+        ax1.plot(times[:imax], yaebx(times[:imax], ampl, gamma), color='r', linestyle='-', label ="$B_0. \exp(\gamma t), \ with\ \gamma =${:5.5f} (expected 0.09)".format(gamma))
+        ax1.axvline(0, 0, yaebx(times[imax],ampl,gamma), color = 'red',linestyle = '--')
+        ax1.axvline(times[imax]-d, 0, yaebx(times[imax],ampl,gamma), color = 'red',linestyle = '--')
+        ax1.legend()
+        ax1.set_xlabel("t - Time")
+    
+        r = Run(os.path.join(os.curdir, "ion_ion_beam1d"))
+        dt = times[1]-times[0]
```
loic-darrieu, line 189: Redundant with line 176 so should be removed

tests/functional/ionIonBeam/ion_ion_beam1d.py
loic-darrieu: There's no particular reason indeed, both work so we could keep `0.`

nicolasaunai:
closed after merging #726

PR #713, opened 2023-01-09
add divB to run getters and cell_custom for cell centered user custom…

nicolasaunai:


pyphare/pyphare/pharesee/run.py
```diff
@@ -89,6 +89,30 @@ def _compute_current(patch):
                 {"name":"Jy", "data":Jy, "centering":centering["Jy"]},
                 {"name":"Jz", "data":Jz,"centering":centering["Jz"]})
 
+
+def _divB2D(Bx,By, xBx, yBy):
+    dxbx = (Bx[1:,:]-Bx[:-1,:])/(xBx[1]-xBx[0])
```
rochSmets, line 94: I'm never very happy with calculating a dx as the diff of value [0] and [1]. as a patch is defined on a grid, why dont we have a dedicated data like a tuple `(dx, dy, dz)` ?

pyphare/pyphare/pharesee/run.py
```diff
@@ -89,6 +89,30 @@ def _compute_current(patch):
...
+        raise ValueError("divB is 0 by construction in 1D")
+
+    elif patch.box.ndim == 2:
+        By  = patch.patch_datas["By"].dataset[:]
+        Bx  = patch.patch_datas["Bx"].dataset[:]
+        xBx = patch.patch_datas["Bx"].x
+        yBy = patch.patch_datas["By"].y
+        divB = _divB2D(Bx,By, xBx, yBy)
+
+        return ({"name":"cell_custom", "data":divB, "centering":"dual"},)
```
rochSmets, line 110: Why is the centering a single value ? for `ndim=2`, the 3 components of the B fields depends on `x` and `y`, so do the associated div B. I would thus expect `"centering" : ("dual", "dual")` no ?

rochSmets (approved):

pyphare/pyphare/pharesee/run.py
nicolasaunai: is now resolved in #714. Closing this PR since #714 has this code too.

PR #714, opened 2023-01-13
hierarchy of MPI processes ranks

nicolasaunai:
is on top of #713 

```python
ranks = run.GetRanks(50)

vmax = 40
fig, (ax1,ax2, ax3) = plt.subplots(nrows=3, figsize=(10,20))
ranks.plot(qty="cell_custom", vmin=0, vmax=vmax, plot_patches=True, ax=ax1, 
        cmap="tab20", patchcolors=['k','k','k'], ls=["-", "--", "-"],levels=(0,),
       lw=[0.5,0.3,0.4])#, xlim=(80,130),ylim=(94,130))
ranks.plot(qty="cell_custom", vmin=0, vmax=vmax, plot_patches=True, ax=ax2, 
        cmap="tab20", patchcolors=['k','k','k'], ls=["-", "--", "-"],levels=(1,),
       lw=[0.5,0.3,0.4])#, xlim=(80,130),ylim=(94,130))


ranks.plot(qty="cell_custom", vmin=0, vmax=vmax, plot_patches=True, ax=ax3, 
        cmap="tab20", patchcolors=['k','k','k'], ls=["-", "--", "-"],levels=(2,),
       lw=[0.5,0.3,0.4])
fig.tight_layout()
```

image

image

image

### DIFF
--- a/pyphare/pyphare/core/gridlayout.py
+++ b/pyphare/pyphare/core/gridlayout.py
@@ -19,7 +19,6 @@ yee_centering = {
         'rho':'primal', 'P':'primal',
         'Vthx':'primal', 'Vthy':'primal', 'Vthz':'primal',
         'tags':'dual',
-        'cell_custom':'dual'
     },
     'y' : {
         'Bx':'dual', 'By':'primal', 'Bz':'dual',
@@ -30,7 +29,6 @@ yee_centering = {
         'rho':'primal', 'P':'primal',
         'Vthx':'primal', 'Vthy':'primal', 'Vthz':'primal',
         'tags':'dual',
-        'cell_custom':'dual'
     },
     'z' : {
         'Bx':'dual', 'By':'dual', 'Bz':'primal',
@@ -41,7 +39,6 @@ yee_centering = {
         'rho':'primal', 'P':'primal',
         'Vthx':'primal', 'Vthy':'primal', 'Vthz':'primal',
         'tags':'dual',
-        'cell_custom':'dual'
     }
 }
 yee_centering_lower = {
@@ -65,14 +62,16 @@ class YeeCentering(object):
         self.centerZ = yee_centering["z"]
 
 
-def yeeCoordsFor(origin, nbrGhosts, dl, nbrCells, qty, direction, withGhosts :int =False):
+def yeeCoordsFor(origin, nbrGhosts, dl, nbrCells, qty, direction, withGhosts =False, **kwargs):
 
     assert direction in direction_to_dim, f"direction ({direction} not supported)"
-    assert qty in yee_centering[direction] or qty in yee_centering_lower[direction], f"qty ({qty} not supported)"
     if qty in yee_centering_lower[direction] and qty not in yee_centering[direction]:
         qty = qty[0].upper() + qty[1:]
 
-    centering = yee_centering[direction][qty]
+    if "centering" in kwargs:
+        centering = kwargs["centering"]
+    else:
+        centering= yee_centering[direction][qty]
 
     offset = 0
     dim = direction_to_dim[direction]
@@ -284,7 +283,7 @@ class GridLayout(object):
         return x
 
 
-    def yeeCoordsFor(self, qty, direction, withGhosts=True):
+    def yeeCoordsFor(self, qty, direction, withGhosts=True, **kwargs):
         """
         from a qty and a direction, returns a 1d array containing
         the coordinates where the qty is defined, including the ghost nodes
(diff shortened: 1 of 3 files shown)


pyphare/pyphare/pharesee/run.py
```diff
@@ -89,6 +89,42 @@ def _compute_current(patch):
...
+
+        return ({"name":"cell_custom", "data":divB, "centering":"dual"},)
+
+    else:
+        raise RuntimeError("dimension not implemented")
+
+
+
+def _get_rank(patch):
+    if patch.box.ndim == 2:
```
PhilipDeegan, line 118: little comment maybe

> # for setting color  of the domain per patch using the patch id

pyphare/pyphare/pharesee/run.py
```diff
@@ -89,6 +89,42 @@ def _compute_current(patch):
...
+        divB = _divB2D(Bx,By, xBx, yBy)
+
+        return ({"name":"cell_custom", "data":divB, "centering":"dual"},)
+
+    else:
+        raise RuntimeError("dimension not implemented")
+
+
+
+def _get_rank(patch):
```
PhilipDeegan, line 117: using `rank` could be slightly confusing in the realm of MPI

PhilipDeegan (approved):

pyphare/pyphare/pharesee/run.py
nicolasaunai: I don't understand your comment, it is indeed the MPI rank that's extracted.

PhilipDeegan: ok I thought it was per patch id

pyphare/pyphare/pharesee/run.py
```diff
@@ -89,6 +89,42 @@ def _compute_current(patch):
...
+    if patch.box.ndim == 2:
+        shape = patch.box.shape
+        ghost_shape = (shape[0]+5, shape[1]+5)
```
rochSmets, line 120: I know we made this choice that we have 5 ghost points... but it would be nicer to write this value somewhere (close to refinement_ratio=2 for example) so that it appears more explicit

rochSmets (approved):

pyphare/pyphare/pharesee/run.py
nicolasaunai: ok TODO, I did this because passing the shape as is didn't work but I have not thought that it was not +5-5 anymore so why does it work as is.... to investigate

nicolasaunai: ok this is now fixed. The issue was that FieldData.x (or y, z) calls layout.yeeCoordsFor() which assumes withGhosts=True so not making a dataset that has the shape of the ghost box results in a shape mismatch with the coordinates when calling pcolormesh in plot2d.

nicolasaunai:
I also took the opportunity of this PR to solve an inconvenience when making a hierarchy with compute_hier_from.
When doing this, one can build a hierarchy for any custom quantity. When doing that, ex. for ranks here, or divB, one typically wants to provide an appropriate name in `({"name":"divB", "data":divB, "centering":["dual","dual"]},)`

but that resulted in an error, since plot2d and yeeCoordsFor assumed that the field name had to be in the yee_centerings dict in gridlayout. This is now solved

pyphare/pyphare/core/gridlayout.py
```diff
@@ -62,14 +62,16 @@ def __init__(self):
...
     assert direction in direction_to_dim, f"direction ({direction} not supported)"
-    assert qty in yee_centering[direction] or qty in yee_centering_lower[direction], f"qty ({qty} not supported)"
     if qty in yee_centering_lower[direction] and qty not in yee_centering[direction]:
         qty = qty[0].upper() + qty[1:]
 
-    centering = yee_centering[direction][qty]
+    if "centering" in kwargs:
+        centering = kwargs["centering"]
+    else:
+        centering= yee_centering[direction][qty]
```
PhilipDeegan, line 74: ->
 
```
centering = kwargs.get("centering", yee_centering[direction][qty])
```

pyphare/pyphare/core/gridlayout.py
```diff
@@ -288,8 +295,13 @@ def yeeCoordsFor(self, qty, direction, withGhosts=True):
         if qty in yee_centering_lower[direction] and qty not in yee_centering[direction]:
             qty = qty[0].upper() + qty[1:]
 
-        return yeeCoordsFor(self.origin, self.nbrGhosts(self.interp_order, yee_centering[direction][qty]),
-                            self.dl, self.box.shape, qty, direction, withGhosts=withGhosts)
+        if "centering" in kwargs:
+            centering = kwargs["centering"]
+        else:
+            centering= yee_centering[direction][qty]
```
PhilipDeegan, line 301: same here for `centering = kwargs.get(...)`

nicolasaunai: believe it or not that's what I did first, but that didn't work and raised key error because `qty` is not in the dict...

rochSmets (approved):
:+1:

PR #735, opened 2023-07-12
magnetic flux across grids

nicolasaunai:
## Add a new way to Coarsen a field, specialized for the magnetic field: 

- adds magnetic field coarsener
   - caveat : this is not 3D  
- previous coarsener becomes the `DefaultFieldCoarsener` and is used for the density and bulk velocity
- `FieldCoarserOperator`now takes a `FieldCoarsenerPolicy` as a template parameter

## Adding specific ways to refine the magnetic and electric fields: 

- `FieldRefineOperator` now takes a `FieldRefinerPolicy` as a template parameter, which is used in `refine()` to fill the fine filed given the coarse field and a coarse index.
- add objects `ElectricFieldRefiner` and `MagneticFieldRefiner`
- previous `FieldRefiner` now becomes `DefaultFieldRefiner` and is used by density and bulk velocity

## Conserving magnetic flux at level interfaces

- magnetic field components are defined on cell faces and its value on a face only depends on the electric components on the edges of that face. Hence knowing the electric field everywhere on the ghost box we can calculate B everywhere on the ghost box. 
    - Hence, B is now calculated over the whole **ghost** box using faraday.
    - This adds `evalOnGhostBox` in `GridLayout` which is called in `Faraday` instead of `evalOnBox`
    - Magnetic field ghosts are not filled anymore. the messenger method fillMagneticGhost is removed.
    - magnetic patch ghosts **only** are filled in `postSynchronize`. This is needed because reset of field data by coarsened values may break overlap consistency with neighbor patches. 
        - This uses the new `RefinerType::PatchGhostField` in `refiner_pool.hpp`
        - This adds a new type of schedule created for intra level only, associated with constexpr `RefinerType::PatchGhostField`
        - The hybridModel is the only one registering a magnetic field in `ghostMagnetic` (as opposed to also the SolverPPC which was also registering Bpred)
     - the electric field on level ghost nodes, obtained from refinement, cannot be anymore obtained with space/time inteprolation as before. This is because we need the magnetic field obtained at the end of the 4 fine substeps on fine level border faces, to be **exactly** the one obtained on the shared coarse faces during the underlying coarse step. This means that the electric field Faraday evaluates on level ghost nodes, must be the same as the one that was last used on next coarser overlaped nodes to obtain the coarse magnetic field. This means  : 
         - solverPPC now sets `Eavg` in the modelInfo `ghostElectric`  instead of `Epred`, since Eavg is the one the next coarser level uses last to update its magnetic field
         - the hybridhybrid messenger strat now adds various convenience method such as `fillStaticRefiner_` and the refiner pool `addStaticRefiner` and all necessary overloads for scalar and vector quantities.  

## bug fix
- density ghost nodes was previously filled with only spatial interpolation. This was a mistake since as a result the density on level ghost nodes was obtained from refining the next coarser values, defined in the future. This is now fixed by using the new refiner pool method `addTimeRefiner` in the `registerGhostComms_()` method. 
- the refiner pool `addTimeRefiner` overload for vector fields is now also used to register ghost communications for the bulk velocity (through calling `fillTimeRefiners_` 

## Regriding trick

During regriding, a level may shrink or grow. In the first case, some parts of the new level border end up on faces that were formerly fine domain faces. These fine faces had their magnetic field self consistently obtained from Faraday. Each face has its own value, i.e. fine level border faces enclosed in the same coarse face do not have equal values. This is inconsistent with flux conservation since next fine step we assume fine faces evolve based on the same electric induction to arrive at the same flux as the coarse level. Border fine faces thus need to be reset by refinement, which will get them equal both to the underlying coarse face's value. The drawback is that when changing the fine face value obtained self-consistently on the border, we screw-up divB in these fine border cells. Since the refined value is obtained from next coarser, which previously through coarsening recieved the averaged of both fine faces, the re-assignement does not change the total flux through the border faces, but only screw up divB of enclosed fine cells.

In the latter case, i.e. when the level grows, new level faces (on border or not) must take their value from refinement from next coarser. But fixing these faces' values will screw up divB in the fine level at the former level boundary since at this former boundary, one side has self consistent fine flux, and the other is set by refinement from coarser. 

## Misc. : 

- prepareStep() is now called in the hybrid level initializer because.... **SAY WHY**
- quantity_communicator.hpp becomes communicator.hpp
- communicator.hpp is split in two files :  refiner_pool.hpp and synchronizer_pool.hpp 
- a new overload of Communicator::makeRefiner() is created to communicated a scalar quantity with space and time refinement.

### DIFF
--- a/src/amr/solvers/solver_ppc.hpp
+++ b/src/amr/solvers/solver_ppc.hpp
@@ -116,7 +116,7 @@ private:
                     double const currentTime, double const newTime);
 
 
-    void average_(level_t& level, HybridModel& model);
+    void average_(level_t& level, HybridModel& model, Messenger& fromCoarser, double const newTime);
 
 
     void moveIons_(level_t& level, Ions& ions, Electromag& electromag, ResourcesManager& rm,
@@ -179,11 +179,12 @@ void SolverPPC<HybridModel, AMR_Types>::fillMessengerInfo(
     auto& modelInfo = dynamic_cast<amr::HybridMessengerInfo&>(*info);
 
     auto const& Epred = electromagPred_.E;
+    auto const& Eavg  = electromagAvg_.E;
     auto const& Bpred = electromagPred_.B;
 
-    modelInfo.ghostElectric.emplace_back(Epred);
-    modelInfo.ghostMagnetic.emplace_back(Bpred);
-    modelInfo.initMagnetic.emplace_back(Bpred);
+    modelInfo.ghostElectric.emplace_back(amr::VecFieldDescriptor{Eavg});
+    // modelInfo.ghostMagnetic.emplace_back(Bpred);
+    modelInfo.initMagnetic.emplace_back(amr::VecFieldDescriptor{Bpred});
 }
 
 
@@ -247,19 +248,20 @@ void SolverPPC<HybridModel, AMR_Types>::advanceLevel(std::shared_ptr<hierarchy_t
 
     predictor1_(*level, hybridModel, fromCoarser, currentTime, newTime);
 
-
-    average_(*level, hybridModel);
+    average_(*level, hybridModel, fromCoarser, newTime);
 
     saveState_(*level, hybridState.ions, resourcesManager);
+
     moveIons_(*level, hybridState.ions, electromagAvg_, resourcesManager, fromCoarser, currentTime,
               newTime, core::UpdaterMode::domain_only);
 
     predictor2_(*level, hybridModel, fromCoarser, currentTime, newTime);
 
 
-    average_(*level, hybridModel);
+    average_(*level, hybridModel, fromCoarser, newTime);
 
     restoreState_(*level, hybridState.ions, resourcesManager);
+
     moveIons_(*level, hybridState.ions, electromagAvg_, resourcesManager, fromCoarser, currentTime,
               newTime, core::UpdaterMode::all);
 
@@ -285,6 +287,7 @@ void SolverPPC<HybridModel, AMR_Types>::predictor1_(level_t& level, HybridModel&
     auto& electromag       = hybridState.electromag;
     auto levelNumber       = level.getLevelNumber();
 
+    std::cout << "start pred 1 on level " << levelNumber << "\n";
 
     {
         PHARE_LOG_SCOPE("SolverPPC::predictor1_.faraday");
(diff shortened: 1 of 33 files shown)


tests/amr/data/field/coarsening/test_linear_coarsen.hpp
```diff
@@ -143,7 +143,40 @@ struct FieldCoarsenTestData
     }
 };
 
-template<std::size_t dimension>
+template<std::size_t dimension, typename Coarsener>
```
PhilipDeegan, line 146: `std::size_t dimension` can be inferred from centering array I think

```
template<typename Coarsener>
```

and 

```
coarsen<Coarsener>(0, fineBoxX, coarseBoxX, centeringX, ratio, em);
```

tests/amr/data/field/coarsening/test_coarsen_field.py
```diff
@@ -4,15 +4,18 @@
 
 import os
 import sys
-import numpy as np
+from multiprocessing import Value
+from socket import AI_PASSIVE
```
PhilipDeegan, line 8: eh?

src/core/data/grid/gridlayout.hpp
```diff
@@ -37,7 +37,10 @@ namespace core
     constexpr bool has_physicalQuantity_v = has_physicalQuantity<T>::value;
 
 
-    constexpr int centering2int(QtyCentering c) { return static_cast<int>(c); }
+    constexpr int centering2int(QtyCentering c)
```
PhilipDeegan, line 40: `constexpr static` ? 

don't see a class instance (constexpr or not) needed here

src/amr/solvers/solver_ppc.hpp
```diff
@@ -467,10 +481,11 @@ void SolverPPC<HybridModel, AMR_Types>::corrector_(level_t& level, HybridModel&
             resourcesManager->setTime(B, *patch, newTime);
         }
 
-        fromCoarser.fillMagneticGhosts(B, levelNumber, newTime);
+        std::cout << "faraday done on level " << levelNumber << "\n";
+        // fromCoarser.fillMagneticGhosts(B, levelNumber, newTime);
```
PhilipDeegan, line 485: rm or?

src/amr/solvers/solver_ppc.hpp
```diff
@@ -467,10 +481,11 @@ void SolverPPC<HybridModel, AMR_Types>::corrector_(level_t& level, HybridModel&
...
     }
 
-
+    // TODO misses AMPERE here
```
PhilipDeegan, line 488: should this be done before merging?

nicolasaunai: probably, it's weirdly been missing since the beginning....

src/core/data/grid/gridlayout.hpp
nicolasaunai: I don't get the comment, it's just a free function why do you talk about class instances

src/amr/solvers/solver_ppc.hpp
nicolasaunai: have created issue #742 to deal with it in isolation

src/core/data/grid/gridlayout.hpp
PhilipDeegan: >  it's just a free function

yeah nvm

PR #736, opened 2023-07-14
attempt to remove excessive allocations in updater

PhilipDeegan:


src/core/data/particles/particle_array.hpp
```diff
@@ -230,6 +230,22 @@ class ParticleArray
     auto& vector() { return particles_; }
     auto& vector() const { return particles_; }
 
+    auto& update_from(This const& that)
+    {
+        bool resize_bigger = this->size() < this->size();
+        if (resize_bigger)
+            this->resize(this->size() + static_cast<std::size_t>(this->size() * .05));
```
nicolasaunai, line 237: I'm wondering about this 0.05 over-allocation coef. How does it compare to the size of 1 element (1 particle?)
Also we could write it `(1+coef)*this->size()`?

src/core/data/particles/particle_array.hpp
```diff
@@ -230,6 +230,22 @@ class ParticleArray
     auto& vector() { return particles_; }
     auto& vector() const { return particles_; }
 
+    auto& update_from(This const& that)
+    {
+        bool resize_bigger = this->size() < this->size();
```
nicolasaunai, line 235: this this is always false?

src/core/data/particles/particle_array.hpp
```diff
@@ -230,6 +230,22 @@ class ParticleArray
...
+        else
+            this->resize(that.size());
+        std::copy(that.begin(), that.end(), this->begin());
+        this->box_     = that.box_;
+        this->cellMap_ = that.cellMap_;
+        return *this;
+    }
+
+
+    void shrink_to_fit() { particles_.shrink_to_fit(); }
```
nicolasaunai, line 247: Mayby wrong reasons but always had some bad feelings about this `shrink_to_fit` function.
Had some issues in the past were it led to unexpected results, can't remember well.
found this : http://slashslash.info/2013/11/capacity-members-for-vector-shrink_to_fit/

nicolasaunai (commented):
thx for the pr, here are some preliminary non exhaustive uncomplete and quickly made comments :)

src/core/data/particles/particle_array.hpp
PhilipDeegan: yep it should be this vs that

src/core/data/particles/particle_array.hpp
PhilipDeegan: > I'm wondering about this 0.05 

me too

> How does it compare to the size of 1 element (1 particle?)

Assuming it floors it would be 1+0 (if I get what you meant)

> Also we could write it (1+coef)*this->size()?

sure

the idea here (which you probably got), is that we might get an array in another patch that has like one or two more particles, and we'd have have to reallocatate again if we didn't over allocate previously, so trying to avoid that, while also not taking too much additional memory

src/core/data/particles/particle_array.hpp
PhilipDeegan: I commented on this not always working in updater.hpp, alternatively, we can have a unique_ptr<vector> and reset/reallocate it if we want to free up some memory

src/core/data/particles/particle_array.hpp
PhilipDeegan: we might also be able to just delete this function and rely on `operator=` , tbc

src/core/numerics/ion_updater/ion_updater.hpp
```diff
@@ -27,8 +27,9 @@ template<typename Ions, typename Electromag, typename GridLayout>
 class IonUpdater
 {
 public:
-    static constexpr auto dimension    = GridLayout::dimension;
-    static constexpr auto interp_order = GridLayout::interp_order;
+    static constexpr auto dimension          = GridLayout::dimension;
+    static constexpr auto interp_order       = GridLayout::interp_order;
+    static constexpr std::uint16_t n_samples = 64; // artibrary
```
PhilipDeegan, line 32: artibrary -> arbitrary

src/core/numerics/ion_updater/ion_updater.hpp
```diff
@@ -62,10 +64,38 @@ class IonUpdater
...
+        sample_sizes_[sample_iter_] = particles.size();
+        ++sample_iter_;
+        if (sample_iter_ == n_samples)
+        {
+            sample_iter_     = 0;
+            auto const delta = static_cast<std::size_t>(tmp_particles_->size() * shrink_delta);
+            if (any(sample_sizes_, [&](auto const& s) { return s > delta; }))
+                return setter(particles, *tmp_particles_.get()); // do not shrink
+        }
+        tmp_particles_ = std::make_unique<ParticleArray>(Box{}); // reallocate to shrink
```
PhilipDeegan, line 91: might not need a unique_ptr even, the following should also work

```
tmp_particles_ = std::move(std::vector<T>(0));
```

PhilipDeegan:
might be simplest to reset the updater state during a regrid
so it is allowed to reallocate and potentially release memory

src/amr/solvers/solver_ppc.hpp
```diff
@@ -94,6 +94,7 @@ class SolverPPC : public ISolver<AMR_Types>
                               double const currentTime, double const newTime) override;
 
 
+    void regrid() override { ionUpdater_.reset(); }
```
nicolasaunai, line 97: name `regrid` may be misleading as the operation does not regrid per se. I suggest `applyOnRegrid()` maybe, or `onRegrid()`

src/amr/solvers/solver.hpp
```diff
@@ -85,6 +85,8 @@ namespace solver
 
 
 
+        virtual void regrid() {} // do what you need to do on regrid
```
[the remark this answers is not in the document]
PhilipDeegan: overriding is optional

src/core/data/particles/particle_array.hpp
```diff
@@ -233,6 +233,17 @@ class ParticleArray
 
     auto& box() const { return box_; }
 
+
+    auto& update_from(This const& that)
```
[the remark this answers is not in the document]
nicolasaunai: or `replace_from` rather, or equivalently more explicit regarding the underlying copy

nicolasaunai, line 237: I know it's not from this PR, but I'm curious regarding the need to define `This` actually, we could just use `ParticleArray` (without the template specialization) that would be clearer.

```cpp

template<typename T>
struct A
{
  auto& update_from(A const& from)
  {
      std::cout << __LINE__ << "\n";
      return *this;
  }
};
```

src/core/numerics/ion_updater/ion_updater.hpp
```diff
@@ -62,10 +62,25 @@ class IonUpdater
     void updateIons(Ions& ions, GridLayout const& layout);
 
 
+    void reset()
+    {
+        // clear memory
+        tmp_particles_ = std::move(ParticleArray{Box{}});
```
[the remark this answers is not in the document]
nicolasaunai: 🤷‍♂️ ok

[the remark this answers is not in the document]
PhilipDeegan: previously I had it as unique_ptr which avoid some of this potential confusion

nicolasaunai (approved):

PR #737, opened 2023-07-20
add __mul__ operator on hierarchies, hierarchie specialisation and sqrt on hierarchies

rochSmets:
in `pyphare/pyphare/scripts` there is a `operators.ipynb` working on a tiny harris run on kaa. It illustrates the use of these operators.
TBC...

### DIFF
--- a/pyphare/pyphare/pharesee/hierarchy.py
+++ b/pyphare/pyphare/pharesee/hierarchy.py
@@ -74,16 +74,20 @@ class FieldData(PatchData):
             )
         return self._z
 
+
     def primal_directions(self):
         return self.size - self.ghost_box.shape
+
 
     def __str__(self):
         return "FieldData: (box=({}, {}), key={})".format(
             self.layout.box, self.layout.box.shape, self.field_name
         )
 
+
     def __repr__(self):
         return self.__str__()
+
 
     def select(self, box):
         """
@@ -109,8 +113,10 @@ class FieldData(PatchData):
                 return self.dataset[lower[0] : upper[0] + 1, lower[1] : upper[1] + 1]
         return np.array([])
 
+
     def __getitem__(self, box):
         return self.select(box)
+
 
     def __init__(self, layout, field_name, data, **kwargs):
         """
@@ -128,6 +134,7 @@ class FieldData(PatchData):
         self.name = field_name
         self.dl = np.asarray(layout.dl)
         self.ndim = layout.box.ndim
+        self.box = layout.box
         self.ghosts_nbr = np.zeros(self.ndim, dtype=int)
 
         if field_name in layout.centering["X"]:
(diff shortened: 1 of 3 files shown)


pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -1061,6 +1378,109 @@ def dist_plot(self, **kwargs):
...
+            return VectorField(new_patch_level, domain_box, time=time_left)
+
+        elif isinstance(right_operand, (int, float)):
+
+            patch_levels = self.patch_levels
+            domain_box = self.domain_box
+
+            names = ['x', 'y', 'z']
+
+            for time in list(self.time_hier.keys()):
```
nicolasaunai, line 1440: there is `times()`

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -1061,6 +1378,109 @@ def dist_plot(self, **kwargs):
...
+                    new_patches = {}
+
+                    assert(all (l.id == r.id for l, r in zip(lvl_left.patches, lvl_right.patches)))
+                    for (patch_left, patch_right) in zip(lvl_left.patches, lvl_right.patches):
+                        layout = patch_left.layout
+
+                        if ilvl_left not in new_patches:
+                            new_patches[ilvl_left] = []
+
+                        new_patches[ilvl_left].append(patch_left * patch_right)
```
nicolasaunai, line 1427: So this is were you're using the `*` operator on patches. I think we should do it on patchdatas.

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -262,6 +486,99 @@ def __call__(self, qty, **kwargs):
...
+        left_op = self % operand
+        right_op = operand % self
+
+        new_pd = {}
+
+        for (name, name_left, name_right) in zip(names, left_op.patch_datas.keys(), right_op.patch_datas.keys()):
+            pd_left = left_op.patch_datas[name_left]
+            pd_right = right_op.patch_datas[name_right]
+
+            dset = np.asarray(pd_left.dataset)*np.asarray(pd_right.dataset)
```
nicolasaunai, line 572: This looks like the product component-wise. I think it's misleading as this is neither the dot/cross/outer product

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -1362,11 +1786,17 @@ def hierarchy_fromh5(h5_filename, time, hier, silent=True):
 
                     patch_levels[ilvl] = PatchLevel(ilvl, patches[ilvl])
 
-        diag_hier = PatchHierarchy(
-            patch_levels, domain_box, refinement_ratio, t, data_file
-        )
+        if num_of_components == 1:
```
nicolasaunai, line 1789: I don't think we want to hard-code that we want a Scalar or Vector or Tensor or anything in that function that should just return a `PatchHierarchy`.

Indeed it is not because the h5 has 3 datasets that they necessarily are components of a vectorfield.. could be Bx, N, P.

We should keep these high level mathematical types of `PatchHierarchy`(scalar, vectors, etc). in physical models such as `run`. 




```python
run.GetB() # should return a VectorField
run.GetNi() # should return a ScalarField
run.GetPressure() # should return a TensorField
```

pyphare/pyphare/pharesee/hierarchy.py
```diff
@@ -1362,11 +1786,17 @@ def hierarchy_fromh5(h5_filename, time, hier, silent=True):
...
+
+        elif num_of_components == 3:
+            return VectorField(patch_levels, domain_box, refinement_ratio, t, data_file)
 
-        return diag_hier
+        elif num_of_components == 6 or num_of_components == 9:
+            pass
+
+        else:
+            raise ValueError("what the fuck is this patch data ?")
```
nicolasaunai, line 1799: rm

pyphare/pyphare/scripts/operators.ipynb
```diff
@@ -0,0 +1,286 @@
+{
```
nicolasaunai, line 1: we will consider a notebook for the documentation, maybe not in this PR yet

PR #738, opened 2023-07-20
add of a load balancer for the hybrid model based on NPPC

rochSmets:
It unfortunately will fail the test... just to show how it goes

### DIFF
--- a/src/simulator/simulator.hpp
+++ b/src/simulator/simulator.hpp
@@ -2,7 +2,8 @@
 #ifndef PHARE_SIMULATOR_SIMULATOR_HPP
 #define PHARE_SIMULATOR_SIMULATOR_HPP
 
-
+#include <vector>
+#include <string>
 
 #include "phare_core.hpp"
 #include "phare_types.hpp"
@@ -12,10 +13,8 @@
 #include "core/utilities/mpi_utils.hpp"
 #include "core/utilities/timestamps.hpp"
 #include "amr/tagging/tagger_factory.hpp"
-
-#include <chrono>
-#include <exception>
-#include <unordered_set>
+#include "amr/load_balancing/load_balancer_manager.hpp"
+#include "amr/load_balancing/load_balancer_estimator_hybrid.hpp"
 
 
 namespace PHARE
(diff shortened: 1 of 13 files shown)


src/simulator/simulator.hpp
```diff
@@ -262,6 +261,15 @@ void Simulator<dim, _interp, nbRefinedPart>::hybrid_init(initializer::PHAREDict
     auto hybridTagger_ = amr::TaggerFactory<PHARETypes>::make("HybridModel", "default");
     multiphysInteg_->registerTagger(0, maxLevelNumber_ - 1, std::move(hybridTagger_));
 
+
+    auto lbm_ = std::make_unique<amr::LoadBalancerManager<dim>>(dict);
+    auto lbe_ = std::make_unique<amr::LoadBalancerEstimatorHybrid<PHARETypes>>(
```
PhilipDeegan, line 266: should probably be `make_shared` given `addLoadBalancerEstimator` takes a shared_ptr

which seems to result in ASAN issues 

```
==14494==ERROR: AddressSanitizer: new-delete-type-mismatch on 0x602000078e30 in thread T0:
  object passed to delete has wrong type:
  size of the allocated type:   16 bytes;
  size of the deallocated type: 8 bytes.
    #0 0x7f96282ba3c8 in operator delete(void*, unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:164
    #1 0x7f9624613649 in std::default_delete<PHARE::amr::LoadBalancerHybridStrategy<PHARE::PHARE_Types<1ul, 1ul, 2ul> > >::operator()(PHARE::amr::LoadBalancerHybridStrategy<PHARE::PHARE_Types<1ul, 1ul, 2ul> >*) const /usr/include/c++/12/bits/unique_ptr.h:95
    #2 0x7f9624613649 in std::unique_ptr<PHARE::amr::LoadBalancerHybridStrategy<PHARE::PHARE_Types<1ul, 1ul, 2ul> >, std::default_delete<PHARE::amr::LoadBalancerHybridStrategy<PHARE::PHARE_Types<1ul, 1ul, 2ul> > > >::~unique_ptr() /usr/include/c++/12/bits/unique_ptr.h:396
    #3 0x7f9624613649 in PHARE::amr::LoadBalancerEstimatorHybrid<PHARE::PHARE_Types<1ul, 1ul, 2ul> >::~LoadBalancerEstimatorHybrid() /path/to/phare/roch/src/amr/load_balancing/load_balancer_estimator_hybrid.hpp:38

...

```

nicolasaunai: One aspect to take care of is that load balancing level 0 is currently not supported in PHARE. At lease [because of this](https://github.com/PHAREHUB/PHARE/blob/master/src/amr/level_initializer/hybrid_level_initializer.hpp#L61-L84).

the `is_regridding` is in the `else` not root. Currently PHARE thinks initializing the root level always means that it's done at t=0 with python functions. It's calling ` model.initialize(level);` whereas non root level call `model.initLevel()`.

We should probably have a dedicated PR for testing balancing level 0 without introducing other stuff.

PR #740, opened 2023-08-31
count sort

nicolasaunai:


src/core/utilities/sorting.hpp
```diff
@@ -0,0 +1,78 @@
...
+public:
+    CountingSort(std::size_t size = 0)
+        : tmp_(size)
+    {
+    }
+
+    using value_type = typename Array::value_type;
+    using SortBox    = core::Box<int, dim>;
+
+    void setup(std::size_t nbr_elements, SortBox const& box)
```
nicolasaunai, line 25: I've put the allocations in a separate routine so it can be done elsewhere and not occur when sorting.
It's not great yet since the box given here to allocate the histogram must be the same as the one used by the cellGetter passed to sort() otherwise it's shit.

particle array given to sort() must also be consistent with what's been given as `nbr_elements` otherwise nonsense... 

TBC

tests/core/utilities/sorting/test_sorting.cpp
```diff
@@ -0,0 +1,102 @@
...
+            auto const ny = this->lastCellIndex[1] + 1;
+            auto const nz = this->lastCellIndex[2] + 1;
+            return particle.iCell[2] + particle.iCell[1] * nz + particle.iCell[0] * ny * nz;
+        }
+    };
+    sorter.sort(this->particles, cellFattener);
+
+
+    // check that the particles are sorted
+    for (std::size_t iPart = 0; iPart < this->nbrParticles; ++iPart)
```
PhilipDeegan, line 86:
```
for (std::size_t iPart = 1; iPart < this->nbrParticles; ++iPart)
    EXPECT_LE(cellFattener(this->particles[iPart - 1]), cellFattener(this->particles[iPart]));
```

?

PR #741, opened 2023-09-01
constexpr product

nicolasaunai:
This modifies the `core::product`  that is not constexpr and constexpr init of the `product(ConstArray<T,dim>(value)}` is useful

### DIFF
--- a/src/core/utilities/types.hpp
+++ b/src/core/utilities/types.hpp
@@ -249,12 +249,11 @@ constexpr Multiplies product(Container const& container, Multiplies init = 1)
     // using  :
     // return std::accumulate(container.begin(), container.end(), init,
     // std::multiplies<Multiplies>()); will not be constexpr until C++20.
-    auto result = init;
     for (auto const& element : container)
     {
-        result *= element;
+        init *= element;
     }
-    return result;
+    return init;
 }
 
 
--- a/src/core/utilities/types.hpp
+++ b/src/core/utilities/types.hpp
@@ -232,7 +232,10 @@ namespace core
             return std::string{val};
         return std::nullopt;
     }
-    inline std::optional<std::string> get_env(std::string&& key) { return get_env(key); }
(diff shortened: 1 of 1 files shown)


src/core/utilities/types.hpp
```diff
@@ -232,7 +232,10 @@ namespace core
             return std::string{val};
         return std::nullopt;
     }
-    inline std::optional<std::string> get_env(std::string&& key) { return get_env(key); }
+    inline std::optional<std::string> get_env(std::string&& key)
```
nicolasaunai, line 235: this is formatting...

src/core/utilities/types.hpp
```diff
@@ -241,11 +244,20 @@ namespace core
 namespace PHARE::core
 {
 template<typename Container, typename Multiplies = typename Container::value_type>
-Multiplies product(Container const& container, Multiplies mul = 1)
+constexpr Multiplies product(Container const& container, Multiplies init = 1)
```
PhilipDeegan, line 247: I think the point of this `Multiplies` type is to be able to override the return type

https://github.com/PhilipDeegan/PHARE/blob/mkn/src/core/utilities/types.hpp#L254

nicolasaunai: and the version I pushed breaks that?

PhilipDeegan: no, but I don't get the copy from `init` to `result`

nicolasaunai: ´Cause you’re too smart

PhilipDeegan (approved):

nicolasaunai:
the question now is do we merge that or do we keep the master code and change to cpp20 to get the constexpr accumulate? Maybe we merge since accumulate is not on GPU anyway?

nicolasaunai:
extracted from #740  and not useful for now in master..

PR #743, opened 2023-09-05
magnetic flux across levels

nicolasaunai:
## Add a new way to Coarsen a field, specialized for the magnetic field: 

- adds magnetic field coarsener
   - caveat : this is not 3D  
- previous coarsener becomes the `DefaultFieldCoarsener` and is used for the density and bulk velocity
- `FieldCoarserOperator`now takes a `FieldCoarsenerPolicy` as a template parameter

## Adding specific ways to refine the magnetic and electric fields: 

- `FieldRefineOperator` now takes a `FieldRefinerPolicy` as a template parameter, which is used in `refine()` to fill the fine filed given the coarse field and a coarse index.
- add objects `ElectricFieldRefiner` and `MagneticFieldRefiner`
- previous `FieldRefiner` now becomes `DefaultFieldRefiner` and is used by density and bulk velocity

## Conserving magnetic flux at level interfaces

- magnetic field components are defined on cell faces and its value on a face only depends on the electric components on the edges of that face. Hence knowing the electric field everywhere on the ghost box we can calculate B everywhere on the ghost box. 
    - Hence, B is now calculated over the whole **ghost** box using faraday.
    - This adds `evalOnGhostBox` in `GridLayout` which is called in `Faraday` instead of `evalOnBox`
    - Magnetic field ghosts are not filled anymore. the messenger method fillMagneticGhost is removed.
    - magnetic patch ghosts **only** are filled in `postSynchronize`. This is needed because reset of field data by coarsened values may break overlap consistency with neighbor patches. 
        - This uses the new `RefinerType::PatchGhostField` in `refiner_pool.hpp`
        - This adds a new type of schedule created for intra level only, associated with constexpr `RefinerType::PatchGhostField`
        - The hybridModel is the only one registering a magnetic field in `ghostMagnetic` (as opposed to also the SolverPPC which was also registering Bpred)
     - the electric field on level ghost nodes, obtained from refinement, cannot be anymore obtained with space/time inteprolation as before. This is because we need the magnetic field obtained at the end of the 4 fine substeps on fine level border faces, to be **exactly** the one obtained on the shared coarse faces during the underlying coarse step. This means that the electric field Faraday evaluates on level ghost nodes, must be the same as the one that was last used on next coarser overlaped nodes to obtain the coarse magnetic field. This means  : 
         - solverPPC now sets `Eavg` in the modelInfo `ghostElectric`  instead of `Epred`, since Eavg is the one the next coarser level uses last to update its magnetic field
         - the hybridhybrid messenger strat now adds various convenience method such as `fillStaticRefiner_` and the refiner pool `addStaticRefiner` and all necessary overloads for scalar and vector quantities.  

## bug fix
- density ghost nodes was previously filled with only spatial interpolation. This was a mistake since as a result the density on level ghost nodes was obtained from refining the next coarser values, defined in the future. This is now fixed by using the new refiner pool method `addTimeRefiner` in the `registerGhostComms_()` method. 
- the refiner pool `addTimeRefiner` overload for vector fields is now also used to register ghost communications for the bulk velocity (through calling `fillTimeRefiners_` 

## Regriding trick

During regriding, a level may shrink or grow. In the first case, some parts of the new level border end up on faces that were formerly fine domain faces. These fine faces had their magnetic field self consistently obtained from Faraday. Each face has its own value, i.e. fine level border faces enclosed in the same coarse face do not have equal values. This is inconsistent with flux conservation since next fine step we assume fine faces evolve based on the same electric induction to arrive at the same flux as the coarse level. Border fine faces thus need to be reset by refinement, which will get them equal both to the underlying coarse face's value. The drawback is that when changing the fine face value obtained self-consistently on the border, we screw-up divB in these fine border cells. Since the refined value is obtained from next coarser, which previously through coarsening recieved the averaged of both fine faces, the re-assignement does not change the total flux through the border faces, but only screw up divB of enclosed fine cells.

In the latter case, i.e. when the level grows, new level faces (on border or not) must take their value from refinement from next coarser. But fixing these faces' values will screw up divB in the fine level at the former level boundary since at this former boundary, one side has self consistent fine flux, and the other is set by refinement from coarser. 

## tests

 - I have disabled test_field_coarsening_via_subcycles in both 1D and 2D test_fields_advance. These tests assert that levelghotst of E and B are, as expected, the space/time interpolation of next coarser values. But the code does not refine B onto ghosts anymore, and does not time interpolate E anymore so the test is useless as is. It could be re-factored into testing the same thing for moments, which this PR now make time/space interpolated, but that is an issue in itself (#748).

## Misc. : 

- quantity_communicator.hpp becomes communicator.hpp
- communicator.hpp is split in two files :  refiner_pool.hpp and synchronizer_pool.hpp 
- a new overload of Communicator::makeRefiner() is created to communicated a scalar quantity with space and time refinement.

image

This is the divergence of the magnetic field in `run051`  Harris sheet on 3 levels and 80 cores.

### Summary by CodeRabbit

- New Feature: Introduced `evalOnGhostBox` function for more efficient field evaluations.
- Refactor: Improved grid layout and field coarsening with clearer variable names and better function signatures.
- New Feature: Added `DefaultFieldCoarsener` and `MagneticFieldCoarsener` for more precise field coarsening.
- New Feature: Introduced `ElectricFieldRefiner` and `MagneticFieldRefiner` for enhanced field refinement.
- Refactor: Streamlined level initialization and multiphysics integration for better performance.
- New Feature: Enhanced `makeRefiner` function for improved communication between scalar and vector quantities.
- New Feature: Introduced `SynchronizerPool` for efficient synchronization and management of refiners.
- Refactor: Updated `HybridModel` and `SolverPPC` for better handling of physical models and solvers.
- Test: Updated and improved readability of various test cases for better error handling and testing efficiency.

### DIFF
--- a/src/amr/data/field/coarsening/magnetic_field_coarsener.hpp
+++ b/src/amr/data/field/coarsening/magnetic_field_coarsener.hpp
@@ -1,0 +1,137 @@
+#ifndef PHARE_MAGNETIC_FIELD_COARSENER
+#define PHARE_MAGNETIC_FIELD_COARSENER
+
+#include "core/data/grid/gridlayoutdefs.hpp"
+#include "core/hybrid/hybrid_quantities.hpp"
+#include "core/utilities/constants.hpp"
+
+
+#include <SAMRAI/hier/Box.h>
+#include <stdexcept>
+
+namespace PHARE::amr
+{
+using core::dirX;
+using core::dirY;
+using core::dirZ;
+/** @brief This class gives an operator() that performs the coarsening of N fine nodes onto a
+ * given coarse node
+ *
+ * A MagneticFieldCoarsener object is created each time the refine() method of the
+ * FieldCoarsenOperator is called and its operator() is called for each coarse index.
+ * It is the default coarsening policy and used for any field that does not come with
+ * specific constraints (such as conserving some property in the coarsening process).
+ *
+ *
+ * This coarsening operation is defined so to conserve the magnetic flux.
+ * This is done by assigning to a magnetic field component on a coarse face, the average
+ * of the enclosed fine faces
+ *
+ */
+template<std::size_t dimension>
+class MagneticFieldCoarsener
+{
+public:
+    MagneticFieldCoarsener(std::array<core::QtyCentering, dimension> const centering,
+                           SAMRAI::hier::Box const& sourceBox,
+                           SAMRAI::hier::Box const& destinationBox,
+                           SAMRAI::hier::IntVector const& ratio)
+        : centering_{centering}
+        , sourceBox_{sourceBox}
+        , destinationBox_{destinationBox}
+
+    {
+    }
+
+    template<typename FieldT>
+    void operator()(FieldT const& fineField, FieldT& coarseField,
+                    core::Point<int, dimension> coarseIndex)
+    {
+        // For the moment we only take the case of field with the same centering
+        TBOX_ASSERT(fineField.physicalQuantity() == coarseField.physicalQuantity());
+
+        core::Point<int, dimension> fineStartIndex;
+
+        fineStartIndex[dirX] = coarseIndex[dirX] * this->ratio_;
+
+        if constexpr (dimension > 1)
+        {
+            fineStartIndex[dirY] = coarseIndex[dirY] * this->ratio_;
+            if constexpr (dimension > 2)
+            {
+                fineStartIndex[dirZ] = coarseIndex[dirZ] * this->ratio_;
+            }
+        }
+
+        fineStartIndex = AMRToLocal(fineStartIndex, sourceBox_);
+        coarseIndex    = AMRToLocal(coarseIndex, destinationBox_);
+
+        // the following kinda assumes where B is, i.e. Yee layout centering
+        // as it only does faces pirmal-dual, dual-primal and dual-dual
+
+        if constexpr (dimension == 1)
+        {
+            // in 1D div(B) is automatically satisfied so using this coarsening
+            // opertor is probably not better than the default one, but we do that
+            // for a kind of consistency...
+            // coarse flux is equal to fine flux and we're 1D so there is flux partitioned
+            // only for By and Bz, Bx is equal to the fine value
+
+            if (centering_[dirX] == core::QtyCentering::primal) // bx
+            {
+                coarseField(coarseIndex[dirX]) = fineField(fineStartIndex[dirX]);
+            }
+            else if (centering_[dirX] == core::QtyCentering::dual) // by and bz
+            {
+                coarseField(coarseIndex[dirX])
+                    = 0.5 * (fineField(fineStartIndex[dirX] + 1) + fineField(fineStartIndex[dirX]));
+            }
+        }
+
+        if constexpr (dimension == 2)
+        {
+            if (centering_[dirX] == core::QtyCentering::primal
+                and centering_[dirY] == core::QtyCentering::dual)
+            {
+                coarseField(coarseIndex[dirX], coarseIndex[dirY])
+                    = 0.5
+                      * (fineField(fineStartIndex[dirX], fineStartIndex[dirY])
+                         + fineField(fineStartIndex[dirX], fineStartIndex[dirY] + 1));
+            }
+            else if (centering_[dirX] == core::QtyCentering::dual
+                     and centering_[dirY] == core::QtyCentering::primal)
+            {
+                coarseField(coarseIndex[dirX], coarseIndex[dirY])
+                    = 0.5
+                      * (fineField(fineStartIndex[dirX], fineStartIndex[dirY])
+                         + fineField(fineStartIndex[dirX] + 1, fineStartIndex[dirY]));
+            }
+            else if (centering_[dirX] == core::QtyCentering::dual
+                     and centering_[dirY] == core::QtyCentering::dual)
+            {
+                coarseField(coarseIndex[dirX], coarseIndex[dirY])
+                    = 0.25
+                      * (fineField(fineStartIndex[dirX], fineStartIndex[dirY])
+                         + fineField(fineStartIndex[dirX] + 1, fineStartIndex[dirY])
+                         + fineField(fineStartIndex[dirX], fineStartIndex[dirY] + 1)
+                         + fineField(fineStartIndex[dirX] + 1, fineStartIndex[dirY] + 1));
+            }
+            else
+            {
+                throw std::runtime_error("no magnetic field should end up here");
+            }
+        }
+        else if constexpr (dimension == 3)
+        {
+            throw std::runtime_error("Not Implemented yet");
+        }
+    }
+
+private:
+    std::array<core::QtyCentering, dimension> const centering_;
+    SAMRAI::hier::Box const sourceBox_;
+    SAMRAI::hier::Box const destinationBox_;
+    static int constexpr ratio_ = 2;
+};
+} // namespace PHARE::amr
+#endif
--- a/src/amr/data/field/refine/electric_field_refiner.hpp
+++ b/src/amr/data/field/refine/electric_field_refiner.hpp
@@ -1,0 +1,321 @@
+#ifndef PHARE_ELECTRIC_FIELD_REFINER_HPP
+#define PHARE_ELECTRIC_FIELD_REFINER_HPP
+
+#include <SAMRAI/hier/Box.h>
+
+#include "amr/resources_manager/amr_utils.hpp"
+#include "core/utilities/constants.hpp"
+#include "core/data/grid/gridlayoutdefs.hpp"
+#include "core/utilities/point/point.hpp"
+
+#include <cstddef>
+
+namespace PHARE::amr
+{
+
+using core::dirX;
+using core::dirY;
+using core::dirZ;
+
+template<std::size_t dimension>
+class ElectricFieldRefiner
+{
+public:
+    ElectricFieldRefiner(std::array<core::QtyCentering, dimension> const& centering,
+                         SAMRAI::hier::Box const& destinationGhostBox,
+                         SAMRAI::hier::Box const& sourceGhostBox,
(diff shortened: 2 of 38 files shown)


src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -0,0 +1,299 @@
...
+            fineField(locFineIdx[dirX]) = coarseField(locCoarseIdx[dirX]);
+        }
+        else if constexpr (dimension == 2)
+        {
+            // ilc: index local coarse
+            // ilf: index local fine
+            auto ilcx = locCoarseIdx[dirX];
+            auto ilcy = locCoarseIdx[dirY];
+            auto ilfx = locFineIdx[dirX];
+            auto ilfy = locFineIdx[dirY];
```
PhilipDeegan, line 78: can be const?

src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -0,0 +1,299 @@
...
+
+
+            // Ey
```
PhilipDeegan, line 81: any reason why `y` is first?

src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    SAMRAI::hier::Box const fineBox_;
+    SAMRAI::hier::Box const coarseBox_;
+    std::array<core::QtyCentering, dimension> const centerings_;
+};
+} // namespace PHARE::amr
+
+
+#endif // !PHARE_ELECTRIC_FIELD_REFINER_HPP
```
PhilipDeegan, line 299: `!` ?

src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -0,0 +1,299 @@
...
+        TBOX_ASSERT(coarseField.physicalQuantity() == fineField.physicalQuantity());
+
+        auto locFineIdx = AMRToLocal(fineIndex, fineBox_);
+        auto coarseIdx{fineIndex};
+        for (auto& idx : coarseIdx)
+            idx = idx / 2;
+        auto locCoarseIdx = AMRToLocal(coarseIdx, coarseBox_);
+
+
+        if constexpr (dimension == 1)
```
PhilipDeegan, line 51: would it be feasible to have dim specific functions?

src/amr/messengers/refiner_pool.hpp
```diff
@@ -26,50 +26,64 @@ namespace amr
...
+                              bool ghosts = false)
+        {
+            addStaticRefiner(descriptor, descriptor, refineOp, key, rm, ghosts);
+        }
+
+
 
         /**
-         * @brief add a QuantityCommunicator to the Communicators based on the given descroptors and
+         * @brief  Add a Reginer
```
PhilipDeegan, line 71: `Reginer` lol

src/amr/messengers/refiner_pool.hpp
```diff
@@ -26,50 +26,64 @@ namespace amr
...
+         * add a QuantityCommunicator to the Communicators based on the given descroptors and
```
PhilipDeegan, line 72: `descroptors` 🚁

src/amr/data/field/refine/electric_field_refiner.hpp
nicolasaunai: should be

src/amr/data/field/refine/electric_field_refiner.hpp
nicolasaunai: I'm crazy

src/amr/data/field/refine/electric_field_refiner.hpp
nicolasaunai: done

src/amr/data/field/refine/electric_field_refiner.hpp
PhilipDeegan: It's in other places too so maybe it's some auto header setup thing

src/amr/messengers/mhd_hybrid_messenger_strategy.hpp
```diff
@@ -34,7 +34,8 @@ namespace amr
         }
 
         /**
-         * @brief allocate allocate the internal resources to the hybrid and MHD resourcesManagers
+         * @brief allocate allocate the internal resources to the hybrid and MHD
```
PhilipDeegan, line 37: double allocate

(I know it's formatting, but removing an "allocate" might help it use one line)

tests/amr/data/field/coarsening/test_coarsen_field.py
```diff
@@ -4,15 +4,18 @@
 
 import os
 import sys
-import numpy as np
+from multiprocessing import Value
+from socket import AI_PASSIVE
```
PhilipDeegan, line 8: 🤔

src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -0,0 +1,321 @@
...
+{
+public:
+    ElectricFieldRefiner(std::array<core::QtyCentering, dimension> const& centering,
+                         SAMRAI::hier::Box const& destinationGhostBox,
+                         SAMRAI::hier::Box const& sourceGhostBox,
+                         SAMRAI::hier::IntVector const& ratio)
+        : fineBox_{destinationGhostBox}
+        , coarseBox_{sourceGhostBox}
+        , centerings_{centering}
+    {
```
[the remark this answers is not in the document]
PhilipDeegan: `centering` is compile time sized to `dimension` - so this is nonsense.

src/amr/data/field/coarsening/magnetic_field_coarsener.hpp
```diff
@@ -0,0 +1,139 @@
...
+                coarseField(coarseIndex[dirX], coarseIndex[dirY])
+                    = 0.25
+                      * (fineField(fineStartIndex[dirX], fineStartIndex[dirY])
+                         + fineField(fineStartIndex[dirX] + 1, fineStartIndex[dirY])
+                         + fineField(fineStartIndex[dirX], fineStartIndex[dirY] + 1)
+                         + fineField(fineStartIndex[dirX] + 1, fineStartIndex[dirY] + 1));
+            }
+            else
+            {
+                {
```
PhilipDeegan, line 121: redundant braces?

src/amr/data/field/coarsening/magnetic_field_coarsener.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *
+ */
+template<std::size_t dimension>
+class MagneticFieldCoarsener
+{
+public:
+    MagneticFieldCoarsener(std::array<core::QtyCentering, dimension> const centering,
+                           SAMRAI::hier::Box const& sourceBox,
+                           SAMRAI::hier::Box const& destinationBox,
+                           SAMRAI::hier::IntVector const& ratio)
```
PhilipDeegan, line 38: are you using `SAMRAI::hier::IntVector const& ratio` cause it was taken from a code sample?
do we need it considering we only support `ratio==2` (assuming it's the same thing)

src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+
+    template<typename FieldT>
+    void refine1D_(FieldT const& coarseField, FieldT& fineField,
+                   core::Point<int, dimension> const& locFineIdx,
+                   core::Point<int, dimension> const& locCoarseIdx)
+    {
+        // if dual, then Ex
+        // if even fine index, we're on top of coarse, we take 100% coarse overlaped fieldValue
```
PhilipDeegan, line 82: minor typo `overlaped` -> `overlapped`

src/amr/data/field/refine/field_refine_operator.hpp
```diff
@@ -27,168 +27,161 @@ class AFieldRefineOperator
...
-         *
-         */
-        void refine(SAMRAI::hier::Patch& destination, SAMRAI::hier::Patch const& source,
-                    int const destinationId, int const sourceId,
-                    SAMRAI::hier::BoxOverlap const& destinationOverlap,
-                    SAMRAI::hier::IntVector const& ratio) const override
-        {
-            using FieldGeometry = typename FieldDataT::Geometry;
+    /**
+     * @brief Given a  set of box on a  fine patch, compute the interpolation from
```
PhilipDeegan, line 71: multiple double whitespace

tests/simulator/__init__.py
```diff
@@ -105,7 +116,7 @@ def defaultPopulationSettings(sim, density_fn, vbulk_fn):
 
 def makeBasicModel(extra_pops={}):
     sim = ph.global_vars.sim
-    _density_fn_periodic = globals()["density_"+str(sim.ndim)+"d_periodic"]
+    _density_fn_periodic = globals()["density_" + str(sim.ndim) + "d_periodic"]
```
PhilipDeegan, line 119: minor - could be `f"density_{sim.ndim}d_periodic"`

PhilipDeegan (approved):

src/amr/data/field/coarsening/magnetic_field_coarsener.hpp
nicolasaunai: I'm taking the IntVector because that's what the `coarsen()` function gives it, that function comes from SAMRAI and I didn't want to convert it for no good reason. You may wonder why this function takes this argument even though `ratio_` is constexpr set to 2 and that argument is not even used. But that class is used as a template param in `coarsen` called `FieldCoarsenerPolicy` and the other "policy" , namely `DefaultFieldCoarsener` uses this ratio....

tests/simulator/__init__.py
PhilipDeegan: I thought so, but could still be better ;)

PR #744, opened 2023-09-16
upgrade to C++20

PhilipDeegan:
## Summary by CodeRabbit

- **New Features**
  - Introduced an option `PHARE_COMPILER_WORKAROUNDS` to control compiler false positives.

- **Enhancements**
  - Updated the C++ standard to C++20.
  - Improved type deduction and function invocation with `std::invoke_result_t`.
  - Enhanced `Point`, `Box`, and `GridLayout` classes with refined constructors and comparison operators.

- **Bug Fixes**
  - Addressed GNU compiler flags support with new conditional flags.

- **Refactor**
  - Removed copy and move constructors, and assignment operators in `ParticlesPack`.
  - Updated various templates and type mechanisms for better code consistency.

- **Tests**
  - Refactored test structures for better readability and maintainability.
  - Added a new `dump` method in diagnostics tests for better control over diagnostic data management.


nicolasaunai:
would it work with gcc13?

nicolasaunai:
> would it work with gcc13?

https://gist.github.com/nicolasaunai/eb5bc3540f9be6b46b335ca1d8882a45
warning on both gcc12 and gcc13

nicolasaunai:
only seen in -O2/-O3, no warning in O0 or O1.

PhilipDeegan:
I think the issue is the wrong generate function is being used, it should be [the array](https://github.com/PHAREHUB/PHARE/blob/master/src/core/utilities/types.hpp#L314) version I think

src/core/utilities/types.hpp
```diff
@@ -232,7 +232,10 @@ namespace core
             return std::string{val};
         return std::nullopt;
     }
-    inline std::optional<std::string> get_env(std::string&& key) { return get_env(key); }
+    inline std::optional<std::string> get_env(std::string&& key)
+    {
+        return get_env(key);
+    }
```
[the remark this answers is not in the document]
PhilipDeegan: pretty sure this suggestion would cause infinite recursion, not fix it

src/amr/data/particles/particles_data.hpp
```diff
@@ -178,7 +178,7 @@ namespace amr
             using Packer = core::ParticlePacker<dim>;
 
             auto getParticles = [&](std::string const name, auto& particles) {
-                auto const keys_exist = core::generate(
+                std::array<bool, Packer::n_keys> keys_exist = core::generate(
                     [&](auto const& key) { return restart_db->keyExists(name + "_" + key); },
                     Packer::keys());
```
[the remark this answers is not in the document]
PhilipDeegan: :clown_face:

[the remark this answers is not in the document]
PhilipDeegan: `keys_exist ` was previously a vector, so this assertion is not necessary anymore, thanks

src/core/utilities/types.hpp
```diff
@@ -355,13 +355,41 @@ NO_DISCARD auto any(Container const& container, Fn fn = to_bool)
...
     return std::none_of(container.begin(), container.end(), fn);
 }
 
+
+template<typename SignedInt, typename UnsignedInt>
+bool diff_sign_int_equals(SignedInt const& i0, UnsignedInt const& i1)
+{
+    static_assert(std::is_unsigned_v<UnsignedInt>);
+    static_assert(std::is_signed_v<SignedInt>);
+    static_assert(sizeof(UnsignedInt) >= sizeof(SignedInt), "Bad int comparison!");
```
PhilipDeegan, line 370: it's not exactly "bad"

It's just more complicated if we need to start checking if one int is bigger than the max size of the other

simpler/safer for now to just error if byte sizes are too different

src/core/utilities/types.hpp
```diff
@@ -355,13 +355,41 @@ NO_DISCARD auto any(Container const& container, Fn fn = to_bool)
...
+template<typename Int0, typename Int1>
+bool int_equals(Int0 const& i0, Int1 const& i1)
+{
+    if constexpr (std::is_same_v<Int0, Int1>)
+        return i0 == i1;
+    else
+    {
+        if constexpr (std::is_unsigned_v<Int0> and std::is_signed_v<Int1>)
+            return diff_sign_int_equals(i1, i0);
+        if constexpr (std::is_unsigned_v<Int1> and std::is_signed_v<Int0>)
```
nicolasaunai, line 386: else if constexpr ?

res/cmake/def.cmake
```diff
@@ -2,11 +2,24 @@
 
 # Per compiler CXXFLAGS
 set (PHARE_FLAGS ${PHARE_FLAGS} )
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-  set (PHARE_FLAGS ${PHARE_FLAGS} )
-else() # !Clang
-  set (PHARE_FLAGS ${PHARE_FLAGS} --param=min-pagesize=0 )
-endif() # clang
+
+if(PHARE_COMPILER_WORKAROUNDS)
```
nicolasaunai, line 6: this seems complex... how stable is that, do you expect we won't need that soon or it'll have to stay

nicolasaunai (approved):
do we make the crossing to cpp20?

res/cmake/def.cmake
PhilipDeegan: I expect we need the compile flags just cause it is failing without them

this mechanism is just to allow it to be switched off if undesirable for some reason

src/core/utilities/types.hpp
PhilipDeegan: pretty sure it's working as if it's already like that, but I can make it more explicit sure

PhilipDeegan:
`ParticlePack` struct init is failing in cpp20, I've resolved this on my gpu branch so to fix here soon

.github/workflows/cmake_macos.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         - os: macos-13
           dylib_path: /usr/local/lib/gcc/current
         - os: macos-latest
-          dylib_path: /opt/homebrew/Cellar/gcc/13.2.0/lib/gcc/current
+          dylib_path: /opt/homebrew/Cellar/gcc/14.1.0/lib/gcc/current
```
PhilipDeegan, line 24: https://github.com/orgs/community/discussions/124141

PhilipDeegan: this isn't related to this PR, but the container GCC has changed so it fails without a change

nicolasaunai (approved):

Issue #752, opened 2023-10-13
compute pressure tensor for diagnostics


nicolasaunai:
- make that available in phare in
- should be done at runtime in cpp
- only 6 components for the 3x3 tensor should be stored

PR #762, opened 2023-10-24
pressure deposit

nicolasaunai:
#752 

Computing the pressure is like using the ParticleToMesh except we need to deposit neither 1 nor the velocity, but the velocity squared. The idea thus is to generalize the ParticleToMesh operator so, per particle, to deposit only 1 field at a time, given the field, the particle and a functor that, given the particle would:
- for the density, return 1
- for the flux, return vx, or vy or vz
- for the pressure, return vx*vx,  vx*vy etc.

This is done so that one day if needed to compute something else, like the heat flux (v^3) we can simply add more functors.


nicolasaunai:
test @coderabbitai review

Issue #749, opened 2023-10-02
Caliper undefined reference


nicolasaunai:
` cmake -DwithCaliper` 

leads to `undefined reference to symbol _ZN3MPI3Win4FreeEv`
when  building with SAMRAI pre-installed.
why ? and why not with SAMRAI as subproject?

PhilipDeegan:
could be BLT related
https://github.com/LLNL/SAMRAI/issues/222

PhilipDeegan:
This is blocking the TC caliper nightly build, to look at more in depth

PhilipDeegan:
ah, it appears this goes away with 


```
cmake -DCMAKE_CXX_FLAGS="-DOMPI_SKIP_MPICXX=1" ..
```

PhilipDeegan:
short term, add the define, long term, have something for for dealing with different mpi impls?

PR #776, opened 2023-11-13
system interrogation for mpi/etc

PhilipDeegan:
This should
1. resolve https://github.com/PHAREHUB/PHARE/issues/749
2. give us a method to check the system before building to see what's available as a sort of "use everything available" cmake setup

This requires updating caliper build on TC to verify, but I've checked locally and it's fine

It can possibly be run from our regular cmake too, and if it was default on, the caliper builds wouldn't need changing

## Summary by CodeRabbit

- **New Features**
  - Introduced a new search functionality with a search bar and results display.
  - Added a new option to enable Position Independent Code during build configuration.
  - Implemented a new build configuration option to enable Interprocedural Optimization by default.

- **Enhancements**
  - Improved the inclusion of MPI headers across various files to centralize and streamline MPI-related configurations.
  - Updated the build scripts to facilitate easier configuration and building of the project with new options.

- **Bug Fixes**
  - Fixed an issue with the handling of MPI headers and warning suppression by introducing a new header file that manages MPI inclusion and compiler warnings more effectively.

- **Documentation**
  - Updated `.gitignore` to exclude newly generated files from version control.

- **Refactor**
  - Refactored the MPI header inclusion to use a new centralized header, enhancing maintainability and consistency across the codebase.


src/core/def/mpi.hpp
```diff
@@ -0,0 +1,16 @@
+
```
PhilipDeegan, line 1: could be `phare_mpi.hpp` to really try to avoid any filename conflicts

todo, add header guards

src/amr/data/field/coarsening/default_field_coarsener.hpp
```diff
@@ -1,6 +1,9 @@
 #ifndef PHARE_DEFAULT_FIELD_COARSENER_HPP
 #define PHARE_DEFAULT_FIELD_COARSENER_HPP
 
+
+#include "core/def/mpi.hpp"
```
PhilipDeegan, line 5: it's either this or we modify the CXXFLAGS, which I think is not my preferred approach

basically anywhere samrai is imported we need to configure somethings before it, includes it's own config header so we don't get things defined there from 
image

alternatively it could be a phare samrai def hjeader, which internally calls the mpi def header

PhilipDeegan: This can be seen compiling/linking with caliper [here](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Phil_Build/79071?buildTab=log&focusLine=0&logView=flowAware)

src/core/def/mpi.hpp
nicolasaunai: yeah ok

tests/amr/tagging/test_tagging.cpp
```diff
@@ -1,4 +1,6 @@
 
+#define OMPI_SKIP_MPICXX 1
```
nicolasaunai, line 2: remove?

nicolasaunai (commented):
Ok I added few comments we mentioned this morning.
Could you also add some comment in the tools/config/ directory, maybe some readme or something, that just briefly explain the intent behind these files.

nicolasaunai (approved):

tools/config/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+cmake_minimum_required (VERSION 3.22)
```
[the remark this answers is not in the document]
PhilipDeegan: likely

PR #786, opened 2023-11-26
force elapsed time test to save restart after first advance

PhilipDeegan:
passing on my fork https://github.com/PhilipDeegan/PHARE/pull/79


tests/simulator/config.py
```diff
@@ -1 +1,7 @@
 project_root = "@PHARE_PROJECT_DIR@"
+
+at = "@" # try to work without cmake config
+
+if project_root == f"{at}PHARE_PROJECT_DIR{at}":
```
PhilipDeegan, line 5: not related, but I keep making this change locally on every branch just to run tests from root and not the build dir

tests/simulator/test_restarts.py
```diff
@@ -295,22 +292,23 @@ def test_restarts_elapsed_time(self, ndim, interp, simInput, expected_num_levels
...
         local_out = self.unique_diag_dir_for_test_case(
             f"{out}/elapsed_test", ndim, interp
         )
-        simput["restart_options"]["dir"] = local_out
-        import datetime
+        diag_dir0 = local_out
+        diag_dir1 = f"{local_out}_n2"
 
-        seconds = 9
+        seconds=1 # dump on first advance always!
```
nicolasaunai, line 306: how are we sure what the comment says is true?

PhilipDeegan: with `autodump=false` I think it's correct

nicolasaunai (approved):

Issue #739, opened 2023-08-11
Log everything relevant for a run


nicolasaunai:
Each run should automatically log as many informations as possible to later retrieve the context in which it was done.
Here we list things that could be relevant:

Where should that data live?
- a possibility is that it's exported in every h5 file the run generates
- if no h5 is written (crashed before the first dump, or no diag required) we still need that somewhere

## code information
- [ ] git commit of PHARE that has been used to run. Warning, it should be the commit used by the build which is ran, not the commit of the code in the source dir at the time the run is performed.
- [ ] keep the message of that commit?

## runtime and build context
- machine on which the run was performed and any relevant information about that machine
    - warning: keep track of machine on which the code was built and the one on which is ran? (front end VS nodes) 
    - [ ] # of cores available + # of core used
    - [ ] memory available
    - [ ] system and version
- [ ] python version
- [ ] version of python lib we use
- [ ] c++ libs versions
- [ ] MPI version
- [ ] Cmake version
- [ ] HDF5 version
- [ ] all dependencies versions
- [ ] build date and time
- [ ] build path
- [ ] source path
- [ ] run date and time

## run intentions
- [ ] allow users to enter a kind of Doc String for the performed simulation, as parameter of the Simulation() for instance.

## Reader
- [ ] display all above information in a nicely structured document (Markdown?)

PhilipDeegan:
for `code information`

we can add the git commit to the build info, which can then be used in the runtime info, and diagnostics

for runtime information, the mechanism [is here](https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/cpp/validate.py#L87)

which I think tries to account for the three checkboxes

```
 # of cores available + # of core used
 memory available
 system and version
```

the exported build time info is [here](https://github.com/PHAREHUB/PHARE/blob/master/tools/config/config.py#L20-L30)

which includes

```
 python version
 MPI version
 Cmake version
 HDF5 version
```

what's remaining for build time

```
version of python lib we use # not sure how this could differ really, and doesn't always exist at build time?
c++ libs versions # these things are mostly in diagnostics
all dependencies versions # this is more release related (define all)
build date and time
build path
source path
```

what's remaining for run time

```
 run date and time # kinda get this from the file timestamp
```

for `run intentions`, a `description` field was added to the python simulation class, which is entirely serialized into restart files and diagnostics

PhilipDeegan:
I think the `phare_configurator` cmake option resolves this

[...]

PR #788, opened 2023-12-02
system reporting

PhilipDeegan:
for discussion for #739


res/cmake/def.cmake
```diff
@@ -10,6 +10,7 @@ if (phare_configurator)
   execute_process(
     COMMAND ./tools/config/cmake.sh
     WORKING_DIRECTORY ${PHARE_PROJECT_DIR}
+    COMMAND_ERROR_IS_FATAL ANY
```
PhilipDeegan, line 13: this supposedly needed cmake 3.19 so hence the version upgrade
2.5 years since release I think is reasonable, for cmake 3.20.1

nicolasaunai (approved):

Issue #802, opened 2023-12-21
add a Patch Id in the patch hierarchy creation


rochSmets:
Since this commit :
https://github.com/PHAREHUB/PHARE/commit/37a4d1750faffe8f2e7a7efe7860e56e4db0d9bf
in the ctor of a PatchHierarchy, the hierarchy contains a single-quantity hierarchy which can be accessed through a dict (see https://github.com/PHAREHUB/PHARE/blob/618c8f66651a2c4ad90207335dbb66e54e678dbc/pyphare/pyphare/pharesee/hierarchy.py#L588-L606)
But when the Patches are created, they miss a `patch.id`.

PR #804, opened 2024-01-05
add global patch id to patch diag attributes

PhilipDeegan:
for #802

h5dump shows 
image


PhilipDeegan:
the patch id is in the path, as an attribute is redundant

PR #809, opened 2024-02-08
nd restart tests

PhilipDeegan:
Currently the python restarts tests are only 1d
This PR allows for 2d and eventually 3d tests

### DIFF
--- a/tests/simulator/__init__.py
+++ b/tests/simulator/__init__.py
@@ -1,8 +1,11 @@
 import os
 import unittest
+
+from copy import deepcopy
+
+
 from datetime import datetime
 import pyphare.pharein as ph, numpy as np
-from pyphare.pharein import ElectronModel
 
 
 def parse_cli_args(pop_from_sys=True):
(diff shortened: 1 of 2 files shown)


tests/simulator/test_restarts.py
```diff
@@ -205,7 +219,16 @@ def count_levels_and_patches(qty):
...
                         else:
-                            np.testing.assert_equal(pd0.dataset[:], pd1.dataset[:])
+                            if cpp.mpi_size() == 1:
+                                np.testing.assert_equal(pd0.dataset[:], pd1.dataset[:])
+                            else:
+                                # rank order for shared nodes not guaranteed
+                                np.testing.assert_allclose(
+                                    pd0.dataset[:],
+                                    pd1.dataset[:],
+                                    atol=1e-14,  # 1e-15 fails empirically
```
PhilipDeegan, line 229: is flakey, could be 1e-12/13?

PhilipDeegan: [still flakey, fails sometimes as is](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_BuildGithubPrClang/81379?buildTab=log&linesState=4939&logView=flowAware&focusLine=6784)

res/cmake/def.cmake
```diff
@@ -11,6 +11,10 @@ endif() # clang
 set (PHARE_LINK_FLAGS )
 set (PHARE_BASE_LIBS )
 
+if(NOT DEFINED PHARE_MPIRUN_POSTFIX)
```
PhilipDeegan, line 14: without this, running multiple instances of mpirun -n 2 python3 ... effectively only lets run one process, the rest are descheduled until the first is finished

we could not default it here but set it manually in TC builds, or have it as an option defaulted to ON

PhilipDeegan: this was moved to the configurator, so to rm

tests/simulator/__init__.py
```diff
@@ -260,3 +265,11 @@
...
                 if os.path.exists(diag_dir):
                     shutil.rmtree(diag_dir)
+
+    def hierarchy_by_box(self, hier):
+        # comparing hierarchies by index isn't always right
+        box_hier = {i: {} for i in range(len(hier.levels()))}
+        for ilvl, patches in box_hier.items():
+            for patch in hier.level(ilvl).patches:
+                patches[str(patch.box)] = patch
+        return box_hier
```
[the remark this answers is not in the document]
PhilipDeegan: this is used as part of a test. so it is tested.

Issue #785, opened 2023-11-26
account for possible different charges of ion populations


nicolasaunai:
We can have multiple populations, each with its charge.
But the density used in Ohm and in Electrons is the total ion *particle*density, which is wrong if populations have different charges.
We should allow for calculating the total charge density $\rho_c = \sum_{pop} Z_{pop}n_{pop}$

We should *change* `rho_` to be the charge density, since the total particle density will not be used anyway.

PhilipDeegan:
I think #861 closes this

PR #817, opened 2024-03-26
account for ion charges

alexandre-de-larminat:
could this do the job for issue #785? I didn't check if density was used anywhere to actually mean Ni.

## Summary by CodeRabbit

- **New Features**
	- Introduced the concept of charge to ion populations, enhancing the simulation's accuracy by enabling charge-based calculations and representations.
- **Enhancements**
	- Improved ion population calculations by incorporating the charge and mass of ions, allowing for more precise transformations and analyses.
- **Tests**
	- Added tests to verify the correct implementation of the new charge attribute in ion populations, ensuring reliability and accuracy.

### DIFF
--- a/src/core/data/ions/ion_population/ion_population.hpp
+++ b/src/core/data/ions/ion_population/ion_population.hpp
@@ -33,6 +33,7 @@ namespace core
         IonPopulation(initializer::PHAREDict const& initializer)
             : name_{initializer["name"].template to<std::string>()}
             , mass_{initializer["mass"].template to<double>()}
+            , charge_{initializer["particle_initializer"]["charge"].template to<double>()}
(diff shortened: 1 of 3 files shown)


alexandre-de-larminat:
welp, that was a massive fail

nicolasaunai (commented):
Thanks for the PR.
I think adding the charge density is not very complicated but a bit more tricky than what is proposed.
The charge already is a parameter per `Particle` and not per Population. Reason is that later when implemented, charge exchange is a particle process so we could have particles per populations with different charge states.
Also, this PR changes the massDensity, which should not be affected by charge.
chargeDensity(), however, is needed so to be used, say, in Ohm's law or in Electrons. But number density is still needed, to get the bulk velocity.

nicolasaunai:
closing as the right one will probably not have much code in common

PR #879, opened 2024-08-09
tagging thresholds from python

nicolasaunai:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new refinement method, "tagging," enhancing simulation configurability with a customizable tagging threshold.
	- Updated the tagging strategy to allow runtime configuration of threshold values.
	- Enhanced tagger creation with a dictionary-based approach for improved flexibility in specifying models and methods.

- **Bug Fixes**
	- Improved test coverage by adding scenarios for both valid and invalid tagger creation, ensuring better error handling.

- **Documentation**
	- Updated documentation for the Simulation class to include the new tagging refinement method and its parameters.

### DIFF
--- a/src/amr/tagging/default_hybrid_tagger_strategy.hpp
+++ b/src/amr/tagging/default_hybrid_tagger_strategy.hpp
@@ -19,9 +19,8 @@ class DefaultHybridTaggerStrategy : public HybridTaggerStrategy<HybridModel>
 
 public:
     DefaultHybridTaggerStrategy(initializer::PHAREDict const& dict)
-        :threshold_{[&](){return (dict.contains("threshold")) ? dict["threshold"].template to<double>() : 0.1;}()}
+        : threshold_{cppdict::get_value(dict, "threshold", 0.1)}
     {
-
     }
     void tag(HybridModel& model, gridlayout_type const& layout, int* tags) const override;
 
--- a/src/amr/tagging/default_hybrid_tagger_strategy.hpp
+++ b/src/amr/tagging/default_hybrid_tagger_strategy.hpp
@@ -6,8 +6,7 @@
 #include "core/data/vecfield/vecfield_component.hpp"
 #include "core/data/ndarray/ndarray_vector.hpp"
 #include <cstddef>
-#include <iostream>
-
+#include "initializer/data_provider.hpp"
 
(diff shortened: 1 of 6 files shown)


src/simulator/simulator.hpp
```diff
@@ -269,8 +269,16 @@ void Simulator<dim, _interp, nbRefinedPart>::hybrid_init(initializer::PHAREDict
     multiphysInteg_->registerAndSetupMessengers(messengerFactory_);
 
     // hard coded for now, should get some params later from the dict
-    auto hybridTagger_ = amr::TaggerFactory<PHARETypes>::make("HybridModel", "default");
-    multiphysInteg_->registerTagger(0, maxLevelNumber_ - 1, std::move(hybridTagger_));
+    if (dict["simulation"]["AMR"]["refinement"].contains("tagging"))
```
nicolasaunai, line 272: I found it weird that we create the Tagger even though we may not be tagging.
And with this new PR it would not work anyway (because the Tagger now takes the threshold from the dict it's now given, and there would not be any threshold in that dict if tagging is not specified by the user)

src/amr/tagging/default_hybrid_tagger_strategy.hpp
```diff
@@ -17,8 +16,17 @@ class DefaultHybridTaggerStrategy : public HybridTaggerStrategy<HybridModel>
     using gridlayout_type           = typename HybridModel::gridlayout_type;
     static auto constexpr dimension = HybridModel::dimension;
 
+
 public:
+    DefaultHybridTaggerStrategy(initializer::PHAREDict const& dict)
+        :threshold_{[&](){return (dict.contains("threshold")) ? dict["threshold"].template to<double>() : 0.1;}()}
```
PhilipDeegan, line 22: you don't want to use `cppdict::get_value(dict, "threshold", 0.1),`  ?

also the lambda is redundant `dict.contains("threshold") ? dict["threshold"].template to<double>() : 0.1;` is the same

PhilipDeegan (approved):

PhilipDeegan:
you need to rebase master @nicolasaunai

PhilipDeegan:
the macos issue fix was merged already, I don't think you need to bother rebasing for that

PR #894, opened 2024-09-19
nondeterministic?

PhilipDeegan:
## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted the behavior of the data unpacking process, which may enhance performance in parallel processing scenarios. 
	- Removed the enforcement of deterministic unpack ordering in the scheduling process.

- **New Features**
	- Improved flexibility in data exchange across shared borders, potentially accommodating a wider range of operational contexts. 
	- Simplified the implementation of the `getPatternName` method for better code readability.


PhilipDeegan:
we see that with this we change the mpi wait calls from "wait all" to "wait some", the some I guess being ranks of neighbouring patches rather than *all* ranks

nicolasaunai (approved):

PR #897, opened 2024-10-02
keep particle overallocation on restore state

PhilipDeegan:
keep initial capacity during save/restore state in solver ppc

## Summary by CodeRabbit

- **New Features**
	- Introduced a method for efficiently adding particle data to maps, enhancing the management of particle states.
	- Improved handling of particle ranges during updates, optimizing transfers between ghost and domain arrays.

- **Bug Fixes**
	- Refined logic for determining particles entering the domain from the ghost layer, improving overall efficiency.

- **Documentation**
	- Updated comments for better clarity on operations related to domain and ghost particles.

### DIFF
--- a/src/amr/solvers/solver_ppc.hpp
+++ b/src/amr/solvers/solver_ppc.hpp
@@ -152,6 +152,22 @@ private:
     std::unordered_map<std::string, ParticleArray> tmpDomain;
     std::unordered_map<std::string, ParticleArray> patchGhost;
 
+    template<typename Map>
+    static void add_to(Map& map, std::string const& key, ParticleArray const& ps)
+    {
+        // vector copy drops the capacity (over allocation of the source)
+        // we want to keep the overallocation somewhat - how much to be assessed
+        ParticleArray empty{ps.box()};
+
+        if (!map.count(key))
+            map.emplace(key, empty);
+        else
+            map.at(key) = empty;
+
+        auto& v = map.at(key);
+        v.reserve(ps.capacity());
+        v.replace_from(ps);
+    }
 
 }; // end solverPPC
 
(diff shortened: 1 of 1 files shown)


src/amr/solvers/solver_ppc.hpp
```diff
@@ -152,6 +152,22 @@ class SolverPPC : public ISolver<AMR_Types>
...
+    static void add_to(Map& map, std::string const& key, ParticleArray const& ps)
+    {
+        // vector copy drops the capacity (over allocation of the source)
+        // we want to keep the overallocation somewhat - how much to be assessed
+        ParticleArray empty{ps.box()};
+
+        if (!map.count(key))
+            map.emplace(key, empty);
+        else
+            map.at(key) = empty;
```
nicolasaunai, line 165: is that not to have dangling keys?

src/core/numerics/ion_updater/ion_updater.hpp
```diff
@@ -259,9 +259,13 @@ void IonUpdater<Ions, Electromag, GridLayout>::updateAndDepositAll_(Ions& ions,
                                                    interpolator_, layout, inGhostBox, inGhostLayer);
 
             auto& particleArray = particleRange.array();
-            particleArray.export_particles(
-                domainParticles, [&](auto const& cell) { return isIn(Point{cell}, domainBox); });
-
+            {
+                auto enteredInDomain
```
nicolasaunai, line 263: I am not sure this is what you think it is.
Given the name `enteredInDomain` it appears you mean it is the range or ghost particles that have entered the domain after their move.
However, and if I understand correctly, this is the range start:pivot from the inDomainBox() partition that is performed on the ghost range `start:InGhostLayer.iend()` but `inGhostLayer` is the range of ghost particles that have stayed in the ghost **layer** (not box), i.e. not in domain, not outside ghost box.

see : 

```cpp 
            return isIn(Point{cell}, ghostBox) and !isIn(Point{cell}, domainBox);
```

therefore I don't think that any particle in that range can actually be found in domain, resulting in `enteredInDomain` range always being empty and no copy happens.

am I right?

nicolasaunai: btw, if I am right, the fact that it passes the tests is concerning, in particular for such an annoying-to-debug part of the code, we may be smart to have a better test of that to prevent from merging crap leading to nasty bugs a year from now :)

PhilipDeegan: well, I think it's strange that we find any particles *in domain* from such a range, as it seems

PhilipDeegan:
fillIonghosts still under 2% since reverting to export particles
 
image

nicolasaunai (approved):

PR #911, opened 2024-10-20
safer nu

nicolasaunai:
replaces 

$$
-\nu_0 \frac{B}{n} L\nabla^2\mathbf{j}
$$

which becomes too small when $B\to 0$, typically around reconnection X points.

Capture d’écran 2024-10-19 à 16 06 24

There, increasing $\nu_0$ is not solving the problem, it only increases dissipation in regions where B is non-zero and cannot prevent the dissipation to go to zero with B.

The proposed formula is:

$$
-\nu_0 (\frac{B}{n+n_\epsilon}+1) L\nabla^2\mathbf{j}
$$

which tends to:

with

$$
-\nu_0  L\nabla^2\mathbf{j}
$$

when $B\to 0$

and 

$$
-\nu_0 (\frac{B}{n_\epsilon}+1) L\nabla^2\mathbf{j}
$$

when $n\to 0$

For reference, here is what we get un run055 with constant hyper-resistivity $\nu=0.002$:

image

### DIFF
--- a/src/core/numerics/ohm/ohm.hpp
+++ b/src/core/numerics/ohm/ohm.hpp
@@ -360,7 +360,8 @@ private:
     auto spatial_hyperresistive_(VecField const& J, VecField const& B, Field const& n,
                                  MeshIndex<VecField::dimension> index) const
     { // TODO : https://github.com/PHAREHUB/PHARE/issues/3
-        auto const lvlCoeff = 1. / std::pow(4, layout_->levelNumber());
+        auto const lvlCoeff        = 1. / std::pow(4, layout_->levelNumber());
+        auto constexpr min_density = 0.1;
 
         auto computeHR = [&](auto BxProj, auto ByProj, auto BzProj, auto nProj) {
             auto const BxOnE = GridLayout::project(B(Component::X), index, BxProj);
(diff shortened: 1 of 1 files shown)


nicolasaunai:
Ok so here is a result for this new PR.
All panels show the electric field $E_y$ in a symmetric reconnection simulation.
In all simulations $\nu_0 = 0.002$

The upper panel uses constant hyper-resistivity

$$
-\nu_0 \nabla^\mathbf{j}
$$

We see that it is too dissipative. Typically if the term is non-negligible on L0, dissipation will be too large on L2 (because the grid scale decreases relative to the dissipative scale imposed by $\nu_0$.

Middle panel uses `spatial` hyper-mode:


$$
-\nu_0\frac{B}{n} \nabla^\mathbf{j}
$$

from master.
In this case, we see the dissipation is basically non-existent around the X line and the electric field gets super strong.
The reason is that in the X point region, $B\to 0$ so the hyper-resistive term becomes negligible no matter how large $\nu_0$ is.


Bottom panel uses the proposed formula

$$
-\nu_0 \left(\frac{B}{n + n_\epsilon} + 1\right)\nabla^2\mathbf{j}
$$

with $\n_\epsilon=0.1$

In contrast to the master version, dissipation clearly operates around the X point despite $\nu_0$ being the same and $B\to 0$, showing the effect of the new formula. We also see the simulation is less dissipative overall than the `constant` hyper mode, showing the benefit of having the level-dependent coefficient.

hyperres

UCaromel (approved):

PR #921, opened 2024-11-03
introduce charge, mass and particle densities

rochSmets:
The charge and mass of each ion population is now treated correctly.


nicolasaunai:
On build fails.
I think what should be done is not testing function VS data on a point-wise basis for such noisy quantities but rather computing the standard deviation between function and data over the whole dataset and put a (empirical) threshold on that. This is what we do for density initialization tests and it's quite less sensitive to noise.

nicolasaunai (commented):
https://github.com/rochSmets/PHARE/pull/2

tests/simulator/initialize/test_density_init.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+    def assert_close_enough(h, H):
+        for lvl_h, lvl_H in zip(h.levels(time).values(), H.levels(time).values()):
+            for patch_h, patch_H in zip(lvl_h.patches, lvl_H.patches):
+                pd_h = patch_h.patch_datas["value"]
+                pd_H = patch_H.patch_datas["value"]
+                ghosts_num = pd_h.ghosts_nbr[0]
+
+                if pd_H.ndim == 1:
```
PhilipDeegan, line 266: I think you can replace all this with

```
                dset_h = pd_h[pd_h.box]
                dset_H = pd_H[pd_H.box]
```

pyphare/pyphare/pharesee/hierarchy/fromfunc.py
```diff
@@ -0,0 +1,40 @@
+from pyphare.pharesee.hierarchy.hierarchy_utils import compute_hier_from
+
+import numpy as np
```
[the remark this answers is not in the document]
rochSmets: fixed

pyphare/pyphare/pharesee/hierarchy/fromfunc.py
```diff
@@ -0,0 +1,40 @@
...
+
+    return compute_hier_from(compute_, hier, **kwargs)
+
+
+def hierarchy_from_func(func, hier, **kwargs):
+    if hier.ndim == 1:
+        return hierarchy_from_func1d(func, hier, **kwargs)
+    if hier.ndim == 2:
+        return hierarchy_from_func2d(func, hier, **kwargs)
+
```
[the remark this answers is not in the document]
rochSmets: fixed

pyphare/pyphare/pharesee/hierarchy/__init__.py
```diff
@@ -39,4 +40,7 @@ def hierarchy_from(
     if simulator is not None and qty is not None:
         return hierarchy_from_sim(simulator, qty, pop=pop)
 
+    if func is not None and hier is not None:
+        return hierarchy_from_func(func, hier, **kwargs)
+
```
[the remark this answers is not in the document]
rochSmets: fixed

src/amr/tagging/default_hybrid_tagger_strategy.hpp
```diff
@@ -36,7 +36,7 @@ void DefaultHybridTaggerStrategy<HybridModel>::tag(HybridModel& model,
     auto& By = model.state.electromag.B.getComponent(PHARE::core::Component::Y);
     auto& Bz = model.state.electromag.B.getComponent(PHARE::core::Component::Z);
 
-    auto& N = model.state.ions.density();
+    auto& N = model.state.ions.chargeDensity();
```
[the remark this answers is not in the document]
rochSmets: ok

tests/core/numerics/ion_updater/test_updater.cpp
```diff
@@ -909,7 +924,7 @@ TYPED_TEST(IonUpdaterTest, thatNoNaNsExistOnPhysicalNodesMoments)
     {
         for (auto ix = ix0; ix <= ix1; ++ix)
         {
-            auto& density = pop.density();
+            auto& density = pop.particleDensity();
             auto& flux    = pop.flux();
 
             auto& fx = flux.getComponent(Component::X);
```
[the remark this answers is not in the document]
rochSmets: fixed

pyphare/pyphare_tests/test_pharesee/test_geometry_2d.py
```diff
@@ -177,11 +177,11 @@ def test_large_patchoverlaps(self, expected):
             collections=[
                 {
                     "boxes": overlap_boxes,
-                    "facecolor": "yellow",
+                    "value": 1,
```
PhilipDeegan, line 180: is there some reason for this I don't see?

tests/core/numerics/interpolator/test_main.cpp
```diff
@@ -632,7 +634,7 @@ class ACollectionOfParticles_1d : public ::testing::Test
             part.v[2]     = +1.;
             particles.push_back(part);
         }
-        interpolator(makeIndexRange(particles), rho, v, layout);
+        interpolator(makeIndexRange(particles), rho, rho_c, v, layout);
```
PhilipDeegan, line 637: is there some check that could be added to the tests `DepositCorrectlyTheirWeight_*d` 

[eg](https://github.com/PHAREHUB/PHARE/blob/8473002ed2d85ffb7a1fa8b7efc27711fb51436b/tests/core/numerics/interpolator/test_main.cpp#L654-L657)

nicolasaunai: probably... @rochSmets ?

tests/simulator/test_advance.py
```diff
@@ -289,10 +289,28 @@ def base_test_overlaped_fields_are_equal(self, datahier, coarsest_time):
                         # seems correct considering ghosts are filled with schedules
                         # involving linear/spatial interpolations and so on where
                         # rounding errors may occur.... setting atol to 5.5e-15
-                        assert_fp_any_all_close(slice1, slice2, atol=5.5e-15, rtol=0)
+                        assert_fp_any_all_close(slice1, slice2, atol=2.5e-14, rtol=0)
```
nicolasaunai, line 292: do you remember why this was changed

nicolasaunai: the proposed value is not consistent with the comment that has been changed accordingly. I think it's nice if we can keep track of why we raise these bars by providing the build that failed for previous threshold

pyphare/pyphare/pharesee/hierarchy/hierarchy.py
```diff
@@ -431,7 +445,8 @@ def plot1d(self, **kwargs):
                     qty = pdata_names[0]
 
                 layout = patch.patch_datas[qty].layout
-                nbrGhosts = layout.nbrGhostFor(qty)
+                # nbrGhosts = layout.nbrGhostFor(qty) # bad !!!
```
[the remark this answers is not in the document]
nicolasaunai: maybe because `nbrGhostFor(qty)` requires "qty" to be in layout.hybridQuantities which it may not, any remembering @rochSmets ?

nicolasaunai:
@rochSmets I made a  [PR on your repo](https://github.com/rochSmets/PHARE/pull/4) to take most  @PhilipDeegan 's comments into account. I have not added a test in test_interpolator as he suggested I let you see if you want/can do that I think it is not a bad idea (more unit tests is probably better than not enough). I let you see the comments otherwise.
In my Pr I also removed the plot_2d_patch stuff that got here for no reason, reverting the function plot_2d_patch to current master version.

src/core/data/ions/ions.hpp
```diff
@@ -131,28 +120,26 @@ namespace core
...
 
 
-            std::transform(std::begin(vx), std::end(vx), std::begin(density), std::begin(vx),
+            std::transform(std::begin(vx), std::end(vx), std::begin(massDensity_), std::begin(vx),
                            std::divides<Float>{});
-            std::transform(std::begin(vy), std::end(vy), std::begin(density), std::begin(vy),
+            std::transform(std::begin(vy), std::end(vy), std::begin(massDensity_), std::begin(vy),
                            std::divides<Float>{});
-            std::transform(std::begin(vz), std::end(vz), std::begin(density), std::begin(vz),
+            std::transform(std::begin(vz), std::end(vz), std::begin(massDensity_), std::begin(vz),
```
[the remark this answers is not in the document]
rochSmets: the problem is clear and important, the fix not marvelous... this velocity is per population, so on a given cell, the density can be strictly zero...

pyphare/pyphare/pharesee/hierarchy/fromfunc.py
```diff
@@ -0,0 +1,63 @@
...
+        return (
+            {"name": "value", "data": func(x_, **kwargs), "centering": patch_datas[ref_name].centerings},
+        )
+
+    return compute_hier_from(compute_, hier, **kwargs)
+
+
+def hierarchy_from_func2d(func, hier, **kwargs):
+    from pyphare.pharesee.hierarchy.hierarchy_utils import compute_hier_from
```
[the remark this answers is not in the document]
rochSmets: :+1:

pyphare/pyphare/pharesee/hierarchy/fromfunc.py
```diff
@@ -0,0 +1,63 @@
...
+    def compute_(patch_datas, **kwargs):
+        ref_name = next(iter(patch_datas.keys()))
+        x_ = patch_datas[ref_name].x
+        y_ = patch_datas[ref_name].y
+
+        return (
+            {"name": "value", "data": func(x_, y_, **kwargs), "centering": patch_datas[ref_name].centerings},
+        )
+
+    return compute_hier_from(compute_, hier, **kwargs)
```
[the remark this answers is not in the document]
rochSmets: :+1:

tests/core/numerics/interpolator/test_main.cpp
rochSmets: very nice suggest... I then add a charge for each part and a test on `rho_c`, the charge density. the same test for the mass could appear in testing the `ion_populations`...later !

tests/simulator/initialize/test_density_init.py
```diff
@@ -0,0 +1,332 @@
...
+    H1.plot(ax=ax4, vmin=3.75, vmax=5, cmap=cmap, title="expected mass density : 2d")
+    h2.plot(ax=ax5, vmin=2.2, vmax=3, cmap=cmap, title="computed charge density : 2d")
+    H2.plot(ax=ax6, vmin=2.2, vmax=3, cmap=cmap, title="expected charge density : 2d")
+
+    plt.tight_layout()
+    plt.savefig("nCheck.pdf", dpi=300)
+
+
+if __name__ == "__main__":
+    main()
```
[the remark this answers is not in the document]
rochSmets: yep, Nico add the lines,

pyphare/pyphare/pharesee/hierarchy/hierarchy.py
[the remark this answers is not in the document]
rochSmets: nop ?!

tests/simulator/test_advance.py
rochSmets: my bad... I dont remember why it went so,

.gdbinit
```diff
@@ -0,0 +1,3 @@
+break __assert_fail
```
nicolasaunai, line 1: I don't think we should have this file as part of the project do we?

tests/core/numerics/interpolator/test_main.cpp
```diff
@@ -541,6 +544,7 @@ class ACollectionOfParticles_1d : public ::testing::Test
             part.iCell[0] = 20; // AMR index
             part.delta[0] = 0.5;
             part.weight   = 0.4;
+            part.charge   = 1.85;
```
nicolasaunai, line 547: While we (you) still remember, I think we should take 5minutes to add a little explanation for these numbers. Because I know that in a year we will not remember what they mean and why we chose these specific values.

nicolasaunai (commented):
2 little things before it can go I think

tests/simulator/test_advance.py
nicolasaunai: so it means that it fails with 5.5e-15?

nicolasaunai:
ok this PR fails because the comparison between the hierarchy loaded from a function and that from diagnostics (from particles) don't agree even at the first decimal. We have 4.99 vs expected 5.15 on the failing build, at 1D, supposedly 2D would be even worse (if the exception would allow it to run).

I think it is "working" but the check is not reliable.

We have several options:

The first one would be to increase the number of particles per cell from 2000 to 20000, that would lower the noise by a factor of 3 and very probably almost never fail.

In the initialization test `_test_density_is_as_provided_by_user`  we don't compare the two hierarchies on a per cell basis. Rather, we compute the stddev per patch between expected and actual, and impose a threshold on that standard deviation. I think it's easier not to fail on such test, although it's strictly speaking not exactly the same...


Let's discuss that.


There are also comments not yet taken into account on this pr.

.gdbinit
PhilipDeegan: possible could be added to `.gitignore`

PR #953, opened 2025-02-18
minimized deps for cpp_etc / etc

PhilipDeegan:
related to #958

Change cpp_etc target to no longer link to libphare_simulator, which was unnecessary originally as we only need samrai and hdf5 for version info

potential todos: do no build an other libraries for phare except the necessary shared libs

i.e. dictator, initializer, cpp pybind libs

the issue, so we think is that during link time optimization there is some symbol which is expected but is not found, but the symbol, should never be used, so it's understandable that it's not found, just not why it is expected.

It seems like the issue is in diagnostics for cppdict and highfive attributes

and some other general maintenance around pybind and include/pragmas

### DIFF
--- a/src/initializer/python_data_provider.hpp
+++ b/src/initializer/python_data_provider.hpp
@@ -6,18 +6,8 @@
 
 #include "initializer/data_provider.hpp"
 
-#include "core/def/pragma_disable.hpp"
-
-// clang-format off
-DISABLE_WARNING(shadow, shadow-field-in-constructor-modified, 42)
-
-#undef HAVE_SYS_TIMES_H // included in python again, possibly with different value
-#undef HAVE_UNISTD_H
-
 #include <pybind11/embed.h> // everything needed for embedding
 #include <pybind11/functional.h>
-ENABLE_WARNING(shadow, shadow-field-in-constructor-modified, 42)
-// clang-format on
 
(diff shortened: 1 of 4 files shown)


PhilipDeegan:
```Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down```

nicolasaunai:
I think we should split the update of the CI from the other edits related to the pybind link errors? The latter are still under investigation and/or I'm not sure how the modifs here address the issue. I don't understand everything but the link error seems related to symbols in our archives dependencies being not visible for the shared pb lib that nevertheless need them.

nicolasaunai:
see #956

nicolasaunai:
> this doesn't have the linking issue anymore on fc41

I don't understand why?

PhilipDeegan:
> > this doesn't have the linking issue anymore on fc41
> 
> I don't understand why?

PR description comment updated

nicolasaunai (approved):

PR #962, opened 2025-02-21
Remove part test

nicolasaunai:
These files are not used it seems. There is no cpp test using them.
Since we do python tests on particles on refined levels I think this can be removed.


PhilipDeegan:
seems to have some overlap with #954

nicolasaunai:
> seems to have some overlap with #954

that's right, is on top and forgot to rebase apparently

PhilipDeegan (approved):

PR #965, opened 2025-02-21
remove diagnostic timestamp test overlaping another

nicolasaunai:
Maybe it was useful for something... but I can't tell what just by looking at it.
It's going further in time that the second test but it's testing more or less the same thing.
If really needed we can probably make the second test run further too (?) but the reason needs to be explained I think.


nicolasaunai:
This test was made to run for a long time to make sure we do not have rounding errors in the way the time is set in PHARE. Before, time was incrementally increase by dt, leading to rounding errors, leading to key errors while reading diags. Now with the ConstantTimeStamper, we iterate the time INDEX and get the current time increment **from start time**, which removes the rounding error.

Issue #958, opened 2025-02-20
link error on fedora 41 in release mode


nicolasaunai:
The library `cpp_etc` does not link on fedora 41 in release mode.
see [error](https://gist.github.com/nicolasaunai/d2f4ef14dda872fcb963f748629cfa0e)
Some symbols are not found.

This is possibly the result of semantic interposition rules, usage of pre-compile archives as dependencies of the pybind lib etc.
To be understood.

One idea to test is to stop making our pybind libs depend on pre-compiled libraries but rather depend from another target' sources.

nicolasaunai:
Just to continue digging a bit : how could that related to semantic interposition @UCaromel ? 
https://maskray.me/blog/2021-05-09-fno-semantic-interposition
https://www.youtube.com/watch?si=m8phT36P00fovFxc&t=65&v=vtz8S10hGuc&feature=youtu.be

PhilipDeegan:
my guess is it's [this](https://github.com/PHAREHUB/PHARE/blob/107a6a7de54d2f761ace8ddddad9c843d64491bd/src/diagnostic/diagnostics.hpp#L51) that needs visiblity=hidden

UCaromel:
I dug the subject a little bit more yesterday and finally understood what i was missing. Basically, the fact that GCC uses semantic interposition means that symbol resolution for shared libraries is postponed (after compile time). The fact that clang doesn't have semantic interposition means it assumes locality of the shared libraries' symbols. This means that the compiler will prefer often an inlining instead of a symbol lookup.

If that is the case i expect that we could expect the error to disappear by setting compiler and linker flags to -fno-semantic-interposition on a GCC build. 
Last time I tried, i didn't manage to solve the issue but i only had the compiler flag.

UCaromel:
I tried with:
```
cmake -DCMAKE_CXX_FLAGS=-fno-semantic-interposition -DCMAKE_SHARED_LINKER_FLAGS=-fno-semantic-interposition ..
```
and it didn't seem to solve the issue.

```
[100%] Linking CXX shared module ../../pybindlibs/cpp_etc.cpython-313-x86_64-linux-gnu.so
cd /home/caromel/Documents/test/PHARE/build/src/python3 && /usr/bin/cmake -E cmake_link_script CMakeFiles/cpp_etc.dir/link.txt --verbose=1
[...]

PR #969, opened 2025-03-06
make simulator an interface - no binary target

PhilipDeegan:
It's not necessary for the simulator component to be a library

All the binary was used for is the "makeAtRuntime" stuff which is only used for phare-exe

so I moved it there, which is a compiled binary already

another potential solution for #958 - ([appears it is](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Phil_Build/85477))


src/simulator/CMakeLists.txt
```diff
@@ -3,16 +3,15 @@ project(phare_simulator)
...
-target_link_libraries(${PROJECT_NAME} PUBLIC
+add_library(${PROJECT_NAME} INTERFACE)
+
+target_link_libraries(${PROJECT_NAME} INTERFACE
   phare_initializer
   phare_amr # for mpicc
   phare_diagnostic
 )
+
 set_property(TARGET ${PROJECT_NAME} APPEND_STRING PROPERTY LINK_FLAGS " ${PHARE_LINK_FLAGS}")
```
nicolasaunai, line 17: TORM

nicolasaunai (approved):

PR #970, opened 2025-03-06
particle array impl isolation

### LINKED ISSUE #967: ParticleArray copy construct is inconsistent
Currently on master, if you copy the particle array class, the cellmap of the new particle array object, will continue to reference the particles that *were* copied, not the new memory.

It seems we only use the copy construct/assignment in tests where the cellmap is not used.

Solution

Custom (non-defaulted) copy construct/assignemnt functions, where we rebuild the cellmap for the new instance.

PhilipDeegan:
closes #967

currently the internal impl of the particle array somewhat leaks out as we have to use the copy with back_inserter in order to use the ParticleArray::push_back function, which in turns updates the cell map internally.

solution is to rely on a custom copy construct which itself keeps the new cellmap consistent

### DIFF
--- a/src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
+++ b/src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
@@ -481,8 +481,9 @@ namespace amr
                     auto& levelGhostParticlesNew = pop.levelGhostParticlesNew();
                     auto& levelGhostParticles    = pop.levelGhostParticles();
 
-                    levelGhostParticlesOld = std::move(levelGhostParticlesNew);
-                    levelGhostParticles    = levelGhostParticlesOld;
+                    std::swap(levelGhostParticlesNew, levelGhostParticlesOld);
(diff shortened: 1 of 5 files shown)


PhilipDeegan:
the current state of this PR [segfaults](https://gist.githubusercontent.com/PhilipDeegan/d2c63b46486bbf5734bb3edb4ae01642/raw/f0cfef52fe4c5c6f862bd0b1584e6d2be287dd91/seger)

so, the conclusion is either, the to the messenger strategy is bad, or if not, then what?

nicolasaunai (commented):
`test_particle_array_operators` and  `test_move_swap_etc` are not very informative about what the test really does IMO. The test is about checking particlearray remains consistent after being swapped (and copied?)

`test_particle_array_consistency.cpp` and `test_is_consistent_after_swap_copy` ?

PhilipDeegan:
I forgot to update the name, todo

nicolasaunai (approved):
gtm

PR #972, opened 2025-03-06
suggestive constexpr tuple validNbrParticles

PhilipDeegan:
getting the tuple/set of possible refined particle count per dim/interp for @UCaromel mostly


tests/core/utilities/meta/test_meta_utilities.cpp
```diff
@@ -0,0 +1,30 @@
+#include <string>
```
PhilipDeegan, line 1: torm (copy paste)

PR #979, opened 2025-03-13
MHD/Hybrid Simulator merging

UCaromel:
A few things about this merging:
- the instantiation of all the possible parameters for MHD (in file **cpp_mhd_parameters.hpp**) takes a lot of compilation and link time.
- the current implementation can be improve iterating over the enums for selection directly instead of a tuple of them.
- there might be some hidden cost in these loops i am not able to explain (for example having a return condition in the loops instead of the declaring function themselves seems a lot slower).

Room for improvements i can think of:
- Resistivity and hyper-resistivity can be de-templated easily and likely with little cost on runtime performance.
- Hall is trickier but likely doable.
- We can probably limit by default a single time integrator designated to each spacial scheme (reconstruction).


src/amr/physical_models/mhd_model.hpp
```diff
@@ -44,12 +45,7 @@ class MHDModel : public IPhysicalModel<AMR_Types>
         resourcesManager->allocate(state, patch, allocateTime);
     }
 
-
-
-    virtual void
-    fillMessengerInfo(std::unique_ptr<amr::IMessengerInfo> const& /*info*/) const override
-    {
-    }
+    virtual void fillMessengerInfo(std::unique_ptr<amr::IMessengerInfo> const& info) const override;
```
PhilipDeegan, line 48: if there are to be no further subclasses, you don't need `virtual` - just `override`

UCaromel:
replaced by #1070

PR #980, opened 2025-03-14
let pybind find python

PhilipDeegan:
seems to be a needed as a consequence of https://github.com/pybind/pybind11/pull/5553/files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,6 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 include(CheckCXXCompilerFlag)
 find_program(Git git)
-find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)
 
 set(PHARE_PROJECT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -19,17 +19,16 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
 find_program(Git git)
-find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)
 
 set(PHARE_PROJECT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
 include_directories(${PHARE_PROJECT_DIR}/src)
 
 include("${PHARE_PROJECT_DIR}/res/cmake/options.cmake")
+include("${PHARE_PROJECT_DIR}/res/cmake/dep.cmake")
```
nicolasaunai, line 29: you changed the location, is that because it matters and if yes how and do we need to add a comment so it's not changed and leads to some obscure error?

PhilipDeegan: 1. this was a test
2. the [pybind pr](https://github.com/pybind/pybind11/pull/5561) should make this unnecessary
3. how would it lead to an error?
4.

PhilipDeegan: it would actually lead to some strangeness, `PHARE_EXEC_LEVEL_MIN` is used setting up gtest

let's hope the pybind pr solves the issue

PR #986, opened 2025-03-27
initial variant support for resource manager

PhilipDeegan:


PhilipDeegan:
this is more fleshed out on [my branch](https://github.com/PhilipDeegan/PHARE/commit/04a9c1dbccc0534d541f77498d0fa5748c3cfcf2) for rochs [close branch](https://github.com/rochSmets/PHARE/blob/close)

PR #987, opened 2025-03-28
drop unused field filters

PhilipDeegan:
we sometimes see a spurious werror in the field filter code

similar to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104165

but these filters are not used anymore since removing nans from moments (I think it was that anyway)


nicolasaunai (approved):
gtm

PR #994, opened 2025-04-10
compute overlap refac

nicolasaunai:
This PR suggests a refac of the computation of overlaps in python hierarchies.
The strategy is less efficient but simpler to understand and maintain. We test the overlap of each patchdata with all others in all their possible periodic shift versions.
probably needed for 3D


PhilipDeegan:
obsolete

PR #1006, opened 2025-05-27
Refine patch strategy

UCaromel:
This PR implements the divergence free refinement formulas from Toth and Roe (2002). It uses samrai's RefinePatchStrategy, which allows us to add preprocessing and post-processing functionalities on creation of a schedule. We are able to use the post-processing to code these formulas because the values of the new fine faces (that don't share their location with old coarse faces) only need the information from the previously refined faces that shared their location with the old coarse faces.

Note:
- This needs further documentation, that we might want to add to the readthedoc.
- This is a first version, which will most likely need some refactoring


pyphare/pyphare/core/gridlayout.py
```diff
@@ -229,7 +229,8 @@ def particleGhostNbr(self, interp_order):
 
     def nbrGhosts(self, interpOrder, centering):
         if self.field_ghosts_nbr == -1:
-            return int((interpOrder + 1) / 2) + self.particleGhostNbr(interpOrder)
+            nGhosts = int((interpOrder + 1) / 2) + self.particleGhostNbr(interpOrder)
```
nicolasaunai, line 232: I think this function would deserve a very well written docstring that explains why this choice of number of ghosts because this starts to be a bit cryptic...

src/amr/data/field/field_geometry.hpp
```diff
@@ -201,6 +202,20 @@ namespace amr
             return box;
         }
 
+        static SAMRAI::hier::BoxContainer toFieldBoxes(SAMRAI::hier::BoxContainer const& boxes,
```
PhilipDeegan, line 205: does this make sense after your recent PR @nicolasaunai ?

src/amr/data/field/refine/magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,425 @@
...
+    static constexpr std::size_t dimension = FieldDataT::dimension;
+
+    MagneticRefinePatchStrategy(ResMan& resourcesManager)
+        : rm_{resourcesManager}
+    {
+    }
+
+    void registerIDs(std::optional<int> bx_id, std::optional<int> by_id, std::optional<int> bz_id)
+    {
+        if (bx_id)
```
PhilipDeegan, line 34: not sure why they're optionals, is this resource manager related?

src/amr/data/field/refine/magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,425 @@
...
+            bx_id_ = bx_id;
+        if (by_id)
+            by_id_ = by_id;
+        if (bz_id)
+            bz_id_ = bz_id;
+    }
+
+    void setPhysicalBoundaryConditions(SAMRAI::hier::Patch& patch, double const fill_time,
```
PhilipDeegan, line 42: virtual function overrides should be marked as `override`

src/core/data/grid/gridlayout.hpp
```diff
@@ -1323,7 +1323,14 @@ namespace core
...
         {
-            return (interp_order + 1) / 2 + nbrParticleGhosts();
+            if constexpr (((interp_order + 1) / 2 + nbrParticleGhosts()) % 2 != 0)
+            {
+                return (interp_order + 1) / 2 + nbrParticleGhosts() + 1;
+            }
+            else
+            {
+                return (interp_order + 1) / 2 + nbrParticleGhosts();
+            }
```
PhilipDeegan, line 1333: could we not consider something which is a bit more obvious at a glance?

for instance

```
static_assert(interp_order > 0 and interp_order);
constexpr static  ghosts = std::array{2,2,4};
return ghosts[interp_order - 1]; 
```

src/amr/data/field/refine/magnetic_refine_patch_strategy.hpp
nicolasaunai: I feel this is a left-over from the second method when we used the oldLevel and sometimes didn't need it because refining not from regrid?

It seems we need these variables because once in postProcess SAMRAI only gives us the coarse and fine patches but we are left to know with which variable concretely to work.

But I find this way of doing this a bit too low level.
As we have discussed, I think, I would rather take VecFields in the MagneticRefinePatchStrategy, they are views anyway, and since we have the RM, we can set them on the coarse and fine patches as needed.

PhilipDeegan: it does, but if you can be sure that the resource should always exist in a particular context you can dreference it to drop the optional part

see https://github.com/PhilipDeegan/PHARE/blob/patchghostless/src/amr/resources_manager/resources_manager.hpp#L323-L331

src/core/data/grid/gridlayout.hpp
nicolasaunai: yes !
but also we need to write a decent comment to explain those magic numbers.

tests/core/numerics/interpolator/test_main.cpp
```diff
@@ -198,7 +198,7 @@ TYPED_TEST(AWeighter, computesPrimalBSplineWeightsForAnyParticlePosition)
     using GridLayout_t   = typename AWeighter_t::GridLayout_t;
 
     static_assert(Interpolator_t::interp_order == GridLayout_t::interp_order);
-    assert(GridLayout_t::nbrGhosts() == Interpolator_t::interp_order + 1);
+    // assert(GridLayout_t::nbrGhosts() == Interpolator_t::interp_order + 1);
```
nicolasaunai, line 201: should we remove or change rather than commeting out?

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
```diff
@@ -733,228 +772,178 @@ namespace amr
...
-        {
-            auto& Bx       = B(core::Component::X);
-            auto& By       = B(core::Component::Y);
-            auto& Bz       = B(core::Component::Z);
-            auto const& dx = layout.meshSize()[0];
-            auto const& dy = layout.meshSize()[1];
 
-            if (loc == 3) // w hi, y hi
+
+        void magneticRegriding_(std::shared_ptr<SAMRAI::hier::PatchHierarchy> const& hierarchy,
```
nicolasaunai, line 777: I think at this end of this function we should run a function that checks that nowhere there is a NaN on any component of the magnetic field left.
This function could be ran only in certain debug mode we have..

src/amr/data/field/field_geometry.hpp
```diff
@@ -201,6 +202,9 @@ namespace amr
             return box;
         }
 
+
+
+
```
PhilipDeegan, line 207: this file probably doesn't need any changes

PhilipDeegan (approved):

PR #1012, opened 2025-06-05
documentation update

nicolasaunai:
Just add markdown file for test and add some explanation about the refinement strategy for the magnetic field.

### DIFF
--- a/doc/source/theory/amr.md
+++ b/doc/source/theory/amr.md
@@ -111,7 +111,7 @@ in the overlaped region, that is assumed of better quality.
 
 ### Hybrid to Hybrid coarsening
 
-When coarsening data between two hybrid PIC levels, PHAR three types of fields:
+When coarsening data between two hybrid PIC levels, PHARE three types of fields:
 
 - the electric field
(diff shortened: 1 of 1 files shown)


Haricharan0310:
Hi @Username, I noticed you’ve been keeping docs up-to-date, always good to see.

I built a small LLM-based tool for maintaining high-level docs (SDK guides, API references, READMEs, tutorials). It auto-updates them on PRs to prevent drift.

No pressure at all, but thought it might be useful: https://github.com/marketplace/deepdocsai

doc/source/dev/dev
PhilipDeegan, line 1: is this important?

PhilipDeegan:
seems to be wip

doc/source/dev/dev
nicolasaunai: still thinking about the best way to have some documentations for developpers. Stuff not useful from a user perspective but important for us, such as ghost filling, level/patch ghost particle handling, etc. 
This was created because of what's in the index.rst below. 
Ideally, this would be from a bridge between doxygen and sphinx.

Issue #1021, opened 2025-06-16
core average faster?


nicolasaunai:
core::average loops over the array twice, first to add, second to half. could it be done in one pass with a lambda and thus faster?

PhilipDeegan:
closed by #1023

PR #1023, opened 2025-06-16
one loop average #1021

PhilipDeegan:


PhilipDeegan:
some proof https://coliru.stacked-crooked.com/a/ea09c3cc5ac0d04a

nicolasaunai (approved):
I've resolved the conflict on GH that was std::size_t instead of size_t

PR #1024, opened 2025-06-16
benchmark maintenence

PhilipDeegan:


tools/bench/amr/data/particles/copy_data.cpp
```diff
@@ -25,7 +25,6 @@ class ParticleDataCopy : public benchmark::Fixture
         data = std::make_unique<_DATA_>();
 
         assert(sourceBox == data->sourceDomain);
-        assert(sourceBox == (PHARE::amr::Box<int, dim>{data->sourceDomain}));
```
PhilipDeegan, line 28: apparently ambiguous in C++20, and kinda unnecessary so rm-ed

nicolasaunai (approved):

Issue #1005, opened 2025-05-21
magnetic refluxing


nicolasaunai:
At the moment, Loring substepped equal inductions guarantees that level border fine faces enclosed in the same coarse face all have equal values, and reach the same value as the shared coarse cell.

This means that coarsening those fine faces onto the coarse face does not change the coarse face, and does not screw up divB onto the coarse cell adjacent to the level border.

However, Loring's algo needs the refinement of the magnetic field on fine faces shared with coarse ones to be zero-th order, otherwise each fine face will have its own value and substepped equal induction will fail reaching the shared coarse face value.

If we drop the zeroth order (which #794 might lead to) we will thus break Loring's method.

This issue is about dropping Loring's method, by implementing the magnetic refluxing after synchronization is done.

Details needs to be written below but roughly speaking this implies:

- keeping the magnetic field on level border faces for each substep
- possibly keeping the electric field on level border edges for each substep?
- after coarsening, update coarse cell fluxes using substep values.

PR #1028, opened 2025-06-18
divergence free refluxing

UCaromel:
This PR's goal is to implement a refluxing strategy for the magnetic field, i order to guaranty divergence free conservation of the magnetic field with arbitrary order refinement. Currently, PHARE implements the Loring (2008) method to conserve the magnetic flux at the coarse fine boundary, but this assume 0th order refinement for the magnetic field.

Our new refluxing strategy is:
- accumulate E_average_2 on the fine level for all substeps in a dedicated buffer.
- coarsen the obtained fluxSum onto the coarse level
- retake the corrector's faraday with newly coarsened flux sum

Note: The coarsening of the fluxSum and the following faraday on the part of the coarse level overlapped by the fine must, must give the exact same result as what we would have obtained through coarsening of the magnetic field. This is not automatic, and in fact the default field coarsener in PHARE did not  permit that. This PR also implements a specific coarsener for that purpose.

implements #1005


src/amr/data/field/coarsening/electric_field_coarsener.hpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    template<typename FieldT>
+    void operator()(FieldT const& fineField, FieldT& coarseField,
+                    core::Point<int, dimension> coarseIndex)
+    {
+        // For the moment we only take the case of field with the same centering
+        TBOX_ASSERT(fineField.physicalQuantity() == coarseField.physicalQuantity());
+
+        core::Point<int, dimension> fineStartIndex;
```
PhilipDeegan, line 53: `
core::Point<int, dimension> fineStartIndex{generate([&](auto const& coarse){
 return coarsecoarse * this->ratio_;}, *coarseIndex)}
`

maybe simpler

UCaromel:
This last commit adds time refinement for E, which would break the previous method we used for fine coarse consistency at interface (Loring and Raeder 2008). This is what we obtain for divB with time refinement for E on top of master as of 09/07/2025.
loring_timeE
This is what we obtain with the refluxing method implemented in this PR.
refluxing_timeE

src/amr/data/field/coarsening/electric_field_coarsener.hpp
```diff
@@ -55,78 +53,92 @@ class MagneticFieldCoarsener
 
         core::Point<int, dimension> fineStartIndex;
```
PhilipDeegan, line 54: if you initialised this on construction it could be const ;)

PR #1062, opened 2025-07-26
fine_boundary_represents_var for copy **before** refinement

nicolasaunai:
See https://github.com/LLNL/SAMRAI/issues/292#issuecomment-3119663614 
- [x] test copy indeed occurs **before** refinement when this variable is set to false.

This PR changes the order of copy and refine operations during schedules. Using fine boundary represent variable boolean set to false in the Variable.

With this PR copies are done *before* refinement. This is tricky and leaves issues for the refinement operator not to overwrite now previously done copy. These issues are addressed in this PR by setting nans.

Consequently, the PR updates the `RefineOperator` so that it assigns values only to the nodes that have NaN values.

This thus sets the NaN on construction of `NdArrayVector`and on the `PatchData` ghost box before ghost schedules apply.

Setting NaNs on construction breaks some tests which are fixed so that fields and vecfields they use are properly initialized to 0.

We also see in this PR that in 1D Jx is not used by Ampere (rightfully) and thus used by Ohm (in JxB) with its "cosntruction-initialization value". Now being constructed-init to NaN, J is explicitly set to 0 on init (model init, fine level init and regrid init).

The PR also adds nice plots for debugging when advance test field overlap equal fails.

image

Blue are overlapped PatchData ghost boxes, red box is the overlap box (pyphare one), red markers indicate lower left corner of the cell that fails

- overwrite_interior now also default behavior for refinement with FieldVariableFillPattern

**Note**: This PR should probably not being merged as is since master currently have multiple components B and E registered to the same RefineAlgorithm and thus SAMRAI uses the first geometry, which now leaves NaNs (rather than wrong 0s) on the mesh. This should be fixed with TensorFieldData.

    - PatchData NaN initialized on construction
    - fix tests failing as result of above
    - comment a field refinement test (useless, wrong refinement op for E,B)
    - debug plots for advance field overlap test
    - copy done before refinement (boolean false in variable)
    - overwrite_interior false also for refinement is default for
      FieldFillPattern
    - J manually init to zero in model init, fine init and regrid init (Jx
      unused in ampere but used in Ohm with its now NaN values)
    - Grid/NdarrayVector take default value overrides (for test)
    - UsableTensorField is default constructed with zero init.

PR #1067, opened 2025-07-30
WIP copy before refinement and TensorFieldData

UCaromel:
This branch is on top of #1061 and #1062.

This adds the TensorFieldOverlap and modifies the TensorFieldData and Geometry to work with it. The goal is to have a SAMRAI algorithm for an entire TensorField without relying on several RegisterRefine with different fillpaterns per geometry, which we saw was not a solution as SAMRAI remove our fill patern in its internal schedules (eg copy to a temporary level) thus putting each component in a single equivalent class and using the same geometry for all components. 
With this PR the overlaps for all components are computed together and used for copy and refinement.

What's WIP:
- debug the failing tests
- refactor the TensorFieldData to use FieldDatas and FieldData functions directly
- remove any debug prints that may still be there


src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
```diff
@@ -763,14 +760,23 @@ namespace amr
...
         void registerInitComms(std::unique_ptr<HybridMessengerInfo> const& info)
         {
+
+            auto b_id = resourcesManager_->getID(info->modelMagnetic);
+            BalgoInit.registerRefine(*b_id, *b_id, *b_id, BfieldRefineOp_,
+                                 overwriteInteriorTFfillPattern);
+
+            // no fill pattern given for this init
+            // will use boxgeometryvariable fillpattern, itself using the 
+            // gield geometry with overwrit_interior true from SAMRAI
```
PhilipDeegan, line 778: `gield` :D

tests/simulator/test_advance.py
```diff
@@ -74,7 +74,7 @@ def getHierarchy(
 
         extra_diag_options["mode"] = "overwrite"
         extra_diag_options["dir"] = diag_outputs
-        self.register_diag_dir_for_cleanup(diag_outputs)
+        # self.register_diag_dir_for_cleanup(diag_outputs)
```
PhilipDeegan, line 77: we probably want to revert this

UCaromel:
replaced by #1068

PR #1070, opened 2025-08-01
Mhd refluxing2

UCaromel:
mhd wip


pyphare/pyphare/core/gridlayout.py
```diff
@@ -43,6 +43,18 @@
         "Pyy": "primal",
         "Pyz": "primal",
         "Pzz": "primal",
+
+        # finite volume mhd quantities are 3ple dual
+        "mhdRho": "dual",
```
PhilipDeegan, line 48: I would guess for all dual quantites, we could just have a `"dual": "dual"` entry per dim?

we might want that for other reasons too tbh

PR #1068, opened 2025-08-01
fix master copy before refinement + tensorfield and refluxing

UCaromel:
- fine_boundary_represents_var for copy **before** refinement
- PatchData NaN initialized on construction
- fix tests failing as result of above
- comment a field refinement test (useless, wrong refinement op for E,B)
- debug plots for advance field overlap test (see plot below)
- copy done before refinement (boolean false in variable)
- overwrite_interior false also for refinement is default for FieldFillPattern
- J manually init to zero in model init, fine init and regrid init (Jx unused in ampere but used in Ohm with its now NaN values)
- Grid/NdarrayVector take default value overrides (for test)
- UsableTensorField is default constructed with zero init.
- TensorFieldData
- with refluxing

image

PR #1075, opened 2025-09-18
Fix master

nicolasaunai:
new version of #1068 after PR comments.

- fine_boundary_represents_var changed so to have copy of overlaped data **before** refinement of non overlaped data when SAMRAI execute a schedule. 
- PatchData NaN initialized on construction,
- fix tests failing as result of above
- comment a field refinement test (useless, wrong refinement op for E,B)
- overwrite_interior false also for refinement is default for FieldFillPattern
- J manually init to zero in model init, fine init and regrid init (Jx unused in ampere but used in Ohm with its now NaN values)
- Grid/NdarrayVector take default value overrides (for test)
- UsableTensorField is default constructed with zero init.
- TensorFieldData override of PatchData, now is how vectors (B, E, ...) are seen from SAMRAI's viewpoint. This allows having all components of a vector when executing a schedule (e.g. refinement) instead of registering N components individually to the same RefineAlgorithm (which does not work well in case of a single PatchData type).
- Refluxing is now re-calculating the magnetic field via faraday equation after coarsening of finer accumulated electric fields. Thus B is not coarsened anymore and divB is preserved at the coarse-fine interface.

- [ ] improve in-code documentation/comments in particular for messenger/refluxing/regriding stuff. 

improved debug plots for advance field overlap test (see plot below)

image


src/amr/data/field/refine/electric_field_refiner.hpp
```diff
@@ -4,14 +4,13 @@
 
 #include "core/def/phare_mpi.hpp" // IWYU pragma: keep
 
+#include <SAMRAI/hier/Box.h>
+
+#include "amr/resources_manager/amr_utils.hpp"
```
PhilipDeegan, line 9: on purpose?

src/amr/messengers/hybrid_hybrid_messenger_strategy.hpp
```diff
@@ -323,8 +309,9 @@ namespace amr
             // nodes may not have been copied correctly, due to a bug in SAMRAI
             // it seems these nodes are only on ghost box border if that border
             // overlaps an old level patch border. See https://github.com/LLNL/SAMRAI/pull/293
-            magPatchGhostsRefineSchedules[levelNumber]->fillData(initDataTime);
-            elecPatchGhostsRefineSchedules[levelNumber]->fillData(initDataTime);
+
+            // magPatchGhostsRefineSchedules[levelNumber]->fillData(initDataTime);
+            // elecPatchGhostsRefineSchedules[levelNumber]->fillData(initDataTime);
```
PhilipDeegan, line 314: If we do not switch to the samrai version https://github.com/nicolasaunai/SAMRAI/tree/feature/srcmask

we will see divb in harris on merge (likely)

PR #1090, opened 2025-10-15
pybind updates for latest master

PhilipDeegan:
seeing compile errors since https://github.com/pybind/pybind11/pull/5866 

should be resolved with the additional `requires` 

also updated to smart_holders instead of shared_ptrs as noted here https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#py-smart-holder


PhilipDeegan:
this is unnecessary strictly since pybind was fixed, but maybe we want it regardless

PR #1091, opened 2025-10-17
QOL high rank run changes

PhilipDeegan:
less inodes

### DIFF
--- a/pyphare/pyphare/simulator/simulator.py
+++ b/pyphare/pyphare/simulator/simulator.py
@@ -294,12 +294,13 @@ class Simulator:
         Support keys:
             RANK_FILES - logfile per rank
             DATETIME_FILES - logfile with starting datetime timestamp per rank
-            NONE - no logging files, display to cout
+            CLI  - no logging files, display to cout
+            NULL - no logging files, no cout
         """
-
-        if "PHARE_LOG" not in os.environ:
-            os.environ["PHARE_LOG"] = "RANK_FILES"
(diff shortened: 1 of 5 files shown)


pyphare/pyphare/simulator/simulator.py
```diff
@@ -294,12 +294,13 @@ def _log_to_file(self):
...
+            CLI  - no logging files, display to cout
+            NULL - no logging files, no cout
         """
-
-        if "PHARE_LOG" not in os.environ:
-            os.environ["PHARE_LOG"] = "RANK_FILES"
         from pyphare.cpp import cpp_lib
 
-        if os.environ["PHARE_LOG"] != "NONE" and cpp_lib().mpi_rank() == 0:
+        logging = os.environ["PHARE_LOG"] = os.environ.get("PHARE_LOG", "RANK_FILES")
```
PhilipDeegan: possible we should default to no files, or maybe if there's more than 1000 ranks or something

pyphare/pyphare/simulator/simulator.py
```diff
@@ -294,12 +294,13 @@ def _log_to_file(self):
         Support keys:
             RANK_FILES - logfile per rank
             DATETIME_FILES - logfile with starting datetime timestamp per rank
-            NONE - no logging files, display to cout
+            CLI  - no logging files, display to cout
+            NULL - no logging files, no cout
```
PhilipDeegan: could be `COUT` alternatively

Issue #1103, opened 2025-11-06
Restarts allow restart_time 0 and "auto"


PhilipDeegan:
Allowing restart time = 0 could be nice to not have to keep the parameter commented out until needed

Restart time "auto" could scan the checkpoints directory for the most recent (latest timestamp) dump

PhilipDeegan:
closed by #1105

PR #1105, opened 2025-11-07
convenience restart options

### LINKED ISSUE #1102: Restarts only keep N most recent dumps
restarts can be quite large, and it's somewhat tedious deleting ones that may be considered redundant.

PhilipDeegan:
give to close 

#1102
#1103

todo: input validations

### DIFF
--- a/pyphare/pyphare/pharein/restarts.py
+++ b/pyphare/pyphare/pharein/restarts.py
@@ -59,6 +59,7 @@ def restart_time(restart_options):
         if restart_options["restart_time"] == "auto":
             return find_latest_time_from_restarts(restart_options)
         return restart_options["restart_time"]
+    return None
 
 
 def find_latest_time_from_restarts(restart_options):


pyphare/pyphare/pharein/restarts.py
```diff
@@ -1,8 +1,75 @@
...
+    """
+    0.006 == "00000.00600"
+    """
+    return "{:0>11.5f}".format(time)
+
+
+def dump(simulator, time, time_step):
+    if not simulator.cpp_sim.dump_restarts(timestamp=time, timestep=time_step):
+        return
+    # else a restart dump was made
```
nicolasaunai: tell me if I get it correctly, this function dumps checkpoint if time is dump time AND it tries to delete some previous ones based on restart options.
I feel it could be clearer if everything below that "else" comment would be in a function like "remove_previous _checkpoints" or some shorter version of that

PR #1111, opened 2025-11-20
direct imports / highfive full inc path

PhilipDeegan:
attempt to fix LSP/clangd 
cleanup dict includes


src/core/utilities/types.hpp
```diff
@@ -221,6 +223,18 @@ namespace core
 
 
 
+    template<typename T>
```
PhilipDeegan: this wasn't meant to be here, but it's on the vtk branch too so it will be likely merged sometime anyway

PR #1115, opened 2025-12-08
overlap Difference hierarchies

PhilipDeegan:
output from new harris diff test

```
diff 0.0
B max:  [np.float64(5.10702591327572e-15), np.float64(0.0), np.float64(0.0)]
ion charge rho max:  [np.float64(0.0), np.float64(0.0), np.float64(0.0)]
ion mass rho max:  [np.float64(0.0), np.float64(0.0), np.float64(0.0)]
E max:  [np.float64(1.8561541192951836e-16), np.float64(0.0), np.float64(0.0)]

diff 0.005
B max:  [np.float64(6.505213034913027e-19), np.float64(0.0), np.float64(0.0)]
ion charge rho max:  [np.float64(0.0), np.float64(0.0), np.float64(0.0)]
ion mass rho max:  [np.float64(0.0), np.float64(0.0), np.float64(0.0)]
E max:  [np.float64(1.3877787807814457e-17), np.float64(0.0), np.float64(0.0)]

diff 0.01
B max:  [np.float64(6.505213034913027e-19), np.float64(0.0), np.float64(0.0)]
ion charge rho max:  [np.float64(0.0), np.float64(0.0), np.float64(0.0)]
ion mass rho max:  [np.float64(0.0), np.float64(0.0), np.float64(0.0)]
E max:  [np.float64(3.469446951953614e-18), np.float64(0.0), np.float64(0.0)]
```

### DIFF
--- a/pyphare/pyphare/pharesee/hierarchy/hierarchy_utils.py
+++ b/pyphare/pyphare/pharesee/hierarchy/hierarchy_utils.py
@@ -147,7 +147,7 @@ def compute_hier_from(compute, hierarchies, **kwargs):
     patch_levels_per_time = []
     for t in reference_hier.times():
         patch_levels = {}
-        for ilvl in range(reference_hier.levelNbr()):
+        for ilvl in reference_hier.levels(t).keys():
             patch_levels[ilvl] = PatchLevel(
                 ilvl, new_patches_from(compute, hierarchies, ilvl, t, **kwargs)
             )
@@ -182,6 +182,10 @@ def new_patchdatas_from(compute, patchdatas, layout, id, **kwargs):
 
 def new_patches_from(compute, hierarchies, ilvl, t, **kwargs):
     reference_hier = hierarchies[0]
+
+    if not reference_hier.has_time(t):
+        return []
+
     new_patches = []
     ref_patches = reference_hier.level(ilvl, time=t).patches
     for ip, current_patch in enumerate(ref_patches):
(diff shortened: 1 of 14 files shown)


pyphare/pyphare/pharesee/hierarchy/hierarchy_utils.py
```diff
@@ -612,6 +612,64 @@
         return reversed(self.failed)
 
 
+def diff_hierarchy(hier):
+    import pyphare.core.box as boxm
+    from pyphare.pharesee.geometry import hierarchy_overlaps
```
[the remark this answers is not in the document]
PhilipDeegan: resolved

pyphare/pyphare/pharesee/hierarchy/hierarchy.py
```diff
@@ -614,6 +614,53 @@ def dist_plot(self, **kwargs):
...
+
+        for ilvl, lvl in hier.items():
+            for patch in lvl:
+                for key, pd in patch.patch_datas.items():
+                    patch.patch_datas[key] = pd.zeros_like()
+        assert not copy.has_non_zero()
+        return copy
+
+    def has_non_zero(self):
+        """only works for fields and vecfields with 1 time"""
```
nicolasaunai: I don't get why "field and vecfields" actually. I get the "1 time" because your take the [0] but not why it should only apply to field and vector fields

PhilipDeegan: I should have said tensorfields, it doesn't work for particles cause it makes no sense

pyphare/pyphare/pharesee/hierarchy/hierarchy.py
```diff
@@ -614,6 +614,53 @@ def dist_plot(self, **kwargs):
...
+                        return True
+        return False
+
+    def max(self, qty=None):
+        """only works for fields and vecfields with 1 time"""
+
+        assert len(self.time_hier) == 1
+
+        hier = (list(self.time_hier.values()))[0]
+        val = [0 for ilvl, lvl in hier.items()]
```
nicolasaunai: why taking 0 as the initial max ? I don't think this works if 0 is not part of the values or I miss the point of that function. Also, if this function intends to return the max per level for a quantity maybe this should lead to the refactoring of `global_max()`which would thus be the max(of that).

pyphare/pyphare/pharesee/hierarchy/hierarchy.py
```diff
@@ -614,6 +614,53 @@ def dist_plot(self, **kwargs):
...
+                    if len(check[0]):
+                        return True
+        return False
+
+    def max(self, qty=None):
+        """only works for fields and vecfields with 1 time"""
+
+        assert len(self.time_hier) == 1
+
+        hier = (list(self.time_hier.values()))[0]
```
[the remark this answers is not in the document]
PhilipDeegan: sure

pyphare/pyphare/pharesee/hierarchy/hierarchy_utils.py
```diff
@@ -612,6 +612,64 @@ def __reversed__(self):
...
+
+    diff_hier = hier.zeros_like()
+    time = list(diff_hier.time_hier.keys())[0]
```
nicolasaunai: should also document it only works for first time.
Or maybe have the zeros_like function account for all times and loop over times in the current code

nicolasaunai (commented):
ok there was also to  update the default smallest patch size to a temporary larger value than what the formula gives, commenting that it's tested manually to work for interp all but 1, and refer to https://github.com/llnl/SAMRAI/issues/311

Also, the overlap diff hierarchy was originally intended to be used in field overlap tests which are atm inconvenient to debug, do we do that in there or in a separate one

src/amr/messengers/field_operate_transaction.hpp
```diff
@@ -14,31 +14,18 @@ namespace PHARE::amr
...
- *
- * Due to the lack of neighbor particle contributions, some domain nodes and ghost nodes
- * have incomplete moments after deposition. The complement of these nodes is what has
- * been deposited on (also incomplete) neighbor nodes.
- *
- * Default SAMRAI transaction calls PatchData::copy and PatchData::packStream
- * This transaction defines these override to these methods to call specific methods
- * of FieldData to perform the += instead of =.
- * These methods are copyAndSum and unpackStreamAndSum.
+/** * @brief FieldBorderOpTransaction is TODO
```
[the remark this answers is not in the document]
nicolasaunai: same as previous comment without Sum ?

pyphare/pyphare/pharesee/hierarchy/hierarchy.py
PhilipDeegan: frankly I don't think global_min/max should even be on the PatchHierarchy class
it makes no sense for particles, if also moved to a free function then that's probably simpler

nicolasaunai: agreed global max should not be for particle hierarchies, makes no sense

PR #1123, opened 2026-01-12
hybrid emergency dumps

UCaromel:


nicolasaunai:
post meeting TODO : have the simulator dict from python taking a param that enables/disable(default) the exception dump

PR #1153, opened 2026-02-13
less costly  box iteration

### LINKED ISSUE #1020: Minimize field access operations
Currently in 3d, we compute the 1d offset for every individual field lookup given the array shape and 3d mesh index

Some previous instruction analyses with omniperf was showing an alarmly large number of int operations which I suspect is related

solution might be to work on something similar to the box iterator so we can reduce operations to mostly just an integer increment

PhilipDeegan:
closes #1020

it's possible we should update field equations to pass the values per index, rather than do the index lookup per operation

the [changes to field_refine_operator.hpp](https://github.com/PhilipDeegan/PHARE/blob/443f89bf116854f73a7a905ab667c2c4eab1bf0f/src/amr/data/field/refine/field_refine_operator.hpp#L50)  are somewhat complex, but hopefully the benchmark results below can justify the change.

benchmark of old field refine operator for scalar field 

```
time ./tools/bench/amr/data/field/refinement/phare_bench_field_refine_bench_field_refinement
tools/bench/amr/data/field/refinement/bench_field_refinement.cpp:41 Time: 33876.3 ms
tools/bench/amr/data/field/refinement/bench_field_refinement.cpp:76 RHO SUM: 201230056
tools/bench/amr/data/field/refinement/bench_field_refinement.cpp:87 0

real 0m34.231s
user 0m33.975s
sys  0m0.149s
```

vs  new

```
time ./tools/bench/amr/data/field/refinement/phare_bench_field_refine_bench_field_refinement
tools/bench/amr/data/field/refinement/bench_field_refinement.cpp:41 Time: 8314.03 ms
tools/bench/amr/data/field/refinement/bench_field_refinement.cpp:77 RHO SUM: 201230056
tools/bench/amr/data/field/refinement/bench_field_refinement.cpp:88 0

real 0m8.682s
user 0m8.417s
sys  0m0.165s
```

[nightly artifacts](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Phil_Build/94226?buildTab=artifacts)

### DIFF
--- a/src/amr/data/field/refine/field_refine_operator.hpp
+++ b/src/amr/data/field/refine/field_refine_operator.hpp
@@ -5,6 +5,7 @@
 #include "core/def/phare_mpi.hpp" // IWYU pragma: keep
 #include "core/utilities/box/box_span.hpp"
 
+#include "amr/amr_constants.hpp"
 #include "amr/data/field/field_data.hpp"
 #include "amr/resources_manager/amr_utils.hpp"
 #include "amr/data/tensorfield/tensor_field_data.hpp"
@@ -85,7 +86,7 @@ void refine_field(core::FieldBox<Field_t>& dst, core::FieldBox<Field_t const>& s
                 refiner(src.field, dst.field, d_amr_point, s_amr_point, d_lcl_point, s_lcl_point,
                         d_span[dst_idx], s_span[src_idx]);
 
-                if (d_amr_point[IDX] % 2 != 0)
+                if (d_amr_point[IDX] % refinementRatio != 0)
                 {
                     ++src_idx;
                     ++s_lcl_point[IDX];
@@ -96,14 +97,14 @@ void refine_field(core::FieldBox<Field_t>& dst, core::FieldBox<Field_t const>& s
                 ++d_lcl_point[IDX];
             }
 
-            if (span_idx % 2 != 0)
+            if (span_idx % refinementRatio != 0)
             {
                 ++s_f_spans;
                 ++s_b_spans;
             }
         }
 
-        if (slab_idx % 2 != 0)
+        if (slab_idx % refinementRatio != 0)
         {
             ++s_f_slabs;
             ++s_b_slabs;
--- a/src/amr/data/field/refine/field_refine_operator.hpp
+++ b/src/amr/data/field/refine/field_refine_operator.hpp
@@ -1,18 +1,14 @@
 #ifndef PHARE_FIELD_REFINE_OPERATOR_HPP
 #define PHARE_FIELD_REFINE_OPERATOR_HPP
 
-
-
-#include "core/def/phare_mpi.hpp" // IWYU pragma: keep
-
 #include "core/def.hpp"
+#include "core/def/phare_mpi.hpp" // IWYU pragma: keep
+#include "core/utilities/box/box_span.hpp"
 
 #include "amr/data/field/field_data.hpp"
+#include "amr/resources_manager/amr_utils.hpp"
 #include "amr/data/tensorfield/tensor_field_data.hpp"
 #include "amr/resources_manager/tensor_field_resource.hpp"
-
-#include "field_linear_refine.hpp"
-#include "field_refiner.hpp"
 
 #include <SAMRAI/tbox/Dimension.h>
 #include <SAMRAI/hier/RefineOperator.h>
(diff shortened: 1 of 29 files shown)


src/core/data/field/field_box.hpp
```diff
@@ -58,36 +61,108 @@ class FieldBox
...
+    {
+        auto& dst_val = dst.field(*dst_it);
+        auto& src_val = src.field(*src_it);
+        dst_val       = std::max(dst_val, src_val);
+    }
 }
 
 
+template<typename Operator>
+void set_field_row_from(auto& dst, auto const* src_data)
```
nicolasaunai: these two functions are the same as above `operate_on_fields` why

src/core/data/field/field_box.hpp
```diff
@@ -58,36 +61,108 @@ class FieldBox
...
+    using value_type = Field_t::value_type;
     // reserve vec before use!
-    auto src_it = lcl_box.begin();
-    for (; src_it != lcl_box.end(); ++src_it)
-        vec.push_back(field(*src_it));
+    std::size_t seek = vec.size();
+    vec.resize(vec.size() + lcl_box.size());
+
+    for (auto const& slab : make_field_box_span(lcl_box, field))
+        for (auto const& row : slab)
```
nicolasaunai: rename row?

src/core/data/grid/gridlayout.hpp
```diff
@@ -1176,6 +1176,14 @@ namespace core
         }
 
 
+        template<typename Field>
+        Box<std::uint32_t, dimension> domainBoxFor(Field const& field) const
+        {
+            return _BoxFor(field, [&](auto const& centering, auto const direction) {
```
nicolasaunai: our private functions have a trailing underscore not a leading one

src/core/data/field/field_box_span.hpp
```diff
@@ -0,0 +1,134 @@
...
+        if constexpr (dim == 2)
+            return (row = Span<value_type>{&arr(row_idx, box.lower[1]), row_size});
+        if constexpr (dim == 3)
+            return (row = Span<value_type>{&arr(slab_idx, row_idx, box.lower[2]), row_size});
+    }
+
+    bool operator==(FieldBoxRows const& that) const { return row_idx == that.row_idx; }
+    bool operator!=(FieldBoxRows const& that) const { return row_idx != that.row_idx; }
+
+    auto _row_size() const { return box.upper[dim - 1] - box.lower[dim - 1] + 1; }
```
nicolasaunai: consider using class rather than struct and public/private keywords if there are functions that are only meant for private. Prefer structs for behaviorless container types or constexpr stuff.

PhilipDeegan: the reason it's like this as it's not strictly finished given I'm still updating parts of the code and do not yet know what the final interface will be

you can assume that when that is done it will be more obvious in terms of accessor visibility

for now the logic/usage is probably most important

nicolasaunai:
> it's possible we might be able to use the md_span here if we were to switch to C++23

then we need to assess that and come back to this PR after?

nicolasaunai:
> > > it's possible we might be able to use the md_span here if we were to switch to C++23
> > 
> > 
> > then we need to assess that and come back to this PR after?
> 
> we might not be able to switch to md_span just yet, so we'll need this either way, then maybe we can dedupe later

why wouldn't we be able?

PhilipDeegan:
not sure why it wouldn't, just that we haven't, so unknown unknowns

other than that I think this is ready, it should hopefully be visible in runtime

nicolasaunai:
have you ran with/without this locally for a full simulation like a harris to see if that leads to a difference, like with phlop?
That'd be nice.

src/amr/data/field/refine/field_refine_operator.hpp
```diff
@@ -28,13 +25,91 @@ using core::dirX;
...
+    CrossLevelIndices<Field_t> indices{dst, src};
+
+    auto d_f_slabs = indices.dst_lcl_span.begin();
+    auto d_b_slabs = indices.dst_amr_span.begin();
+    auto s_f_slabs = indices.src_lcl_span.begin();
+    auto s_b_slabs = indices.src_amr_span.begin();
+
+    auto slab_idx = indices.dst_amr_span.slab_begin();
+
+    for (; d_f_slabs != indices.dst_lcl_span.end(); ++d_f_slabs, ++d_b_slabs, ++slab_idx)
```
nicolasaunai: this is far less obvious than the previous "naive" implementation.
how can we make it more readable? Maybe a comment explaining the logics?

src/amr/data/field/refine/field_refine_operator.hpp
```diff
@@ -28,13 +25,91 @@ using core::dirX;
 using core::dirY;
 using core::dirZ;
 
+template<typename Field_t>
+struct CrossLevelIndices
```
nicolasaunai: maybe the logics of the refinement would be clearer if that was somehow documented: what does it represents

PhilipDeegan: it tracks which fine cells are related to which coarse cells essentially, with cross level indicies, 

some of hte variable names could be more clear that's for sure, repeating the use of very large variable names is somewhat cumbersome, so we might have some way to show how a short variable name relates to a longer, or some comments saying which one is,

PhilipDeegan:
timings on my machine as basically identical but what we can see is variance in instructions and branches 

harris_3d cells = (40, 80, 20) with 3 levels, 4 cores over 5 coarse timesteps


master rank 0 results
```
perf stat -B -e cache-references,cache-misses,cycles,instructions,branches,faults,migrations python3 tests/functional/harris/harris_3d.py

 Performance counter stats for 'python3 tests/functional/harris/harris_3d.py':

    39,871,695,404      cache-references                                                      
     2,135,431,661      cache-misses                     #    5.36% of all cache refs         
   903,143,012,617      cycles                                                                
 2,733,630,811,988      instructions                     #    3.03  insn per cycle            
    73,498,557,998      branches                                                              
           313,579      faults                                                                
               214      migrations                                                            

     173.328680670 seconds time elapsed

     167.924342000 seconds user
       5.243729000 seconds sys
```


with this new box iterator rank 0 results
```
perf stat -B -e cache-references,cache-misses,cycles,instructions,branches,faults,migrations python3 tests/functional/harris/harris_3d.py

 Performance counter stats for 'python3 tests/functional/harris/harris_3d.py':

    39,879,680,938      cache-references                                                      
     2,186,212,762      cache-misses                     #    5.48% of all cache refs         
   910,319,185,397      cycles                                                                
 2,714,382,764,219      instructions                     #    2.98  insn per cycle            
    70,849,802,233      branches                                                              
           310,231      faults                                                                
               180      migrations                                                            

     174.751022567 seconds time elapsed

     169.120946000 seconds user
       5.467868000 seconds sys
```

nicolasaunai:
I remember you said there was a lot of time spent in there, why is it not seen in your timings?

src/core/numerics/ampere/ampere.hpp
```diff
@@ -31,6 +31,9 @@ class Ampere
 
         auto const shrink = ConstArray<std::size_t, dimension>(1);
 
+        // evalOnShrinkedGhostBox calls fn(point, trailingArgs...); trailing args
```
PhilipDeegan, line 34: personally I think this is redundant

PhilipDeegan:
> I remember you said there was a lot of time spent in there, why is it not seen in your timings?

My machine has 100MB of L3 cache which is very much not the norm so I will check on kaa probably

nicolasaunai:
on adastra or JZ seems closer to real use case?

PhilipDeegan:
adastra 12288 cores

[master](https://gist.github.com/PhilipDeegan/a6d3ffaf44cf2d55cf33c79a7788e8e7)
[new box iter](https://gist.github.com/PhilipDeegan/c0c7bb39441da88eaaf72e862b5ac0b8)

master 
SolverPPC::moveIons::fillIonBorders                       114.37ms

new box iter
SolverPPC::moveIons::fillIonBorders                         3.00ms

PhilipDeegan:
claudio's analyses 

```
- SolverPPC::moveIons::fillIonBorders, fillFluxBorders, fillDensityBorders, moveIons — only ~52-62% consistent (roughly a coin flip: e.g. fillIonBorders is faster in 16/42 calls and slower in 26/42). Individual calls swing wildly (+1865%, -97%, +816%, -94%...) with no stable direction. That's the signature of MPI straggler/imbalance noise on a 12k-core run, not a real regression from the box-iterator change. I'd retract that "regression" claim from my earlier summary.
- SolverPPC::predictor1_ and its .faraday::schedules/.ampere::schedules children, Multiphys::standardLevelSynchronization, advanceLevel — 67-76% consistent, and the same direction (faster) every time it's large. That's a much more defensible real signal: the schedule-building/communication path in predictor1 and level-sync genuinely got faster on newboxiter, and that's what's driving the net -5.2%/-4,776ms on Simulator::advance.

So the trustworthy takeaway: predictor1_'s schedules and level-sync are consistently and substantially faster on newboxiter; the moveIons border-fill fluctuations are most likely communication-imbalance noise inherent to the run, not a code-path difference worth chasing.
```

Issue #1095, opened 2025-10-28
code duplication in gridlayout projection


PhilipDeegan:
lots of what looks like copy pastes in projection across quantities

for instance, this block exists four times

https://github.com/PHAREHUB/PHARE/blob/master/src/core/data/grid/gridlayoutimplyee.hpp#L607-L626

PR #1169, opened 2026-03-04
refac projection functions

UCaromel:
This is a proposition for a refactor of our projection function as requested in #1095. It also fixes so 8 point stencils that were using the wrong weights in 3d, such as ```BzToEz()``` and ```ByToEy()```. 
It uses a systematic formula to create a 1d stencil in the requested direction to either project form primal to dual or dual to primal. multi dimensional stencils required when at least 2 centerings need to change with the projections are generated using the tensor product of all the 1d stencil involved.

### DIFF
--- a/src/core/data/grid/gridlayout.hpp
+++ b/src/core/data/grid/gridlayout.hpp
@@ -827,22 +827,8 @@ namespace core
             typename Field::type result = 0.;
 
             for (auto const& wp : wps)
-            {
-                if constexpr (dimension == 1)
-                {
-                    result += wp.coef * field(index[0] + wp.indexes[0]);
-                }
-                if constexpr (dimension == 2)
-                {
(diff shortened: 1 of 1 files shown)


src/core/data/grid/gridlayoutimplyee.hpp
```diff
@@ -370,140 +370,143 @@ namespace core
...
+                }
+            return result;
         }
 
-
-
+        template<auto dir1, auto dir2, auto dir3>
+        static constexpr auto tensorProduct(auto const& s1, auto const& s2, auto const& s3)
+        {
+            constexpr auto N1 = std::tuple_size_v<std::decay_t<decltype(s1)>>;
```
PhilipDeegan: `constexpr auto N1 = s1.size();` does this not work?

I know using tuple size works on std::array, but s1 is not a tuple, if it was the loop would fail

PR #1178, opened 2026-03-23
drop ppc model view

PhilipDeegan:
I think we can just the the LevelLooper instead if we really need to, which for GPU will likely be necessary


nicolasaunai:
we pick the constructor version and passing all physical quantities at () call site

UCaromel (approved):

PR #1193, opened 2026-04-10
try not fill currentGhosts

PhilipDeegan:
Claude, in an attempt to minimize schedules, suggested this.

### DIFF
--- a/src/core/data/grid/gridlayout.hpp
+++ b/src/core/data/grid/gridlayout.hpp
@@ -1186,6 +1186,20 @@ namespace core
             evalOnBox_(field, fn, indices);
         }
 
+        // Iterates over the first ghost layer only (physStart-1 .. physEnd+1).
+        // Safe to use for Ampere after fillMagneticGhosts: B has >= 2 ghost layers so the
+        // curl stencil at ghost depth 1 always has valid B on both sides.
+        template<typename Field, typename Fn>
+        void evalOnFirstGhostLayer(Field& field, Fn&& fn) const
+        {
+            auto indices = [&](auto const& centering, auto const direction) {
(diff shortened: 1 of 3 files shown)


src/core/data/grid/gridlayout.hpp
```diff
@@ -1186,6 +1186,20 @@ namespace core
             evalOnBox_(field, fn, indices);
         }
 
+        // Iterates over the first ghost layer only (physStart-1 .. physEnd+1).
+        // Safe to use for Ampere after fillMagneticGhosts: B has >= 2 ghost layers so the
+        // curl stencil at ghost depth 1 always has valid B on both sides.
+        template<typename Field, typename Fn>
+        void evalOnFirstGhostLayer(Field& field, Fn&& fn) const
```
UCaromel, line 1193: We are using a similar mecanism in mhd. i suggest to make this function slightly more general as we are likely to need it in the mhd pr also:
```cpp
        template<typename Field, typename Fn>
        void evalOnBiggerBox(Field& field, Point<uint32_t, dimension> const& grow, Fn&& fn) const
        {
            auto indices = [&](auto const& centering, auto const direction) {
                auto [start, end] = this->physicalStartToEnd(centering, direction);
                return std::make_pair(start - grow[static_cast<std::size_t>(direction)],
                                      end + grow[static_cast<std::size_t>(direction)]);
            };

            evalOnBox_(field, fn, indices);
        }
```

PhilipDeegan: the plan eventually is to stop using this "indices" stuff here and just do [this](https://github.com/PhilipDeegan/PHARE/blob/mkn/src/core/data/grid/gridlayout.hpp#L1155)

PR #1200, opened 2026-05-01
loop over evalonbox

PhilipDeegan:
Will be useful for MHD to stop doing evalOnGhostBox to stop using the box iterator


nicolasaunai (approved):

PhilipDeegan:
this will likely converge later on what I have done for tiling, which is to have wrappers on std::transform for tiles and non-tiled fields

PR #1205, opened 2026-05-12
single auto template for gridlayout for multi-model flexibility

### LINKED ISSUE #951: ghost nbr constraints
The `GridLayout` gives us the number of ghosts, used in many occasions and in particular useful for SAMRAI.
The number of ghost nodes is different for fields and particles and these two numbers are the upper bounds of a handful of constrains.
For instance, we need X number of ghost cells for particles, and because those need to be pushed, we need Y number of field ghost cells...

The arrival of MHD brings additional constrains, like WENO-3 needs 2 ghost cells but WENO-Z needs 3. 
Coupled to, say, a interp order 1 hybrid that needs 2 ghost cells, WENO-Z would bring the need for 1 additional field ghost cell.

Questions:

- do we want or need the same number of field ghost cells regardless of the level / model? Like, can we keep 2 on hybrid levels and have 5 on MHD ones?
- how do we tell the gridlayout that we have, say, WENO-Z and need 3 field ghost cells

For now the GridLayout knows everything:

```cpp
    template<std::size_t interpOrder>
    NO_DISCARD std::uint32_t constexpr ghostWidthForParticles()
    {
        return (interpOrder % 2 == 0 ? interpOrder / 2 + 1 : (interpOrder + 1) / 2);
    }

        NO_DISCARD static std::size_t constexpr nbrParticleGhosts()
        {
            return ghostWidthForParticles<interp_order>();
        }


        NO_DISCARD std::uint32_t constexpr static nbrDualGhosts_()
        {
            // we need (interp_order+1)/2 field ghosts **outside** the particle ghost layer
            // for ghost particles to interact with the mesh.
            return (interp_order + 1) / 2 + nbrParticleGhosts();
        }

```

PhilipDeegan:
closes #951

basically a model aware gridlayout so whatever needs to be known in some context can be, so interp_order for hybrid is still available, but it's not available for mhd

introduces two new concepts

HybridFieldOptions and HybridOptions, which lets us define compile time values, like ghost width, which are then used as template parameters, giving the GridLayout less responsibility

https://github.com/PHAREHUB/PHARE/blob/ec9bac1131e1df89e03fc15908a99164bfe9333b/src/phare_core.hpp#L40-L47

future work includes making the Field class single auto template too using HybridFieldOptions or MHDFieldOptions, but one thing at a time


src/amr/solvers/solver_ppc.hpp
```diff
@@ -31,8 +31,8 @@ template<typename HybridModel, typename AMR_Types>
 class SolverPPC : public ISolver<AMR_Types>
 {
 private:
-    static constexpr auto dimension    = HybridModel::dimension;
-    static constexpr auto interp_order = HybridModel::gridlayout_type::interp_order;
+    static constexpr auto dimension = HybridModel::dimension;
+    // static constexpr auto interp_order = HybridModel::gridlayout_type::interp_order;
```
PhilipDeegan: torm

src/core/CMakeLists.txt
```diff
@@ -64,8 +63,8 @@ endif(withPhlop)
 
 add_library(${PROJECT_NAME}  ${SOURCES_INC} ${SOURCES_CPP} ${CORE_EXTRA_SOURCES_CPP} )
 target_compile_options(${PROJECT_NAME}  PRIVATE ${PHARE_WERROR_FLAGS})
-target_link_libraries(${PROJECT_NAME}  PRIVATE phare_initializer MPI::MPI_C cppdict
-    PUBLIC ${PHARE_BASE_LIBS}
+target_link_libraries(${PROJECT_NAME}  PRIVATE phare_initializer MPI::MPI_C
+    PUBLIC cppdict ${PHARE_BASE_LIBS}
```
nicolasaunai, line 69: why is cppdict now public?

tests/amr/resources_manager/test_resources_manager.hpp
```diff
@@ -31,7 +34,8 @@ class aResourceUserCollection : public ::testing::Test
 public:
     using Grid_t = Grid<NdArrayVector<1>, HybridQuantity::Scalar>;
     std::unique_ptr<BasicHierarchy> hierarchy;
-    ResourcesManager<GridLayout<GridLayoutImplYee<1, 1>>, Grid_t> resourcesManager;
+    ResourcesManager<PHARE::core::PHARE_Types<PHARE::SimOpts{1, 1}>::Hybrid::GridLayout_t, Grid_t>
```
nicolasaunai: there, interp_order is 0 by default, is that ok?

src/core/data/grid/gridlayout.hpp
```diff
@@ -1117,12 +1120,11 @@ namespace core
             return box;
         }
 
-
         auto AMRGhostBoxFor(auto const& field) const
         {
             auto const centerings = centering(field);
             return grow(AMRBoxFor(field), for_N_make_array<dimension>(
-                                              [&](auto i) { return nbrGhosts(centerings[i]); }));
+                                              [&](auto i) { return options.field_ghost_width; }));
```
nicolasaunai, line 1124: this makes `centerings` unused

src/core/CMakeLists.txt
```diff
@@ -23,7 +23,6 @@ set( SOURCES_INC
      data/vecfield/vecfield.hpp
      data/vecfield/vecfield_component.hpp
      data/vecfield/vecfield_initializer.hpp
-     hybrid/hybrid_quantities.hpp
```
nicolasaunai, line 26: probably should be renamed rather than removed

src/core/models/options/hybrid_options.hpp
```diff
@@ -0,0 +1,46 @@
...
+struct HybridFieldOptions
+{
+    using Quantity = core::HybridQuantity;
+    using Scalar   = Quantity::Scalar;
+    using Vector   = Quantity::Vector;
+    using Tensor   = Quantity::Tensor;
+
+    auto static constexpr dimension    = opts.dimension;
+    auto static constexpr interp_order = opts.interp_order;
+    std::size_t field_ghost_width      = std::array{2, 4, 4}[interp_order - 1];
```
nicolasaunai: shouldn't these be `auto static constexpr` as in MHDFieldOptions?

src/amr/solvers/solver_ppc.hpp
nicolasaunai: indeed

src/core/data/grid/impl/yee/gridlayout_hybrid_yee.hpp
```diff
@@ -0,0 +1,801 @@
...
+ * - physical start/end indexes
+ * - ghost start/end indexes
+ * - numbers of padding cells and physical cells
+ */
+
+
+template<auto options>
+class GridLayoutImplYee
+{
+    using Options = decltype(options);
```
nicolasaunai, line 30: there could be a static_assert that options are indeed hybrid options?

src/python3/patch_data.hpp
```diff
@@ -56,8 +56,9 @@ void setPatchDataFromField(PatchData& pdata, Field const& field, GridLayout& gri
                            std::string patchID)
 {
     setPatchDataFromGrid(pdata, grid, patchID);
-    pdata.nGhosts = static_cast<std::size_t>(
-        GridLayout::nbrGhosts(GridLayout::centering(field.physicalQuantity())[0]));
+    pdata.nGhosts = GridLayout::options.field_ghost_width;
+    // static_cast<std::size_t>(
```
nicolasaunai: TORM

src/phare_core.hpp
```diff
@@ -29,43 +30,65 @@ namespace PHARE::core
...
+        auto static constexpr mhd_options   = MHDOptions<field_options>{};
+        using Grid_t                        = Grid<Array_t, MHDQuantity::Scalar>;
+        using Field_t                       = Field<dimension, MHDQuantity::Scalar>;
+        using VecField_t                    = VecField<Field_t, MHDQuantity>;
+        using SymTensorField_t              = SymTensorField<Field_t, MHDQuantity>;
+        using Electromag_t                  = Electromag<VecField_t>;
+        using GridLayout_t                  = GridLayout<mhd_options>;
+    };
+
+    // deprecated defaults
```
nicolasaunai, line 76: can you be more specific about why they're still here? the comment about their deprecation confuses me

src/core/CMakeLists.txt
PhilipDeegan: because now we do [this](https://github.com/PhilipDeegan/PHARE/blob/7598433d33c9499b0221bc5c1df0224b0d063553/tests/core/data/particle_initializer/test_main.cpp#L2) 

it's either cppdict or phare_initializer, phare_initializer exposes cppdict as public

tests/amr/resources_manager/test_resources_manager.hpp
PhilipDeegan: in this context I don't think it matters, but it could be simpler just to set it to 1

Interpoloation has `static_assert(0<interp<4)` checks so if it was actually used it wouldn't compile

Here it's used to decide the ghost with of particle ghost boxes for ParticleDatas

but I think the comment here suggests what we should do

https://github.com/PhilipDeegan/PHARE/blob/gridlayout/src/amr/resources_manager/resources_manager.hpp#L115

src/core/data/grid/gridlayout.hpp
PhilipDeegan: so it does, weird it didn't werror

src/core/models/options/hybrid_options.hpp
PhilipDeegan: it's not obvious, for Simulator Options we do [this](https://github.com/PhilipDeegan/PHARE/blob/gridlayout/src/phare_simulator_options.hpp#L21-L36)

which lets it be used as both a compile time and runtime mutable class

it probably could be all `auto static constexpr` for these new Options tho

src/core/CMakeLists.txt
PhilipDeegan: we probably need monsieur calude to have some scanner code gen script thing to just generate the entire block per module and we diff it on PR

src/core/data/grid/impl/yee/gridlayout_hybrid_yee.hpp
PhilipDeegan: not sure how you'd do that except with some gross like adding a "bool is_hybrid = true" to HybridOptions and checking that it compiles

src/phare_core.hpp
PhilipDeegan: it's an open question

if we delete them here and now we'd have to make more changes (in tests)

hybrid was always the default, as it was the only one, so it could go either way

src/core/CMakeLists.txt
PhilipDeegan: I also don't have any clang format issues, and we don't compile headers, so I fail to see why they need to be documented, unless you continue to also have clang format issues I suppose

src/core/CMakeLists.txt
PhilipDeegan: https://github.com/PHAREHUB/PHARE/blob/master/src/phare_core.hpp#L20
https://github.com/PHAREHUB/PHARE/blob/master/src/core/errors.hpp#L12

src/core/CMakeLists.txt
nicolasaunai: clangd, which nvim uses, uses the compilation database so I don't think it uses these files.

PhilipDeegan: yes I meant clangd indeeeeeed

tests/amr/resources_manager/test_resources_manager.hpp
PhilipDeegan: I figured out how to drop interp_order entirely from the ResourcesManager as it's only ever needed for ParticlesTypes, which now is available via gridlayout in that context

src/core/CMakeLists.txt
nicolasaunai: so we get rid of them all?

UCaromel (approved):

src/core/CMakeLists.txt
PhilipDeegan: could be the answer

src/phare_core.hpp
nicolasaunai: There is a lack of symmetry introduced by these shortcuts that I find confusing.
phare_types in phare_solver.hpp are taking these shortcuts whereas it would be more obvious that it takes Hybrid:: as the MHD part does.
But more confusing is the PHARE_Types in the simulator PHARE_Types, for which, if I get correctly, low level types like Grid_t, Field_t etc. all must come from the hybrid aliases "silently". If it works it must be that the code that uses them (what is it?) only needs hybrid types and not MHD types, but from reading the code I find it confusing.
Maybe this should be symmetrized a bit

PhilipDeegan: yeah and if we deleted the deprecated "defaults" that would have caused a compile error in phare_solver.hpp
as it should have been "::Hybrid::" but I forgot to update it

nicolasaunai (approved):

PR #1207, opened 2026-05-20
initial changes for rank3 tensor

PhilipDeegan:
there should be a reduced heatflux vecfield option via run.py which looks like 

```
xxx + xyy + xzz
xxy + yyy + yzz
xxz + yyz + zzz
```


PhilipDeegan:
not needed now

PR #1214, opened 2026-05-28
expose TileClustering tile_size and allow_remote_tile_extent params

UCaromel:
Expose Tile clustering parameters to samrai from python. 
- tile_size (default: 8): tile_size used in initial clustering
- allow_remote_tile_extent (default: true): if false, does process local clustering.

### DIFF
--- a/pyphare/pyphare/pharein/simulation.py
+++ b/pyphare/pyphare/pharein/simulation.py
@@ -641,12 +641,43 @@ def check_hyper_resistivity(**kwargs):
 
 
 def check_clustering(**kwargs):
-    valid_keys = ["berger", "tile"]
+    valid_methods = ["berger", "tile"]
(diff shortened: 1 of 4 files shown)


nicolasaunai (approved):

pyphare/pyphare/pharein/simulation.py
```diff
@@ -646,12 +646,43 @@ def check_hyper_resistivity(**kwargs):
 
 
 def check_clustering(**kwargs):
-    valid_keys = ["berger", "tile"]
+    valid_methods = ["berger", "tile"]
```
PhilipDeegan, line 644: can we put something like

```
"""
valid examples include
  "clustering"="tile"
or
  "clustering"={"method":"tile", "tile_size":8}
"""
```

or something do that affect

PhilipDeegan (approved):

nicolasaunai (approved):

Issue #1184, opened 2026-04-05
[Feature] Adaptive time-stepping for the MHD solver


igirault:
## Summary

Currently the time step `dt` is fixed for the entire duration of a simulation: it is set once in `pharein.Simulation` (via `time_step`, `time_step_nbr`, or `final_time`) and passed unchanged to the C++ simulator throughout. For MHD this is inconvenient: the most constraining stability condition depends on the physical state and can vary significantly over time, forcing users to pick a conservatively small, constant `dt`. An adaptive time-stepping mechanism would automatically select the largest `dt` that keeps every active stability condition satisfied at each iteration.

The relevant stability conditions are the multi-dimensional advective CFL, the fast magnetosonic CFL, and — when the Hall term is active — the Whistler wave CFL (which scales as `dx²` and is typically the most restrictive).

## Current code path

- `pyphare/pyphare/pharein/simulation.py` — `check_time()` validates and fixes `dt` at configuration time.
- `pyphare/pyphare/pharein/initialize/general.py` — serialises `dt` into the C++ dictionary once.
- `src/simulator/simulator.hpp` — stores `dt_` as a constant member; `timeStep()` simply returns it.
- `src/amr/multiphysics_integrator.hpp` — `getLevelDt()` is currently a passthrough that returns `dtTime` unchanged. This is the natural hook point.

## Proposed approach

1. **New CFL computation utility** — add a free function (e.g. in `src/core/numerics/` or `src/amr/`) that, given a patch or level, computes the local CFL-limited `dt` by iterating over patches and reducing to a global minimum via MPI. The Whistler condition is only evaluated when the Hall term is active.

2. **Hook into `getLevelDt()`** (`src/amr/multiphysics_integrator.hpp:444`) — replace the passthrough with a call to the CFL utility and return `min(dtTime, dt_cfl)`.

3. **Safety factor** — expose a `cfl_number` parameter in `pharein.Simulation` (default `0.9`) that scales the raw CFL-limited `dt`.

4. **Python-side relaxation** — when `time_step` is not explicitly provided by the user, infer an initial estimate from the configured grid and physical parameters, then let the solver adjust from there. If `time_step` is still supplied it acts as an upper bound.

5. **Diagnostics** — optionally record the chosen `dt` at each iteration (e.g. in the HDF5 output or a lightweight text log) to aid post-hoc analysis.

## Out of scope (for this issue)

- Per-level subcycling (each AMR level advancing with its own `dt`).
[...]

Issue #475, opened 2021-03-18
Variable dt implementations


PhilipDeegan:
If/when PHARE is to support a variable dt, here's somethings that could help

in: src/core/utilities/timestamps.h
``` // untested, but hopefully does the what you want
class VariableTimeStamper : public ITimeStamper
{
public:
    VariableTimeStamper(double const& dt, double const& init_time = 0)
        : dt_{dt}
        , last_change_{init_time}
        , last_time_(init_time)
    {
    }

    double operator+=(double const& new_dt) noexcept override
    {
        assert(new_dt > 0);

        if (new_dt != dt_) // not sure if safe, possibly
        {
            dt_          = new_dt;
            last_change_ = last_time_;
            n_same_      = 0;
        }
        return (last_time_ = last_change_ + (dt_ * ++n_same_));
    }

private:
    std::size_t n_same_ = 0;
    double dt_          = 0;
    double last_change_ = 0;
    double last_time_   = 0;
};
```

and update TimeStamperFactory::create like so (or similar if modified since)

```
    static std::unique_ptr<ITimeStamper> create(initializer::PHAREDict const& dict)
    {
        assert(dict.contains("time_step"));
        auto time_step = dict["time_step"].template to<double>();

        assert(dict.contains("time_stamper"));
        auto type      = dict["time_stamper"].template to<std::string>();

        if (type == "constant")
            return std::make_unique<ConstantTimeStamper>(time_step);

        if (type == "variable")
            return std::make_unique<VariableTimeStamper>(time_step);

        throw std::runtime_error("No TimeStamper exists for key: " + type);
    }
```

PhilipDeegan:
Gemini suggests some variants with kahan (compensated) summation

```C++
class VariableTimeStamper : public ITimeStamper
{
public:
    VariableTimeStamper(double const& dt, double const& init_time = 0)
        : dt_{dt}
        , last_time_(init_time)
        , error_compensation_(0.0)
    {
    }

    double operator+=(double const& new_dt) noexcept override
    {
        assert(new_dt > 0);
        dt_ = new_dt;

        // Kahan Summation Algorithm
        // 1. Subtract the accumulated floating-point error from the new increment
        double y = dt_ - error_compensation_;
        
        // 2. Add the corrected increment to the running total. 
        // High-order bits are updated; low-order bits of 'y' may be lost here.
        double temp = last_time_ + y;
        
        // 3. Recover the exact low-order bits that were dropped in the addition
[...]

PR #1223, opened 2026-06-18
keep dt exact

UCaromel:


PhilipDeegan:
This will eventually have to be replaced to support adaptive timestepping 

https://github.com/PHAREHUB/PHARE/issues/475
https://github.com/PHAREHUB/PHARE/issues/1184

pyphare/pyphare/pharein/diagnostics.py
```diff
@@ -6,8 +6,9 @@
 
 def all_timestamps(sim):
     init_time = sim.start_time()
-    nbr_dump_step = int((sim.final_time - init_time) / sim.time_step) + 1
-    return (sim.time_step * np.arange(nbr_dump_step)) + init_time
+    # consume the authoritative integer step count rather than re-deriving it from
```
nicolasaunai, line 9: I find this comment hard to understand when not looking at the current PR diff.

nicolasaunai (commented):
so @UCaromel @PhilipDeegan what do we do with this PR
should we merge and postpone the re-thinking about all this when adaptive time stepping is considered?

PhilipDeegan (approved):
I think it fixes a problem and should be merged in the mean time

there's many things to consider for adaptive timestepping and so much of this is likely to change

PR #1228, opened 2026-06-24
redo finer schedules on regrid

UCaromel:
Not technically needed with current master i think, as our refinement operators don't reach deep into the ghosts, but this is preparation for the higher order refinement operators. Basically, when refinement needs a stencil that reaches outside the box of the of the next coarser level, it needs data interpolated from the level below. On regrid, if the schedules are not remade for all levels higher than that the regridded level, this intermediary interpolation is going to happen with the old level, which is the wrong data and fail samrai's time check assertion.

### DIFF
--- a/src/amr/multiphysics_integrator.hpp
+++ b/src/amr/multiphysics_integrator.hpp
@@ -339,12 +339,17 @@ namespace solver
             else if (isRegridding)
             {
                 // regriding the current level has broken schedules for which
-                // this level is the source or destination
-                // we therefore need to rebuild them
+                // this level is the source or destination. A regridded level is
+                // the coarse source not only for its immediate child but for
+                // every finer level whose interlevel ghost fill reaches into
+                // it through a recursive coarse-interp chain. This happens for example when
+                // a finer level refinement stencil reaches outside the box of the immediate next
+                // coarser: SAMRAI builds a fill chain that reaches past the immediate parent into a
+                // grandparent (or deeper). Re-register all finer levels so none keep a stale
+                // shared_ptr to the pre-regrid PatchData of this level.
                 auto const finestLvlNbr = hierarchy->getFinestLevelNumber();
-                auto nextFiner = (levelNumber == finestLvlNbr) ? levelNumber : levelNumber + 1;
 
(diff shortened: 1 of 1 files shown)


PhilipDeegan:
is this related to nesting_buffer=0?

nicolasaunai:
> is this related to nesting_buffer=0?

I'd say it is related to `nesting_buffer` in general, but during regrids I'm not so sure.
@UCaromel 1/ what is your thought about that? 2/ are you doing that at the moment for high order, like on some branch?

nicolasaunai:
Also, I think schedule creation is an expensive thing we might want to consider that *only if really necessary*, maybe we should post here a timing difference with/without for a typical functional test.

UCaromel:
I think my initial framing is not fully accurate. The issue is that sometimes ghosts need a grandparent coarse level. If that grandparent level gets regridded without remaking the finer schedules, grandchildren finer levels when needing ghosts that reach outside what the immediate next coarser (parent) can provide will refine grandparent data **from before the regrid**, which is wrong of course. The fact that we mainly ran the code with 3 levels means we never ran into that problem, since we dont regrid L0. This can happen at 4 levels+, for 3 levels, that change is a no-op. we could instead optimize knowing the max number of cells needed for the refinement of **nghosts at the specified spatial refinement operator** -> which could give us the max level depth a fine level could need for its ghosts.

nicolasaunai:
> I think my initial framing is not fully accurate. The issue is that sometimes ghosts need a grandparent coarse level. If that grandparent level gets regridded without remaking the finer schedules, grandchildren finer levels when needing ghosts that reach outside what the immediate next coarser (parent) can provide will refine grandparent data **from before the regrid**, which is wrong of course. The fact that we mainly ran the code with 3 levels means we never ran into that problem, since we dont regrid L0. 

we do regrid L0 on hybrid 

>This can happen at 4 levels+, for 3 levels, that change is a no-op. we could instead optimize knowing the max number of cells needed for the refinement of **nghosts at the specified spatial refinement operator** -> which could give us the max level depth a fine level could need for its ghosts.

ok this is not likely to be an optimization that is useful unless we start running much deeper hierarchies.
But I would write this remark as a comment in the code, no?

PR #1230, opened 2026-06-28
less ghost box iteration

PhilipDeegan:
loop on index instead - do mhd nan check in debug

[nightly artifacts](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Caromel_Build/94400?buildTab=artifacts&focusLine=NaN)


nicolasaunai (approved):
LGTM but @UCaromel should review

igirault (approved):
This seems handy and quite less verbose

UCaromel (approved):

PR #1233, opened 2026-06-29
refactor(mhd): demote Resistivity/HyperResistivity/TimeIntegrator from compile-time to runtime

igirault:
## Summary

Demotes three MHD **compile-time** template parameters to **runtime** selection

1. **`Resistivity` / `HyperResistivity` → runtime `bool`s.** They become runtime params initialized from the value of `eta`/`nu`. No runtime if introduced in compute loops.
2 **`TimeIntegratorType` → runtime selection** Virtual dispatch replaces the integrator selection: `TimeIntegrator` holds a `std::unique_ptr<BaseMHDTimestepper>` chosen by a factory; RK intermediate states move into the base as `extra_states_` exposed via `getRunTimeResourcesViewList`. A generic mechanism is introduced for `enum` from/to `string` conversions.


PhilipDeegan:
Lots of potentially unrelated TimeStepper stuff in the diff

igirault:
I think it is on purpose, to select the time integration scheme at runtime. But maybe this is not done the way you would expect, we can discuss it if necessary

src/core/numerics/MHD_equations/MHD_equations.hpp
```diff
@@ -7,13 +7,11 @@
...
+template<bool Hall>
 class MHDEquations
 {
 public:
-    constexpr static bool hall             = Hall;
-    constexpr static bool resistivity      = Resistivity;
-    constexpr static bool hyperResistivity = HyperResistivity;
+    constexpr static bool hall = Hall;
 
     MHDEquations(double const gamma, double const eta, double const nu)
```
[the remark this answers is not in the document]
igirault: Resistive fluxes are not applied through `MHDEquations::compute(u, J)` — that commented branch was already dead on master, before this PR. The runtime path is elsewhere by design: `godunov_fluxes.hpp` applies `resistive_contributions(eta, ...)` and the hyper-resistive term post-Riemann on transverse-averaged fields (guarded by the runtime `resistivity_`/`hyper_resistivity_` flags introduced here), and constrained transport adds `E += eta*J` / the Laplacian term itself. So runtime `eta`/`nu` selection works; nothing regressed.

Cleaned up in c9c39f34: deleted the dead commented blocks and the now-unused `eta_`/`nu_` members from `MHDEquations` so the class no longer suggests it owns resistivity.

src/amr/solvers/time_integrator/base_mhd_timestepper.hpp
```diff
@@ -4,17 +4,32 @@
...
 {
 template<typename MHDModel>
+class MHDMessenger;
+} // namespace PHARE::amr
+
+namespace PHARE::solver
+{
+template<typename MHDModel, typename MessengerT = amr::MHDMessenger<MHDModel>>
 class BaseMHDTimestepper
 {
```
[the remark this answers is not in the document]
igirault: Done in c9c39f34: `MessengerT` now threaded through `EulerIntegrator`, `TVDRK2Integrator`, `TVDRK3Integrator`, `SSPRK4_5Integrator`, and `TimeIntegrator` (defaulted to `amr::MHDMessenger<MHDModel>`). `SolverMHD`'s `Messenger` now defaults to `TimeIntegratorStrategy::Messenger` with a `static_assert` guarding explicit mismatches.

src/amr/solvers/time_integrator/time_integrator.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "amr/solvers/time_integrator/tvdrk2_integrator.hpp"
+#include "amr/solvers/time_integrator/tvdrk3_integrator.hpp"
+#include "amr/solvers/time_integrator/ssprk4_5_integrator.hpp"
+
+namespace PHARE::solver
+{
+
+namespace detail
+{
+inline MHDOpts::TimeIntegratorType parse_time_integrator_type(std::string s)
```
igirault: to define along with MHDOpts::TimeIntegratorType

src/amr/solvers/time_integrator/ssprk4_5_integrator.hpp
```diff
@@ -38,125 +36,91 @@ class SSPRK4_5Integrator : public BaseMHDTimestepper<MHDModel>
...
     {
         Super::registerResources(model);
-        model.resourcesManager->registerResources(state1_);
-        model.resourcesManager->registerResources(state2_);
-        model.resourcesManager->registerResources(state3_);
-        model.resourcesManager->registerResources(state4_);
         euler_.registerResources(model);
-        // probably we should have the same resources for euler and compute_fluxes
-        // compute_fluxes_.registerResources(model);
+        compute_fluxes_.registerResources(model);
```
igirault: this is maybe redundant to register compute_fluxes here. To check also in other implementations

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -36,7 +36,7 @@ constexpr auto getDirections()
     }
 }
 
-template<auto direction, size_t dim, bool HyperResistivity>
+template<auto direction, size_t dim>
 auto getGrow(int const nghosts)
```
igirault: rajouter un booléen pour remplacer HyperResistivity

UCaromel: could be a class method

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -135,7 +133,7 @@ class Godunov : public GodunovInfo
                                                           riemann_.uct_coefs, {indices...});
 
                         // for energy ExB term
-                        if constexpr (Resistivity || HyperResistivity)
+                        if (resistivity_ || hyper_resistivity_)
```
igirault: if dans une boucle

res/sim/all.txt
```diff
@@ -30,32 +30,22 @@
...
-3,1,6,SSPRK4_5,WENOZ,None,Rusanov,false,false,false
-3,1,6,SSPRK4_5,MP5,None,Rusanov,false,false,false
-# --- coverage-only: gaps not hit by any test ---
-1,1,2,Euler,Linear,MinMod,Rusanov,false,false,false
-1,1,2,Euler,Constant,None,HLL,false,false,false
-1,1,2,Euler,Constant,None,HLLD,false,false,false
+# MHD permutations: dim,interp,refined,reconstruction,limiter,riemann,hall
+# (time integrator, resistivity and hyper-resistivity are now runtime selections,
+#  no longer part of the compile-time permutation)
+2,1,4,WENOZ,None,Rusanov,true
```
nicolasaunai, line 64: is there something that checks that these combinations are complete?

src/amr/solvers/time_integrator/compute_fluxes.hpp
```diff
@@ -43,15 +42,19 @@ class ComputeFluxes
         , constrainedTransportInfo_{ConstrainedTransportInfo_t::FROM(dict["constrained_transport"])}
         , to_primitive_gamma_{dict["to_primitive"]["heat_capacity_ratio"]}
         , to_conservative_gamma_{dict["to_conservative"]["heat_capacity_ratio"]}
+        , needsCurrent_{Hall || fVMethodInfo_.resistive() || fVMethodInfo_.hyperResistive()}
     {
+        // eta/nu are duplicated into both info structs
+        assert(fVMethodInfo_.eta == constrainedTransportInfo_.eta
```
nicolasaunai, line 52: the duplication is weird...

src/amr/solvers/time_integrator/compute_fluxes.hpp
```diff
@@ -43,15 +42,19 @@ class ComputeFluxes
...
+               && fVMethodInfo_.nu == constrainedTransportInfo_.nu);
     }
 
     void operator()(MHDModel& model, auto& state, auto& fluxes, auto& bc, level_t& level,
                     double const newTime)
     {
         ToPrimitiveConverter_t{level, model}(state, to_primitive_gamma_, newTime);
 
-        if constexpr (Hall || Resistivity || HyperResistivity)
+        if (needsCurrent_)
```
nicolasaunai, line 67: I don't get why the if constexpr Hall has disappeared for a runtime bool

src/core/numerics/constrained_transport/constrained_transport.hpp
```diff
@@ -25,6 +25,8 @@ class ConstrainedTransport
         , eta_{eta}
         , nu_{nu}
         , hyper_mode_{hyper_mode}
+        , resistivity_{eta != 0.0}
```
nicolasaunai: `eta>0` probably better

src/core/numerics/ohm/ohm.hpp
```diff
@@ -20,6 +20,9 @@ struct OhmInfo
     double const nu;
     HyperMode const hyper_mode;
 
+    bool resistive() const { return eta != 0.0; }
```
nicolasaunai: `var>0`

src/core/numerics/godunov_fluxes/godunov_utils.hpp
```diff
@@ -320,78 +297,102 @@ struct AllFluxesNames
...
-    constexpr static auto HyperResistivity = Equations::hyperResistivity;
+    using Field                     = VecField::field_type;
+    constexpr static auto dimension = VecField::dimension;
 
 public:
     GodunovState() = default;
 
+    // resistivity / hyper-resistivity are now runtime selections, so the transverse magnetic
+    // field storage is always allocated.
     NO_DISCARD auto getCompileTimeResourcesViewList()
```
igirault: pass transverse fields as runtime resources to prevent allocation when theyr are not needed (ri ideal mhd)

src/core/numerics/ohm/ohm.hpp
```diff
@@ -20,6 +20,9 @@ struct OhmInfo
...
+    bool hyperResistive() const { return nu != 0.0; }
```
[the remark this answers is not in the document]
igirault: yes + boolean names should be isResistive or something like that + nu/eta >0

src/core/numerics/constrained_transport/constrained_transport.hpp
```diff
@@ -13,7 +13,7 @@
 namespace PHARE::core
```
igirault, line 12: fichier à dégager

igirault: dans un autre PR

res/sim/all.txt
PhilipDeegan: each valid line constructs https://github.com/PHAREHUB/PHARE/blob/master/src/phare_simulator_options.hpp#L21

which supports defaulted values, so it would only fail if you used invalid values, or too many

src/core/numerics/constrained_transport/upwind_constrained_transport.hpp
```diff
@@ -31,6 +30,8 @@ class UpwindConstrainedTransport : UpwindConstrainedTransportInfo
     UpwindConstrainedTransport(UpwindConstrainedTransportInfo const& info, GridLayout const& layout)
         : Super{info}
         , layout_{layout}
+        , resistivity_{info.resistive()}
```
igirault: booléen pas explicite, toujours avoir un nom type is...

src/core/numerics/constrained_transport/upwind_constrained_transport.hpp
```diff
@@ -47,15 +48,15 @@ class UpwindConstrainedTransport : UpwindConstrainedTransportInfo
         layout_.evalOnBox(Ey, [&](auto&... args) mutable { EyEq_(ct_state, Ey, B, {args...}); });
```
igirault: mutable pas nécessaire

src/core/numerics/constrained_transport/upwind_constrained_transport.hpp
```diff
@@ -47,15 +48,15 @@ class UpwindConstrainedTransport : UpwindConstrainedTransportInfo
...
 
             auto& Jx = J(Component::X);
             auto& Jy = J(Component::Y);
             auto& Jz = J(Component::Z);
 
-            if constexpr (Resistivity)
+            if (resistivity_)
             {
                 layout_.evalOnBox(
                     Ex, [&](auto&... args) mutable { resistive_contribution_(Ex, Jx, {args...}); });
```
igirault: mutable toujours pas nécessaire

src/core/numerics/constrained_transport/upwind_constrained_transport_utils.hpp
```diff
@@ -23,29 +23,18 @@ class UpwindConstrainedTransportState
     {
         if constexpr (dimension == 1)
```
igirault, line 67: passer toutes ces ressources en runtime

src/python3/mhd_resolver.hpp
```diff
@@ -216,7 +183,11 @@ struct MHDResolver
     using FVMethodStrategy
         = core::Godunov<GridLayout, Reconstruction_t, RiemannSolver_t, Equations_t>;
 
-    using MHDTimeStepper_t = MHDTimeStepper<FVMethodStrategy>;
+    static constexpr bool is_mhd
+        = (opts.reconstruction_type != MHDOpts::ReconstructionType::Default);
```
igirault: Pour un autre PR: c'est pas top

src/amr/solvers/time_integrator/base_mhd_timestepper.hpp
```diff
@@ -35,21 +48,39 @@ class BaseMHDTimestepper
                          {"timeEtot_fz", core::MHDQuantity::Scalar::ScalarFlux_z}}
         , butcherE_{"timeE", core::MHDQuantity::Vector::E}
     {
+        for (std::size_t i = 0; i < n_extra_states; ++i)
+            extra_states_.emplace_back("state" + std::to_string(i + 1));
     }
```
igirault, line 38: autre PR: refactor les allocations ici et dans solver_mhd

pyphare/pyphare/pharein/initialize/mhd.py
```diff
@@ -15,6 +15,7 @@ def populateDict(sim):
     add_double("simulation/algo/fv_method/hyper_resistivity", sim.nu)
     add_double("simulation/algo/fv_method/heat_capacity_ratio", sim.gamma)
     add_string("simulation/algo/fv_method/hyper_mode", sim.hyper_mode)
+    add_string("simulation/algo/time_integrator_type", sim.mhd_timestepper)
```
igirault: dégager dans simulation.py les booléens res et hyper-res

doc/source/_static/css/custom.css
PhilipDeegan, line 1: I don't see how this is related to the PR

you might want to rebase master and pop things that are not relevant

PhilipDeegan, line 6: generally text files should end with an empty line to be POSIX compliant

PhilipDeegan:
it's not clear to me how the time integrator changes related to the resistivity/hyper-resistivity changes

could be two PRs?

igirault:
Two changes are independent, could be split indeed

src/core/numerics/constrained_transport/upwind_constrained_transport_utils.hpp
```diff
@@ -2,51 +2,79 @@
...
+                return std::forward_as_tuple(jt_x, rhot_x);
+            else if constexpr (dimension == 2)
+                return std::forward_as_tuple(jt_x, rhot_x, jt_y, rhot_y);
             else
-                return std::forward_as_tuple(vt_x, aL_x, aR_x, dL_x, dR_x, vt_y, aL_y, aR_y, dL_y,
-                                             dR_y);
+                return std::forward_as_tuple(jt_x, rhot_x, jt_y, rhot_y, jt_z, rhot_z);
         }
+
+        VecField jt_x{"j_t_x", MHDQuantity::Vector::VecFlux_x};
```
UCaromel, line 46: Explaining what these transverse field/vecfields are

src/core/numerics/constrained_transport/upwind_constrained_transport_utils.hpp
```diff
@@ -185,6 +185,9 @@ class UpwindConstrainedTransportState
         aR_z{"aR_z", MHDQuantity::Scalar::ScalarFlux_z},
         dL_z{"dL_z", MHDQuantity::Scalar::ScalarFlux_z},
         dR_z{"dR_z", MHDQuantity::Scalar::ScalarFlux_z};
+
+private:
+    std::vector<TransverseResistiveState> resistive_;
```
UCaromel: Add a comment: this is a vector because it has to be because of the current implementation oc getCompleTimeRessourceViewList, it will never have more than one element.

UCaromel: Add the name is not that good

src/core/numerics/godunov_fluxes/godunov_utils.hpp
```diff
@@ -324,47 +323,43 @@ struct AllFluxesNames
...
+        // resource manager sets buffers on these members and not on temporaries.
+        if (isResistive || isHyperResistive)
         {
-            if constexpr (dimension == 1)
-                return std::forward_as_tuple(bt_x);
-            else if constexpr (dimension == 2)
-                return std::forward_as_tuple(bt_x, bt_y);
-            else if constexpr (dimension == 3)
-                return std::forward_as_tuple(bt_x, bt_y, bt_z);
+            bt_.emplace_back("b_t_x", MHDQuantity::Vector::VecFlux_x);
```
UCaromel: bt reserve dim

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -203,14 +224,30 @@ class Godunov : public GodunovInfo
...
-                                throw std::runtime_error("Error - Ohm - unknown hyper_mode");
                         }
                     });
-            }
+            };
+
+            // loop at compile time to avoid runtime checks in compute loops
+            std::visit(
+                [&](auto isResistiveTag, auto isHyperResistiveTag, auto hyperModeTag) {
+                    constexpr bool isResistive      = decltype(isResistiveTag)::value;
```
PhilipDeegan: we could probably do with some helper like [this](https://coliru.stacked-crooked.com/a/8bf60eb6f2267874)

src/core/numerics/constrained_transport/upwind_constrained_transport_utils.hpp
igirault: Would actually need you for that :smiley:

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
igirault: Very nice, much simpler to use. Here is an extension by Claude that also accepts enums https://coliru.stacked-crooked.com/a/969b5615acb3bdd4 It builds on the introduced trait for enum-string conversions.

src/core/utilities/meta/meta_utilities.hpp
```diff
@@ -79,6 +81,95 @@ namespace core
     }
 
 
+    namespace detail
+    {
+        // bool -> asBoolConstant; scoped enum -> asEnumConstant (needs an EnumTraits<T>
+        // specialization, see enum.hpp). Anything else is a programming error at the call site.
+        template<typename T>
+        auto toConstexprVariant(T const& value)
```
PhilipDeegan: I would think this should be two functions

```
auto toConstexprVariant(bool const value){}

template<typename T>
requires(std::is_enum_v<T>)
auto toConstexprVariant(T const& value){}
```

this way, other implementations could exist as needed without conflicts

PhilipDeegan:
A possibly simpler approach than needing all this enum to/from string magic is to [expose the enum to python](https://github.com/PhilipDeegan/PHARE/blob/mkn/src/python3/cpp_etc.cpp#L118-L123), pretty sure you get it for free there, so you could convert the python string, to a direct C++ runtime enum value

not sure it helps converting it to constexpr tho, not sure exactly you need any strings for that either, probably just ints

src/core/utilities/meta/meta_utilities.hpp
igirault: In this case, should `toConstexprVariant` overloads replace the definitions of `toBoolConstant`/`toEnumConstant` ? Or do we keep these last two and declare overloads in `meta_utilities.hpp` as 

```cpp
    namespace detail
    {
        template<typename Bool>
        auto toConstexprVariant(Bool const& value)
            requires(std::is_same_v<Bool, bool>)
        {
            return asBoolConstant(value);
        }

        template<typename Enum>
        auto toConstexprVariant(Enum const& value)
            requires(std::is_enum_v<Enum>)
        {
            return asEnumConstant(value);
        }

        template<typename T>
        constexpr bool always_false_v = false;

        template<typename T>
        auto toConstexprVariant(T const&)
        {
            static_assert(always_false_v<T>,
                          "Constexprifier only supports bool and scoped enum types");
        }
    } // namespace detail

```

PhilipDeegan:
```
        template<typename Bool>
        auto toConstexprVariant(Bool const& value)
            requires(std::is_same_v<Bool, bool>)
```

I don't know why you would do this when you can just let the type system work and set the parameter as a bool

```
template<typename T>
constexpr bool always_false_v = false;

template<typename T>
auto toConstexprVariant(T const&)
{
    static_assert(always_false_v<T>,
                  "Constexprifier only supports bool and scoped enum types");
}
```

this reintroduces the problem which I said we could avoid

igirault: this function
```inline auto toConstexprVariant(bool const value) { return asBoolConstant(value); }```
will accept integers and converts them to bools, this was to avoid that

ok for the second point

Comment (PhilipDeegan) [reply src/core/utilities/meta/meta_utilities.hpp]: ok I see, that's annoying, we typically expect Werrors for `narrowing`, but this one doesn't seem to count

**Comment by igirault** [comment]
Yup but this does not spare the manual declaration of string names, as your example shows. And not a big deal, but you loose the ability to print meaningful names on the cpp side. 

For constexprification, it it enough to assume the enum includes a count member at its end to define a generic `toEnumConstant`.

On this, I would let you decide

**Comment by PhilipDeegan** [comment]
> For constexprification, it it enough to assume the enum includes a count member at its end to define a generic `toEnumConstant`.

This isn't great either as the count will increase the total set of possible parameter permutations

**Comment by igirault** [comment]
So what do we do regarding enum <-> string capability ? Otherwise I addressed all review comments

**Comment by PhilipDeegan** [comment]
> So what do we do regarding enum <-> string capability ? Otherwise I addressed all review comments

Enums are not strings, they are ints, so you should be passing integers from python to C++

[like](https://github.com/PhilipDeegan/PHARE/blob/97b73e61fad5211fbb743a016d908b4865123a59/pyphare/pyphare/pharein/initialize/general.py#L89-L93)

**Comment by igirault** [comment]
Okay I did that and removed meta/enum.hpp stuff. I made Constexprifier stuff rely on a 'count' member for enums

File: src/initializer/data_provider.hpp [hunk 27]
```diff
@@ -6,12 +6,31 @@
...
+ */
+template<typename Enum, typename... Types>
+    requires(std::is_enum_v<Enum>)
+Enum get_value(Dict<Types...> const& dict, std::string const& path, Enum const default_value,
+               char delimiter = '/')
+{
+    return static_cast<Enum>(
+        get_value(dict, detail::split_string(path, delimiter), static_cast<int>(default_value)));
+}
+} // namespace cppdict
```
igirault, line 31: maybe should not be defined like that

src/core/numerics/constrained_transport/upwind_constrained_transport_utils.hpp
PhilipDeegan: todo?

igirault: under the control of Master @UCaromel

nicolasaunai (approved):

Issue #842, opened 2024-05-24
tune tagging mechanism to reduce noisy refinement


nicolasaunai:
At the moment the refinement is triggered a bit easily and creates spurious patches here and there especially from L0 to L1.
This issue is about adjusting a bit the threshold so that it does less that.
Let the tagging the threshold accessible from python (see Nathan's work) so that study can be done easily without recompiling (and useful for end user as well).
Assess the effect of increasing the threshold in various simulations of Harris type.
Assess the effect of less noisy patch creation on the performance and scalability.

nicolasaunai:
this might be linked to #824. We may not be able to reduce the number of calls to regrid but we may not have to explicitly regrid (i.e. change the patch distrib) as often if there was not as much noise in the tags.
Perhaps the high frequency evolution of the patch distribution comes from noise in tags. Tags would, on average, be at the same location, but slightly different, which results in computing slightly different boxes and then effectively regriding.

PR #1238, opened 2026-07-04
tagging: generic field selection, lohner + wavelet (MR) criteria

igirault:
## Extends AMR tagging 
- per-quantity thresholds
- generic field selection
- two new tagging criteria, [Lohner](https://www.sciencedirect.com/science/article/pii/0045782587900983) and [Wavelet-based](https://linkinghub.elsevier.com/retrieve/pii/S0045793018308405). Wavelet-based mostly untested in real conditions and added for fun, can be removed from PR if necessary.

## Modifcations of pre-existing stuff
- legacy tagging criterion has been centered

## Usage from python
```
refinement="tagging",
tagging={
    "method": "lohner",
    "quantities": {"B": 0.1, "rho": 0.4},
    "params": {"reltol": 0.02}
}
```


pyphare/pyphare/pharein/simulation.py
```diff
@@ -574,6 +574,91 @@ def check_refinement(**kwargs):
...
+      the denominator of the estimator (damps the indicator where the field
+      variation is small relative to its magnitude); `abstol` (default 1e-30) is
+      an absolute denominator floor.
+    - wavelet: multiresolution detail criterion (Domingues et al. 2019,
+      10.1016/j.compfluid.2019.06.025). The indicator is the ABSOLUTE prediction
+      error |Q - Qpredicted| (units of Q), so thresholds are not comparable with
+      default/lohner ones. By default the threshold follows Harten's level
+      scaling eps_l = eps / |Omega| * 2^{dim (l - L)} (|Omega| the physical
+      domain volume, L the finest level), refining coarse levels more eagerly;
+      `level_scaling` (default 1) set to 0 uses the per-quantity threshold
```
igirault: level_scaling should be a bool instead of 0/1

nicolasaunai:
related to #842

src/amr/tagging/tagging_criteria.hpp
```diff
@@ -0,0 +1,284 @@
...
+        // tensor product of the 1D weights w(0)=1, w(+/-1)=-/+1/8 (sign flipped for
+        // an odd child), over the 3^dim coarse neighbourhood.
+        int constexpr npts = [] {
+            int n = 1;
+            for (std::size_t d = 0; d < dim; ++d)
+                n *= 3;
+            return n;
+        }();
+
+        double pred = 0.0;
```
UCaromel, line 258: could maybe use the tensor product implementation we already have in https://github.com/PHAREHUB/PHARE/blob/master/src/core/data/grid/impl/yee/gridlayout_mhd_yee.hpp#L587-L636

igirault: I don't think so.

src/simulator/simulator.hpp
```diff
@@ -365,8 +365,12 @@ void Simulator<opts>::mhd_init(initializer::PHAREDict const& dict)
         if (dict["simulation"]["AMR"]["refinement"]["tagging"]["method"].template to<std::string>()
             != "none")
         {
-            auto mhdTagger_ = amr::TaggerFactory<MHDModel>::make(
-                dict["simulation"]["AMR"]["refinement"]["tagging"]);
+            // the wavelet Harten scaling reference L must be the finest level carrying MHD
+            // data (maxMHDLevel_ - 1), not the global finest (maxLevelNumber_ - 1) which in a
```
igirault, line 369: I do not know actually. Makes no difference for a pure MHD run

nicolasaunai (commented):
I do not understand why this PR deletes the existing architecture with the tagging strategy and factory which was there precisely to be able to add new criteria without touching existing ones.

src/amr/tagging/concrete_tagger.hpp
```diff
@@ -1,26 +1,290 @@
...
+    };
+
+    // Translates a common quantity word into the literal field-name token to search for
+    // with the generic resource-tree resolver, per model kind. Any string that isn't one of
+    // the four common words is returned unchanged (raw/legacy field names, e.g. "rho",
+    // "bulkVel", "Bx", keep working exactly as before). Throws for a quantity that is a
+    // recognized common word but has no equivalent field on that model kind yet.
+    inline std::string resolveCommonQuantityAlias(TaggerModelKind kind, std::string const& qty)
+    {
+        if (kind == TaggerModelKind::Hybrid)
```
nicolasaunai, line 70: I don't like that, before we could have had tagging with about any quantity we'd like, including several, it was all contained into the specific strategy. This hard-codes using one only

nicolasaunai:
This PR seems to do several unrelated things:
- it seems that it changes the default/legacy tagging formula
- it changes the way diag tags are computed 
- and it introduces new tagging criteria, re-working the code architecture.

I feel the PR should just be about introducing the new criteria, as necessary, and demonstrate usefulness.
Also, I think it should be a really minimal diff, as far as I understand, it should just be about adding a new tagging strategy and that's all.

igirault:
> it changes the way diag tags are computed

Yes I did this because tagging diag was only available in 2D

> This PR seems to do several unrelated things

Yes indeed, I would even add the "per-quantity tagging" to the list

So I am going to rewrite this PR to focus on the two things we initially wanted:
- per-quantity tagging
- Lohner tagging criterion

nicolasaunai:
Hold on before you do. 
we do not want a per quantity, I think the abstraction on master is fine.
It’s just about adding a new strat

igirault:
> Hold on before you do. we do not want a per quantity

I think we do: pure hydro case cannot rely on B for tagging

PR #1240, opened 2026-07-06
Derived-quantity diagnostics: generic interface (V, P, divB, J, E) + catalogue-based writer dispatch

igirault:
## What this PR does

Diagnostics can now output *derived* quantities — fields computed on the fly at dump time from the primary variables — through one generic interface, instead of hand-wiring each new output through both writer stacks. On top of that interface, the quantity dispatch of the writers themselves is consolidated into per-model catalogues, so adding a quantity no longer means editing ~6 dispatch sites.

New/changed outputs:

| Quantity | Tree | Models | Notes |
|---|---|---|---|
| `V` (velocity) | `/mhd/` | MHD | was hand-wired (`V_diag_`), now derived `rhoV/rho` |
| `P` (pressure) | `/mhd/` | MHD | was hand-wired (`P_diag_`), now derived via EOS |
| `EM_divB` | electromag | MHD + Hybrid | new; was python-only post-processing |
| `EM_J` | electromag | MHD + Hybrid | new; `curl B` via `core::Ampere`, recomputed fresh (not the solver's transient `state.J`) |
| `EM_E` | electromag | MHD | new; full Ohm's law reconstruction (see below). Hybrid keeps exposing its real, stored `E` |

Both writer stacks (phareh5 + vtkhdf) consume everything symmetrically. `V`/`P` paths, centerings and values are unchanged — backward compatible.

## The derived-quantity interface

- `DerivedQuantity<State, GridLayout, rank>` (rank 0 = scalar, rank 1 = vector): `name()`, `centering()`, `category()` (`fluid` | `electromag` — which diagnostic tree it publishes under), `compute(state, layout, out, time)`.
- Concrete implementations per model (`MhdVelocity`, `MhdPressure`, `MhdDivB`, `MhdCurrentDensity`, `MhdElectricField`, `HybridCurrentDensity`, `HybridDivB`), collected in a per-model `DerivedQuantityRegistry` built at `ModelView` construction.
- **Scratch is SAMRAI-managed**: each model registers one all-primal scalar + one all-primal vector buffer (the most demanding centering, so a view of any centering fits). Writers build centering-correct `Field`/`VecField` views over the backing buffer per patch; compute and write stay fused per patch. Views are zeroed before each compute — the buffer is reused across patches/quantities, and curl/laplacian stencils would otherwise leave stale cross-patch data in the ghost layers that end up in the file.
- Centering-only quantity aliases (`ScalarCellCentered`, `VecElike`, …) reuse existing enum values — no `GridLayout` centering-map changes.

The derived `EM_E` for MHD matches the model actually run: ideal `-V×B` + Hall `(J×B)/rho` + resistive `η·J` + hyper-resistive term. The Hall term is gated on the permutation's `hall` flag (threaded sim → `MHDState` → the computer), the hyper-resistive term carries the solver's `minMeshSize²` factor, and `V` is reconstructed from `rho`/`rhoV` projected per-component onto the E-edge (as in hybrid Ohm) rather than mixing cell-centered `V` with edge-projected `B`.

## Writer-dispatch catalogues

Previously both writer stacks hand-copied the same quantity if-chains across `createFiles` / `getDataSetInfo` / `initDataSets` / `write` (h5) and the initializer/writer functors (vtkhdf), and a derived quantity's tree membership was implicit in hardcoded `createFiles` name lists. Now `ModelView` is the single source of truth:

- `forEachFluidQuantity` / `forEachEmQuantity` — enumeration only (names/ranks), safe for null/remote patches and per-level setup;
- `visitActiveFluidQuantity` / `visitActiveEmQuantity` — per-patch dispatch; derived quantities are viewed in scratch and computed only when the consumer needs data (`compute_derived=false` for shape-only passes).

Under the hood, every quantity — primary or derived, fluid or EM tree — is registered exactly once at `ModelView` construction as a (`DiagQtyInfo` = tree/name/attribute-group, lazy `access`) catalogue entry: `access` returns a shallow view (a copy of the state field's view object for primaries; a centering-correct scratch view, computed iff `compute_derived`, for derived quantities). Enumeration and dispatch are implemented once in `BaseModelView` — dispatch resolves the active quantity string through one hash lookup, not per-patch catalogue scans — and the per-model ModelViews reduce to entry registration in their constructors. Both models expose the same catalogue API with a `layout/time/compute_derived` signature, so **both** writer stacks are fully model-generic: the electromag writers' `is_mhd`/`is_hybrid` `if constexpr` guards are gone, and the vtkhdf fluid writer's per-model initializer/writer functor pairs collapsed into a single `FluidInitializer`/`FluidWriter`. (MHD's tensor callback is never invoked — no rank-2 fluid quantities; vtkhdf passes a no-op tensor callback, preserving that stack's existing lack of rank-2 support.)

## Python side

- `ElectromagDiagnostics` accepts `EM_divB` / `EM_J` (+ `EM_E` for MHD) — divB is a new diagnostic (previously python-only post-processing) and lives in the electromag tree from the start.
- New getters `Run.GetMHDdivB` / `Run.GetMHDJ` / `Run.GetMHDE`; `quantities_per_file` updated accordingly.
- The divB patch-data key is the model-neutral `divB` (not `mhdDivB`) — it is an EM-tree quantity shared by both models.
- Minimal (e.g. ideal) MHD dicts still construct: `resistivity`/`hyper_resistivity` default to 0.

**Intended behavior change:** an invalid diagnostic quantity now fails loudly at *configuration* time: the `diagnostics_checker` decorator used to swallow validation `ValueError`s (printed, diagnostic silently unregistered); it now propagates, and a test pins the failure mode.

## Cleanup

Replaces and removes the `V_diag_`/`P_diag_` buffers, `MHDDiagnosticWriter::compute`, and `MhdFluidComputer` — the per-stack duplicated compute logic this interface obsoletes.

## Tests

- C++ unit: centering mapping, interface/registry, scratch-view aliasing, all MHD computers (incl. discrete `div(curl B) ≈ 0` identity, and per-component projection regressions for `E`'s Hall/hyper-resistive terms).
- Functional: `test_mhd_derived_diagnostics.py` (Orszag–Tang: `V == rhoV/rho` patchwise, `divB ≈ 0`, `P` sanity, `EM_divB`/`EM_J`/`EM_E` end to end via both writers, vtkhdf smoke); hybrid `EM_J` + `EM_divB` end to end; `test-diagnostics_1d` incl. `momentum_tensor`.

**Known, out of scope:** the MHD functional test's `V == rhoV/rho` comparison shows a pre-existing float32-ULP-scale difference (~3e-8 on ~19% of cells) in the phareh5 write/read roundtrip, present before this branch; its tolerance was loosened (`rtol=1e-6`, `atol=1e-7`) to unblock this PR's own checks. Root cause not investigated here.


src/diagnostic/diagnostic_model_view.hpp
```diff
@@ -157,19 +231,114 @@ class ModelView<Hierarchy, Model, std::enable_if_t<solver::is_hybrid_model_v<Mod
 public:
     using Model_t                = Model;
     using physical_quantity_type = Model::physical_quantity_type;
+    using State_t                = std::decay_t<decltype(std::declval<Model&>().state)>;
```
igirault, line 393: can't use something simpler Model_t::State_t ?

src/diagnostic/diagnostic_model_view.hpp
```diff
@@ -157,19 +231,114 @@ class ModelView<Hierarchy, Model, std::enable_if_t<solver::is_hybrid_model_v<Mod
...
+    using GridLayout             = typename Super::GridLayout;
```
igirault, line 394: remove typename

src/diagnostic/diagnostic_model_view.hpp
PhilipDeegan: it's not exported on the Model but you can add it as I have done [here](https://github.com/PhilipDeegan/PHARE/blob/mkn/src/amr/physical_models/hybrid_model.hpp#L71)

tests/simulator/test_mhd_derived_diagnostics.py
```diff
@@ -0,0 +1,274 @@
...
+        for ilvl in P.levels():
+            for patch in P.level(ilvl).patches:
+                pdata = interior(patch.patch_datas["mhdP"])
+                self.assertTrue(np.isfinite(pdata).all())
+                self.assertTrue((pdata > 0).all())
+                # Same pre-existing float32 write/read roundtrip precision
+                # loss as the V == rhoV / rho check above (observed max abs
+                # diff here ~7e-9, ~5.3e-8 relative, again ULP-scale for
+                # float32), hence the loosened tolerance.
+                np.testing.assert_allclose(pdata, p_init, rtol=1e-6, atol=1e-7)
```
PhilipDeegan, line 221: this is why we have [this](https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/core/phare_utilities.py#L144)

PR #1241, opened 2026-07-06
Inner boundary

igirault:


pyphare/pyphare/pharein/simulation.py
```diff
@@ -699,6 +699,124 @@ def check_mhd_parameters(**kwargs):
...
+    valid_condition_types = {"reflective", "ionospheric-convection"}
+    condition_type = inner_boundary.get("condition_type", "reflective")
+    if condition_type not in valid_condition_types:
+        raise ValueError(
+            f"Error: inner_boundary condition_type '{condition_type}' is invalid, "
+            f"valid types are {valid_condition_types}"
+        )
+
+    # per-type extra keys (prescribed reservoir values the user must provide)
+    condition_keys = {
```
PhilipDeegan, line 734: maybe something like we do for LoadBalancers could work here

https://github.com/PHAREHUB/PHARE/blob/master/pyphare/pyphare/pharein/load_balancer.py#L9

makes things more obvious about what variables exist or are required

PR #1255, opened 2026-07-19
Add CONTRIBUTING.md, issue templates, and PR template

nicolasaunai:
We currently do not have "rules" for how the code base should change.
This leads to possibly hectic PRs difficult to review about some undiscussed and undocumented features.
These files try to prevent this situation with a more formal approach.
The key is that small-single-issue-a priori-discussed-PRs are less likely to pile up as opposed to current ones.
Let's consider this PR as the last not having a prior issue :D

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -11,6 +11,21 @@ knowledge_base:
     project_keys: []
   linear:
     team_keys: []
+  code_guidelines:
+    enabled: true
+    filePatterns:
+      - '**/.cursorrules'
+      - .github/copilot-instructions.md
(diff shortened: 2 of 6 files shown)


nicolasaunai:
open for discussion

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
igirault, line 1: We can add something to redirect users to discussions or documentation when they would rush to open an issue, like

```
contact_links:
  - name: Question / discussion
    url: https://github.com/PHAREHUB/PHARE/discussions
    about: Ask usage questions, propose ideas, or discuss before opening an issue.
  - name: Documentation
    url: https://phare.readthedocs.io
    about: Check the docs before reporting.
```

igirault (approved):
No requirements for the submitted code : tests, formatting, code conventions, must compile, must not break CI... This is ok for us, but will need refinement for external contributors.

LGTM otherwise

UCaromel (approved):

PR #1263, opened 2026-07-22
make boxes overload spannable rather than iterable

### LINKED ISSUE #1003: isIn for boxcontainer is ambiguous
Since the addition of the box iterator, making a box iterable, 

the following function may lead to some inconsistencies (which I'm seeing on a branch atm)

https://github.com/PHAREHUB/PHARE/blob/83a9afbfa2cda038c970280fb1ab61cd5760ceb1/src/core/utilities/box/box.hpp#L205-L233

suggested refactor, rename to `isInAny`

PhilipDeegan:
a single box is iterable, a single box is not a span, boxes have no data.

closes #1003

### DIFF
--- a/src/core/utilities/box/box.hpp
+++ b/src/core/utilities/box/box.hpp
@@ -3,6 +3,7 @@
 
 
 #include "core/def.hpp"
+#include "core/utilities/span.hpp"
 #include "core/utilities/types.hpp"
 #include "core/utilities/point/point.hpp"
 #include "core/utilities/meta/meta_utilities.hpp"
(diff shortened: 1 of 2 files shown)


src/core/utilities/box/box.hpp
```diff
@@ -259,70 +260,27 @@ struct boxes_iterator
...
-        pointInBox = pointInBox && isIn1D(point[iDim], box.lower[iDim], box.upper[iDim]);
-    if (pointInBox)
-        return pointInBox;
+    for (std::size_t iDim = 0; iDim < SIZE; ++iDim)
+        if (point[iDim] < box.lower[iDim] || point[iDim] > box.upper[iDim])
+            return false;
+    return true;
+}
 
+template<Spannable Boxes>
```
nicolasaunai, line 273: just for my curiosity: in practice this function will not compile if the given type is *just* spannable, it also needs begin/end because of the range-based loop. How do we think of "concepts" in  general here? I thought Concepts were meant to guarantee the passed type is compatible with its usage in the function? If so, the concept to be used here should rather be something like Container (could be  == Spannable + Iterable?)

PhilipDeegan: it's true given the concept as is, the loop should be `for(std::size_t idx = 0....)` rather than a range based loop which requires begin/end.

The span concept probably should include begin/end, data and size, this is consistent with https://en.cppreference.com/cpp/container/span and would prevent confusion with things that are just iterable  (have begin/end, but no data or size)

the current `is_iterable` could probably be converted to a concept containing just being/end

PhilipDeegan: Spannable concept now includes begin/end consistent with std::span

nicolasaunai (approved):

Issue #1110, opened 2025-11-19
document all phare options


nicolasaunai:
We have many many options to configure a simulation, from the python simulation etc. scripts, but also in more hidden environment variables. We need to have a documentation page that gather them all.

PR #1268, opened 2026-07-22
document PHARE environment variable usage

PhilipDeegan:
relates to #1110 but maybe doesn't close

build is irrelevant


nicolasaunai:
Can this go to readthedoc?

PhilipDeegan:
> Can this go to readthedoc?

also or only?

nicolasaunai:
> > Can this go to readthedoc?
> 
> also or only?

Well, this is user documentation so it should be accessible from readthedoc.

doc/source/usage/environment_variables.rst
```diff
@@ -0,0 +1,68 @@
...
+Enables ``phlop`` scope/timing instrumentation. Truthy values: ``1`` or ``true``.
+
+Writes one timing file per rank to ``.phare/timings/rank.<N>.txt``. Read independently on both
+the C++ side (gated behind the ``phlop`` build option) and the Python side (which additionally
+triggers monitoring setup before the C++ simulator is constructed).
+
+``PHARE_H5_CHUNK_SIZE``
+-------------------------
+
+HDF5 dataset chunk size, in elements. Defaults to ``1024``. Affects diagnostic/restart HDF5 I/O.
```
PhilipDeegan: right now it only affects VTKHDF diagnostics, and I don't see how it affects restarts but claude says so 🤷‍♂️

doc/source/usage/environment_variables.rst
```diff
@@ -0,0 +1,68 @@
...
+
+``PHARE_SIM_MON``
+-------------------
+
+Enables periodic resource-usage monitoring during a simulation run. Truthy values: ``true``,
```
nicolasaunai, line 44: shouldn't we say something about where these measurements go and how to use them?

doc/source/usage/environment_variables.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Any other value is rejected. When running via ``pyphare``, the Python side defaults this to
+``RANK_FILES`` if unset (and creates the ``.log/`` directory on rank 0) before the C++ simulator
+starts. When running pure C++ (e.g. unit tests) without going through Python, no redirection
+happens unless this variable is set.
+
+``PHARE_SCOPE_TIMING``
+-----------------------
+
+Enables ``phlop`` scope/timing instrumentation. Truthy values: ``1`` or ``true``.
```
nicolasaunai, line 28: We should probably have a link to phlop's documentation?

doc/source/usage/environment_variables.rst
```diff
@@ -0,0 +1,68 @@
...
+``PHARE_SIM_MON``
+-------------------
+
+Enables periodic resource-usage monitoring during a simulation run. Truthy values: ``true``,
+``1``, ``t`` (case-insensitive). Can be overridden per-run via ``Simulator.run(monitoring=...)``.
+
+``PHARE_DRY_RUN``
+-------------------
+
+If ``1``, defaults ``Simulation.dry_run`` to ``True``, which makes ``Simulator.run()`` a no-op
```
nicolasaunai, line 50: We should probably explain here this mode still initializes the simulation? (if true, I already am confused)

doc/source/usage/environment_variables.rst
```diff
@@ -0,0 +1,68 @@
...
+(returns immediately without advancing). Used for CI dry-run validation of functional test scripts
+without actually executing them.
+
+``PYPHARE_LOG_LEVEL``
+-----------------------
+
+Sets the Python ``logging`` level used by PHARE's logger. One of ``INFO`` (default), ``WARNING``,
```
nicolasaunai, line 57: I think we should explain what the difference is between these levels

doc/source/usage/environment_variables.rst
PhilipDeegan: we could also include this

https://github.com/PhilipDeegan/PHARE/blob/mkn/doc/perf.md

doc/source/usage/environment_variables.rst
PhilipDeegan: there's nothing in the code base for reading them but I did write [this](https://github.com/PhilipDeegan/PHARE/blob/runtime_info/tools/python3/stats.py) which might be useful

doc/source/usage/environment_variables.rst
PhilipDeegan: it doesn't initialize if true, only C++ construction is validated, could be more explicit sure

doc/source/usage/environment_variables.rst
PhilipDeegan: this is only used in one place that I can see, and the entire apparatus could probably be deleted

nicolasaunai: Or extended? Anyway, I think I was confused with PHARE_LOG.

doc/source/usage/environment_variables.rst
nicolasaunai: ok that will come with the simulation documentation. to which that could be linked eventually

doc/source/usage/environment_variables.rst
nicolasaunai: Yes some curated version of that, but a link to phlop makes sense otherwise user reading "enables phlop" are like 😳 maybe

doc/source/usage/environment_variables.rst
nicolasaunai: you don't get phlop to read that?

PhilipDeegan: phlop is for the process monitoring/inspection of RAM and CPU usage, it generates a fairly simple YAML file which the previous linked script scans and prints some stats for

nicolasaunai: ok thanks

PR #1271, opened 2026-07-22
Decouple hybrid and MHD compile surfaces

UCaromel:
## Why

The simulator's type assembly treated hybrid and MHD as one combined axis rather than two independent ones. Every compiled module carried the full hybrid parameter set (`dim`, `interp_order`, `refined_particle_nbr`) *and* the full MHD permutation, regardless of which model it was for.

Three consequences:

- **MHD-only runs instantiate hybrid particle machinery**, and hybrid-only runs instantiate the MHD timestepper/reconstruction stack. Neither is reachable at runtime; both are paid for at compile time and in binary size.
- **The permutation list is a cross-product.** Adding one MHD reconstruction multiplies against every hybrid `interp_order`, so the module count grows in the wrong direction and most of the resulting combinations are meaningless.
- **The two models aren't actually separable.** Anything that wants to touch the MHD side has to reason about hybrid template parameters that have nothing to do with it. That's a standing tax on every subsequent MHD change, and it's what motivated this PR — the coupled MHD-hybrid work needs the two surfaces to be independently addressable before it can go anywhere.

The parameters were never genuinely shared; they were only ever adjacent in the same template list. This makes the separation explicit.

## What changes

The compile surface is split along the model axis. A module carries the hybrid fields iff `interp_order` is set, and the MHD fields iff a `mhd_timestepper` is set, yielding three shapes:

| shape | fields |
|---|---|
| hybrid-only | 3 |
| MHD-only | 8 |
| coupled | 10 |

`res/sim/all.txt` is sectioned to match: 32 + 18 + 1 = 51 modules.

Python side follows symmetrically — `simulator_id()` assembles over both axes independently, `check_interp_order` defaults to 1 for `HybridModel` and 0 for `MHDModel`, and the gridlayout helpers now raise instead of silently wrapping to the `interp_order == 3` ghost value via negative indexing when `interp_order` is 0 with no explicit `ghosts_nbr`.

Two files move to where they belong rather than where they'd accumulated: `mhd_resolver.hpp` out of `python3/` (it has no Python content) into `amr/solvers/`, and `HybridSimState`/`MHDSimState` into `phare_solver.hpp` beside the `HybridStack`/`MHDStack` that use the same idiom.

## Gates

- Compile: **434/434** targets, on each of hybrid-only, MHD-only and coupled permutations.
- Runtime: one Harris timestep on each of the three shapes.
- ctest: ghost-width calculator and gridlayout-integration, unchanged from pre-branch behavior.

## Note on history

Previously based on the `gridlayout` branch so the diff would read cleanly. That branch is now upstream as a squash (PHAREHUB/PHARE#1205), so this is rebased with `--onto` to drop the six duplicated commits and now sits as 2 commits on current master. `git range-diff` confirms both commits unchanged across the rebase.


src/python3/cpp_simulator.hpp
```diff
@@ -55,10 +55,15 @@ void declareSimulator(PyClass&& sim)
         .def("dump_restarts", &Simulator::dump_restarts, py::arg("timestamp"), py::arg("timestep"));
 }
 
+// opts MUST come in through Sim, never from resolve_simulator_options() directly: a constexpr
+// initialized from a dependent expression is itself value-dependent, so DataWrangler<opts> /
+// PatchLevel<opts> stay dependent types and are only processed when this body is instantiated.
+// Read straight from resolve_simulator_options(), they would be non-dependent and get processed
+// even inside a discarded `if constexpr` branch — which breaks MHD-only builds.
 template<typename Sim>
```
PhilipDeegan: `template<auto sim_opts>` could be simpler and not need such a comment

src/simulator/simulator.hpp
```diff
@@ -70,26 +70,24 @@ template<auto opts>
 class Simulator : public ISimulator
 {
 public:
+    // exposed so python bindings can reach the options through the Simulator type: naming them as
+    // `Sim::options` keeps them a dependent expression, which is what makes `if constexpr` on them
+    // actually discard the untaken branch. See declare_etc in python3/cpp_simulator.hpp.
```
PhilipDeegan: imagine if we added this kind of comment for every public export

it's used because it's used is somewhat tautological eventually

src/amr/load_balancing/load_balancer_estimator_hybrid.hpp
```diff
@@ -17,7 +17,7 @@ namespace PHARE::amr
 template<typename PHARE_T>
 class LoadBalancerEstimatorHybrid : public LoadBalancerEstimator
 {
-    using HybridModel = typename PHARE_T::HybridModel_t;
+    using HybridModel = PHARE_T::Hybrid::Model_t;
```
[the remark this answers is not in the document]
PhilipDeegan: false, this is C++20

UCaromel:
Follow up, python gridlayout PR in https://github.com/UCaromel/PHARE/pull/43

UCaromel:
image

PhilipDeegan:
1. "Available at compile time" ≠ "instantiated at compile time." For an implicit instantiation (the normal case — PHARE_Types<opts>::Ions_t used somewhere in a .cpp), the compiler only instantiates the members you actually name. Electrons_t sitting unused in the struct costs nothing if nothing references it in that TU — C++ template instantiation is lazy per-member for class templates. My probe used template struct PHARE::core::PHARE_Types<...> (an explicit instantiation), which forces every single alias and member to instantiate whether used or not — that's why it was useful for inspection, but it's not representative of ordinary compile cost.
2. "Nothing unused remains in the binary" is a linker/optimizer property, not a language guarantee. Templates get instantiated per-TU (so GridLayout_t might get compiled identically in 20 .cpp files), then:
  - the linker deduplicates identical COMDAT sections (weak/inline linkage) across TUs,
  - --gc-sections/LTO strip functions with no remaining callers.

That's usually "on" in this build (-O3, likely -ffunction-sections/LTO in release configs — worth checking res/cmake/*.cmake if you want it guaranteed), but it's an optimization pass, not something the template system itself promises. Without those flags, unused-but-instantiated code can linger in .os and even survive into the final binary as dead symbols.

So the real cost split is: compile-time instantiation cost is paid per translation unit for whatever's actually used in that TU (this is where PHARE's heavy template stack — SimOpts → PHARE_Types → Ions_t/Electrons_t — gets expensive, since many TUs each re-instantiate the same deep chains), while final-binary bloat is cleaned up afterward by the linker/optimizer, not by the templates being "smart" about it.

pyphare/pyphare/core/gridlayout.py
```diff
@@ -205,12 +205,22 @@ def yeeCoordsFor(
...
+        )
+
+
 def HybridGridLayoutFor(
     box, origin, dl, interp_order, ghosts_nbr=None, is_particle_layout=False
 ):
     field_ghosts = [2, 4, 4]
     particle_ghosts = [1, 2, 2]
 
+    check_hybrid_ghosts_or_interp("HybridGridLayoutFor", not ghosts_nbr, interp_order)
```
nicolasaunai: I find this notation `not None` misleading/confusing, is it me?

pyphare/pyphare/cpp/__init__.py
```diff
@@ -13,17 +13,26 @@
...
-        f"{sim.mhd_timestepper}_{sim.reconstruction}_{sim.limiter}_"
-        f"{sim.riemann}_{Hall}_{Res}_{Hyper_Res}"
-    )
+    parts = [str(sim.ndim)]
+
+    if sim.interp_order:
+        parts += [str(sim.interp_order), str(sim.refined_particle_nbr)]
+
+    if sim.mhd_timestepper:
+        Hall = "true" if sim.hall else "false"
```
nicolasaunai: I would name them xxx_active ou xxxON

pyphare/pyphare/pharein/simulation.py
```diff
@@ -210,10 +210,12 @@ def _final_time():
...
 
 def check_interp_order(**kwargs):
-    interp_order = kwargs.get("interp_order", 1)
+    model_options = phare_utilities.listify(kwargs.get("model_options", "HybridModel"))
+    default_interp_order = 1 if "HybridModel" in model_options else 0
+    interp_order = kwargs.get("interp_order", default_interp_order)
 
-    if interp_order not in [1, 2, 3]:
-        raise ValueError("Error: invalid interpolation order. Should be in [1,2,3]")
+    if interp_order not in [0, 1, 2, 3]:
```
nicolasaunai: maybe the list of valid interp orders should be named

pyphare/pyphare/pharein/simulation.py
```diff
@@ -296,6 +298,17 @@ def check_boundaries(ndim, **kwargs):
 
 def check_refined_particle_nbr(ndim, **kwargs):
     interp = kwargs["interp_order"]
+
+    # C++ Simulator Options expects a value for `refined_particle_nbr` even if hybrid
+    # models are not active.
+    if interp == 0:
```
nicolasaunai: that reads weird, I think what you really mean is something like `if not is_hybrid` or something and you use `interp_order==0` as a proxy of that?

src/amr/solvers/mhd_resolver.hpp
```diff
@@ -43,118 +42,107 @@ struct SlopeLimiterSelector;
...
 {
-    static constexpr std::uint32_t nghosts
-        = MHDOpts::reconstruction_nghosts_v<MHDOpts::ReconstructionType::MP5>;
     template<typename GridLayout, typename SlopeLimiter>
     using type = core::MP5Reconstruction<GridLayout, SlopeLimiter>;
 };
 
-template<MHDOpts::ReconstructionType R, MHDOpts::SlopeLimiterType S>
-struct SlopeLimiterSelector
+// No generic primary (forward-declared above only): an (R, S) combo not explicitly listed below
```
nicolasaunai: I don't understand this comment

src/python3/cpp_simulator.hpp
```diff
@@ -55,10 +55,15 @@ void declareSimulator(PyClass&& sim)
         .def("dump_restarts", &Simulator::dump_restarts, py::arg("timestamp"), py::arg("timestep"));
 }
 
+// opts MUST come in through Sim, never from resolve_simulator_options() directly: a constexpr
```
nicolasaunai: I don't get this comment

src/simulator/simulator.hpp
```diff
@@ -343,19 +340,22 @@ void Simulator<opts>::hybrid_init(initializer::PHAREDict const& dict)
...
+    using MHDModel  = typename PHARETypes::MHD::Model_t;
+    using SolverMHD = typename PHARETypes::MHD::Solver_t;
+
+    mhd_.model_ = std::make_shared<MHDModel>(dict["simulation"], mhd_.resman_);
+    mhd_.resman_->registerResources(mhd_.model_->state);
 
     // we register the mhd model for all possible levels in the hierarchy
     // since for now it is the only model available, same for the solver
-    multiphysInteg_->registerModel(0, maxMHDLevel_ - 1, mhdModel_);
+    multiphysInteg_->registerModel(0, maxMHDLevel_ - 1, mhd_.model_);
```
nicolasaunai, line 356: this is not specific to this PR but I find  `maxMHDLevel_ - 1` not intuitive, to me  `maxMHDLevel_` means the max mhd index, but here looks like it's the MHD level count

src/simulator/simulator.hpp
```diff
@@ -440,26 +440,56 @@ Simulator<opts>::Simulator(PHARE::initializer::PHAREDict const& dict,
 
     // we would need a different restart manager for mhd and hybrid if both models are used
 
-    if (find_model("HybridModel"))
+    // the build enabling a model (static_assert in solver::PHARE_Types) is not enough: the dict
+    // still has to ask for one of them by name.
+    bool anyModel = false;
+
+    if constexpr (is_hybrid_v<opts>)
```
nicolasaunai: shouldn't it rather be `has_hybrid_v`than is ?

src/simulator/simulator.hpp
```diff
@@ -440,26 +440,56 @@ Simulator<opts>::Simulator(PHARE::initializer::PHAREDict const& dict,
...
-        hybrid_init(dict);
-        if (dict["simulation"].contains("restarts"))
-            rMan = restarts::RestartsManagerResolver::make_unique(*hierarchy_, *hyb_resman_ptr,
-                                                                  dict["simulation"]["restarts"]);
+        if (find_model("HybridModel"))
+            throw std::runtime_error("HybridModel requested but this build has no hybrid support");
     }
 
-    if (find_model("MHDModel"))
+    if constexpr (is_mhd_v<opts>)
```
nicolasaunai: has ?

src/simulator/simulator.hpp
UCaromel: probably better yes

src/simulator/simulator.hpp
UCaromel: agreed but probably out of scope ?

pyphare/pyphare/core/gridlayout.py
PhilipDeegan: `!static_cast<bool>(ghosts)` when it could just pass teh ghosts and check if it's none

pyphare/pyphare/cpp/__init__.py
PhilipDeegan: this won't exist in a while

src/python3/cpp_simulator.hpp
PhilipDeegan: claudio adding in what he was told to do

src/simulator/simulator.hpp
```diff
@@ -192,12 +190,8 @@ class Simulator : public ISimulator
...
     bool allowEmergencyDumps = false;
 
-    std::shared_ptr<HybridResourceManager_t> hyb_resman_ptr;
-    std::shared_ptr<MHDResourceManager_t> mhd_resman_ptr;
-
-    // physical models that can be used
-    std::shared_ptr<HybridModel> hybridModel_;
-    std::shared_ptr<MHDModel> mhdModel_;
+    solver::HybridSimState<opts, is_hybrid_v<opts>> hyb_;
+    solver::MHDSimState<opts, is_mhd_v<opts>> mhd_;
```
PhilipDeegan: you could probably simplify this and move the conditions to the structs like 

https://coliru.stacked-crooked.com/a/7a307647d2c4fc32

src/simulator/simulator.hpp
```diff
@@ -281,12 +275,15 @@ void Simulator<opts>::diagnostics_init(initializer::PHAREDict const& dict, auto&
 template<auto opts>
 void Simulator<opts>::hybrid_init(initializer::PHAREDict const& dict)
 {
-    hybridModel_ = std::make_shared<HybridModel>(dict["simulation"], hyb_resman_ptr);
-    hyb_resman_ptr->registerResources(hybridModel_->state); // still valid, never moved
+    using HybridModel = typename PHARETypes::Hybrid::Model_t;
+    using SolverPPC   = typename PHARETypes::Hybrid::Solver_t;
```
PhilipDeegan: probably redundant `typename`

pyphare/pyphare/cpp/__init__.py
PhilipDeegan: correction, hall will remain, the other two (res/hyper-res) will not

src/phare_solver.hpp
```diff
@@ -9,68 +9,179 @@
...
+#include "amr/solvers/mhd_resolver.hpp"
+
+#include <memory>
 
 namespace PHARE::solver
 {
+// Bool-specialized holders: the `false` specialization is empty, so an opts value with
+// hybrid (or mhd) disabled never names the corresponding Model_t/Solver_t/etc. as a type —
+// that is what keeps the disabled model's templates from ever being instantiated.
+// the default belongs here and only here: partial specializations may not carry default
```
PhilipDeegan: > the default belongs here and only here

as opposed to? this comment could probably be just 

"compile time model enabling to minimize ununused expansion/binary symbols" 

or something to that effect maybe

src/phare_simulator_options.hpp
```diff
@@ -25,16 +25,33 @@ struct SimOpts
...
     bool HyperResistivity                            = false;
+
+    // derived — two independent axes, never one as the other's negation
+    bool hybrid_enabled = interp_order > 0;
+    bool mhd_enabled    = reconstruction_type != MHDOpts::ReconstructionType::MHDOff;
+
+    // all four MHD axes must agree on off-ness; a member fn keeps SimOpts an aggregate
+    constexpr bool mhd_axes_consistent() const
+    {
+        bool const off = !mhd_enabled;
```
PhilipDeegan: `off` is statement, so maybe `is_off` ?

PhilipDeegan: yeah sorry this is weird, `mhd_enabled` might itself be false, so in that it's not enabled

pyphare/pyphare/pharein/simulation.py
```diff
@@ -296,6 +307,17 @@ def check_boundaries(ndim, **kwargs):
 
 def check_refined_particle_nbr(ndim, **kwargs):
     interp = kwargs["interp_order"]
+
+    # C++ Simulator Options expects a value for `refined_particle_nbr` even if hybrid
+    # models are not active.
+    if not hybrid_is_active(interp):
+        refined_particle_nbr = kwargs.get("refined_particle_nbr", 0)
```
PhilipDeegan: I would say from a client API standpoint, both interp and refined_particle_nbr == 0 are invalid

they key should either not exist or be None maybe, not sure

tests/core/utilities/ghost_width_calculator/test_gridlayout_integration.cpp
```diff
@@ -29,34 +29,44 @@ TEST(GridLayoutIntegration, UsesGhostWidthOrder3)
 
 // MHD ghost width tests - varying by reconstruction stencil
 
+// MHD-only options: interp_order 0 leaves hybrid off, the four MHD axes are all on so
```
PhilipDeegan, line 33: you can put these tests in the `namespace PHARE::core` which might dedupe somethings

nicolasaunai (approved):
LGTM once phil's comments are taken care of

PR #1272, opened 2026-07-22
move mpi utilities/etc out of the core module

### LINKED ISSUE #1179: move mpi out of the core module
Having it in the core module gives the impression that it can be used from the core module, which misleading.

PhilipDeegan:
suggestive refactor, I've no firm opinion on where these things should live outside of core

closes #1179

biggest changes

- ‎‎src/core/def/phare_mpi.hpp renamed to src/phare_mpi.hpp
- src/core/utilities/mpi_utils.cpp renamed to src/mpi/mpi_utils.cpp
- src/core/utilities/mpi_utils.hpp renamed to src/mpi/mpi_utils.hpp


nicolasaunai:
didn't we say ok to merge that?

nicolasaunai (approved):

PR #1278, opened 2026-07-23
Second-order AMR: higher-order field refinement + coarse–fine temporal fixes

UCaromel:
## Why

AMR on `master` is not second-order accurate. The field refinement (coarse→fine spatial interpolation) and the coarse–fine *time* interpolation are only first-order, so an AMR run does not recover the scheme's second-order accuracy near refinement boundaries — accuracy is effectively capped below 2nd order wherever levels exist.

This PR makes AMR genuinely **second-order in both space and time**:
- second-order spatial field refinement, and
- second-order coarse–fine temporal interpolation — including for **B**, which previously had none.

The existing first-order (legacy) refinement path is **kept**. Flipping the default and/or removing the legacy path is a team decision; for now both coexist and are selectable at runtime, so this PR changes no default behaviour.

## How

- **Kernel-based refinement operator** — spatial refinement is expressed as composable stencil kernels rather than a hardcoded scheme. This is what makes 2nd order clean, and makes it easy to extend to higher spatial orders and to swap in a WENO-limited variant.
- **Correct stage times** — RK integrators and the PPC predictor/corrector fill coarse–fine ghosts at the proper sub-step times rather than at endpoint values.
- **Time interpolation everywhere** — coarse–fine temporal interpolation now covers all evolved fields, including **B** (previously static in time).

Selectable from the Python config via the refinement order (and limiter); legacy = order 0.

## Validation

2D AMR spatial convergence (composite L2-of-L1 norm, MC2011). With `refinement_order=2`, both alfven2d and whistler2d converge at slope 2, inside the calibrated gate band (1.70, 2.25):

convergence_hor_branch_o2_simple


src/core/data/grid/impl/yee/gridlayout_mhd_yee.hpp
```diff
@@ -584,6 +585,55 @@ class GridLayoutImplYee
...
+        }
+        else
+        {
+            auto make_p = [](int offset) {
+                Point<int, dimension> p{};
+                p[dir] = offset;
+                return p;
+            };
+
+            if constexpr (order == 0)
```
UCaromel, line 621: torm

PhilipDeegan:
given how broad this is, it's probably worth running the nightly build prior to merge

I suggest you do so [here](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Caromel_Build)

tests/simulator/amr_convergence/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-alfven2d-spatial-o2      test_alfven2d_amr.py   ${CMAKE_CURRENT_BINARY_DIR} AlfvenConvergenceTest.test_spatial_convergence_1_2)
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-alfven2d-spatial-legacy  test_alfven2d_amr.py   ${CMAKE_CURRENT_BINARY_DIR} AlfvenConvergenceTest.test_spatial_convergence_legacy)
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-alfven2d-temporal        test_alfven2d_amr.py   ${CMAKE_CURRENT_BINARY_DIR} AlfvenConvergenceTest.test_temporal_sigma_sweep)
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-whistler2d-spatial-o2     test_whistler2d_amr.py ${CMAKE_CURRENT_BINARY_DIR} WhistlerConvergenceTest.test_spatial_convergence_1_2)
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-whistler2d-spatial-legacy test_whistler2d_amr.py ${CMAKE_CURRENT_BINARY_DIR} WhistlerConvergenceTest.test_spatial_convergence_legacy)
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-whistler2d-temporal       test_whistler2d_amr.py ${CMAKE_CURRENT_BINARY_DIR} WhistlerConvergenceTest.test_temporal_sigma_sweep)
```
PhilipDeegan: these are classed as functional tests essentially and so are executed in "dry run" mode ([see](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_BuildGithubPrGccSs/95353?buildTab=log&linesState=1734&logView=flowAware&focusLine=2587)) and require something like [this](https://github.com/PHAREHUB/PHARE/blob/master/tests/functional/harris/harris_2d.py#L221) now so to not try to plot when in dry run mode

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -0,0 +1,235 @@
...
+#include <memory>
+#include <stdexcept>
+#include <utility>
+
+
+namespace PHARE::amr
+{
+
+/** @brief Tag for the unlimited (smooth) prolongation path. */
+struct NoLimiter
```
PhilipDeegan: maybe either a better name or put it in a subnamespace?

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -0,0 +1,235 @@
...
+        return value;
+    }
+
+    // runtime centering+parity → compile-time stencil: one gather fn per combination, dispatched
+    // by a plain index. Each keeps its natural length; nothing is padded to a common size.
+    using Gatherer_t = double (*)(FieldT const&, Point_t const&);
+    static constexpr std::array<Gatherer_t, nCombined> gatherers_
+        = []<std::size_t... I>(std::index_sequence<I...>) {
+              return std::array<Gatherer_t, nCombined>{&gatherStencil_<I>...};
+          }(std::make_index_sequence<nCombined>{});
```
PhilipDeegan:
```
static constexpr auto gatherer_funcs_ = core::for_N_make_array<nCombined>([](auto i){ return &gatherStencil_<i>});
```

same?

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -0,0 +1,235 @@
...
+        else if constexpr (dimension == 2)
+        {
+            if (std::isnan(dst(fineLocal[0], fineLocal[1])))
+                dst(fineLocal[0], fineLocal[1]) = value;
+        }
+        else
+        {
+            if (std::isnan(dst(fineLocal[0], fineLocal[1], fineLocal[2])))
+                dst(fineLocal[0], fineLocal[1], fineLocal[2]) = value;
+        }
```
PhilipDeegan: pretty sure the dimension checks are redundant

```
if(auto& dst_val = dst(fineLocal); std::isnan(dst_val))
    dst_val = value;
```

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -0,0 +1,235 @@
...
+            return src(anchorLocal[0] + offset[0]);
+        else if constexpr (dimension == 2)
+            return src(anchorLocal[0] + offset[0], anchorLocal[1] + offset[1]);
+        else
+            return src(anchorLocal[0] + offset[0], anchorLocal[1] + offset[1],
+                       anchorLocal[2] + offset[2]);
+    }
+
+    // preserve the legacy NaN-guard: only fill fine indices not already set
+    static void assignFine_(FieldT& dst, Point_t const& fineLocal, double value)
```
PhilipDeegan: `const value`

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+};
+
+
+// ---- factory (declared in field_refiner_kernel.hpp) ---------------------------------------------
+
+template<typename GridLayoutT, typename FieldT>
+std::shared_ptr<IFieldRefineKernel<GridLayoutT, FieldT>>
+makeRefineKernel(int order)
```
PhilipDeegan: `int const`

src/amr/data/field/refine/field_refine_operator.hpp
```diff
@@ -202,6 +204,157 @@ class TensorFieldRefineOperator : public SAMRAI::hier::RefineOperator
...
+ * and forwards each overlap box to the kernel. The legacy templated FieldRefineOperator above is
+ * left byte-identical; this operator is used ONLY on the composite order=2/4 path, so the default
+ * (order absent) carries no behavior change.
+ */
+template<typename GridLayoutT, typename FieldT>
+class KernelFieldRefineOperator : public SAMRAI::hier::RefineOperator
+{
+public:
+    static constexpr std::size_t dimension = GridLayoutT::dimension;
+    using PhysicalQuantity                 = typename FieldT::physical_quantity_type;
```
PhilipDeegan: redundant `typename`

src/amr/messengers/mhd_messenger.hpp
```diff
@@ -497,35 +501,64 @@ namespace amr
...
     private:
+        // Select the field-refinement operators once at construction. order==0 keeps the legacy
+        // per-quantity policies (byte-identical to master); order==2/4 swaps in the composite
+        // runtime kernels. B uses the shared-face magnetic kernel (interior stays Tóth-Roe).
+        void makeRefineOperators_(RefinementConfig const& config)
+        {
+            if (config.order)
+            {
+                auto fieldKernel = [&] {
+                    return std::make_shared<KernelFieldRefineOperator<GridLayoutT, GridT>>(
```
PhilipDeegan: does it need to be shared?

you're invalidating any shared_ptr from the call site by doing std::move in the constructor

src/core/numerics/ampere/ampere.hpp
```diff
@@ -4,13 +4,22 @@
...
+
 namespace PHARE::core
 {
 
 
+enum class AmpereMode { ShrinkedGhost, GrownPhysical };
 
+struct AmpereBox // structural type — usable as NTTP
+{
+    AmpereMode mode      = AmpereMode::ShrinkedGhost;
```
PhilipDeegan, line 17: I thought the design we discussed was more general than this and about "FieldBoxMutators"

src/core/numerics/ampere/ampere.hpp
```diff
@@ -31,19 +40,23 @@ class Ampere
         auto& Jy = J(Component::Y);
         auto& Jz = J(Component::Z);
 
-        Point<std::uint32_t, dimension> shrink;
+        Point<std::uint32_t, dimension> amount;
```
PhilipDeegan, line 43: if you considered the more general approach it could just be baked into gridLayout

like `evalOnMutatedGhostBox<mutator>(...)`

src/simulator/simulator.hpp
```diff
@@ -36,6 +37,19 @@ namespace PHARE
 {
 
 
+//! Read the optional field-refinement selection from the dict. Absent keys ⇒ order 0 ⇒ legacy
+//! operators (no behavior change vs master).
+inline PHARE::amr::RefinementConfig
+refinementConfigFromDict(PHARE::initializer::PHAREDict const& dict)
+{
+    PHARE::amr::RefinementConfig config;
```
PhilipDeegan: what we have in other places is just a static function on the class called `FROM(dict)` [see](https://github.com/PHAREHUB/PHARE/blob/master/src/amr/load_balancing/load_balancer_details.hpp#L38)

tests/amr/data/field/refine/test_field_refine.cpp
```diff
@@ -380,6 +411,127 @@ TYPED_TEST(aFieldLinearRefineIndexesAndWeights, giveACorrectWeightIndexesForDual
...
+{
+    std::array<double, 8> coarse = {1.0, 1.3, 0.4, 2.0, -0.5, 0.7, 0.9, 0.2};
+    std::array<QtyCentering, 1> centering{QtyCentering::dual};
+
+    auto run = [&](auto refiner) {
+        Grid1D src{"c", HybridQuantity::Scalar::rho, 8u};
+        Grid1D dst{"f", HybridQuantity::Scalar::rho, 16u};
+        for (std::size_t i = 0; i < 8; ++i)
+            src(i) = coarse[i];
+        for (std::size_t i = 0; i < 16; ++i)
```
PhilipDeegan: `dst.fill(nan)` ?

tests/amr/data/field/refine/test_field_refine.cpp
```diff
@@ -380,6 +411,127 @@ TYPED_TEST(aFieldLinearRefineIndexesAndWeights, giveACorrectWeightIndexesForDual
...
+// at every order.
+TEST(compositeRefiner1D, dualChildrenConserveCoarseAverage)
+{
+    std::array<double, 8> coarse = {1.0, 1.3, 0.4, 2.0, -0.5, 0.7, 0.9, 0.2};
+    std::array<QtyCentering, 1> centering{QtyCentering::dual};
+
+    auto run = [&](auto refiner) {
+        Grid1D src{"c", HybridQuantity::Scalar::rho, 8u};
+        Grid1D dst{"f", HybridQuantity::Scalar::rho, 16u};
+        for (std::size_t i = 0; i < 8; ++i)
```
PhilipDeegan: ` < src.shape()[0]`

tests/amr/data/field/refine/test_field_refine.cpp
```diff
@@ -380,6 +411,127 @@ TYPED_TEST(aFieldLinearRefineIndexesAndWeights, giveACorrectWeightIndexesForDual
...
+            src(i) = coarse[i];
+        for (std::size_t i = 0; i < 16; ++i)
+            dst(i) = NaN;
+
+        refiner.refineBox(src, dst, boxOf<1>({4}, {11}), centering, boxOf<1>({0}, {15}),
+                          boxOf<1>({0}, {7}), ratio2(1));
+
+        for (int I = 2; I <= 5; ++I)
```
PhilipDeegan: why capitalize?

tests/amr/data/field/refine/test_field_refine.cpp
PhilipDeegan: or even `Grid1D dst{"f", HybridQuantity::Scalar::rho, 16u, NaN};`

tests/amr/data/field/refine/test_field_refine.cpp
```diff
@@ -380,6 +411,127 @@ TYPED_TEST(aFieldLinearRefineIndexesAndWeights, giveACorrectWeightIndexesForDual
...
+                src(ix, iy) = src_at(ix, iy);
+        for (std::size_t ix = 0; ix < 12; ++ix)
+            for (std::size_t iy = 0; iy < 12; ++iy)
+                dst(ix, iy) = NaN;
+
+        refiner.refineBox(src, dst, boxOf<2>({4, 4}, {5, 7}), centering, boxOf<2>({0, 0}, {11, 11}),
+                          boxOf<2>({0, 0}, {5, 5}), ratio2(2));
+
+        // fine x=4 is a shared (even) face, anchor Ix=2; x=5 is interior (odd) → skipped (NaN)
+        for (int Iy = 2; Iy <= 3; ++Iy)
```
PhilipDeegan: capitalize?

tests/core/data/gridlayout/test_linear_combinations_yee.cpp
```diff
@@ -232,3 +232,87 @@ TEST(BzToEz, combinationOk)
...
+
+using PHARE::core::dirX;
+using ImplYee1 = PHARE_Types<PHARE::SimOpts{1, 1}>::Hybrid::GridLayout_t::implT;
+
+namespace
+{
+    // sum of a stencil's coefficients
+    auto coefSum(auto const& row)
+    {
+        double s = 0.;
```
PhilipDeegan: we do have 

`PHARE::core::sum(container);`

and

`PHARE::core::sum_from(container, func);`

tests/core/data/gridlayout/test_linear_combinations_yee.cpp
```diff
@@ -232,3 +232,87 @@ TEST(BzToEz, combinationOk)
...
+    auto coefAt(auto const& row, int offset)
+    {
+        for (auto const& wp : row)
+            if (wp.indexes[dirX] == offset)
+                return wp.coef;
+        return 0.;
+    }
+
+    // value the stencil produces from coarse cell-averages u(I+offset) = f(I+offset)
+    auto applyRow(auto const& row, int I, auto&& f)
```
PhilipDeegan: `I` -> `ix` ?

PhilipDeegan: I think we prefer `span` over `row` generally also

tests/functional/refinement/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+if(NOT ${PHARE_PROJECT_DIR} STREQUAL ${CMAKE_BINARY_DIR})
+  file(GLOB PYFILES "*.py")
+  file(COPY ${PYFILES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+endif()
+
+if(HighFive AND testMPI)
+  ## divB preservation of the higher-order field-refinement operators (order 0/2).
+  ## These tests dump B diagnostics and read them back, so they require HighFive.
+  # exec level 11
+  if(testMPI)
```
PhilipDeegan: redundant if

tests/functional/refinement/divb_refinement.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+if __name__ == "__main__":
+    argv = sys.argv[1:]
+    mode = argv[0] if argv and argv[0] in ("boxes", "tagging") else "boxes"
+    orders = [int(a) for a in argv if a.isdigit()] or [0, 2]
+
+    res = {o: run_variant(mode, o) for o in orders}
+    if cpp.mpi_rank() != 0:
+        sys.exit(0)
```
PhilipDeegan: probably not the best, maybe just do the inverse

```
if cpp.mpi_rank() == 0
    doX()
```

tests/functional/refinement/divb_refinement.py
```diff
@@ -0,0 +1,246 @@
...
+        if not (floor_ok and rel_ok):
+            ok = False
+        print(
+            f"  {label}: fine={m:.3e}  coarse={coarse:.3e}  base={base:.3e}  "
+            f"fine/coarse={m / coarse if coarse else float('inf'):.2f}  "
+            f"fine/base={m / base if base else float('inf'):.2f}  [{status}]",
+            flush=True,
+        )
+    print(f"DIVB_{mode.upper()}_OK" if ok else f"DIVB_{mode.upper()}_FAIL", flush=True)
+    sys.exit(0 if ok else 1)
```
PhilipDeegan: raise instead, this can have unintended side effects

tests/simulator/amr_convergence/compute_errors.py
```diff
@@ -0,0 +1,141 @@
...
+            sl.append(slice(lo, hi + 1))
+        mask[tuple(sl)] = 0.0
+    return mask
+
+
+def composite_errors(run, final_time, fine_box, refinement_ratio=2, ndim=2):
+    """(per_level eps, composite eps) -- L2-of-L1 over conserved vars."""
+    f1 = conserved_fields(run, final_time)
+    f0 = conserved_fields(run, 0.0)
+    assert f1.keys() == f0.keys(), (
```
PhilipDeegan: asserts can be disabled with `python -O` so we probably prefer exceptions or something that always triggers

tests/simulator/amr_convergence/test_alfven2d_amr.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # dump the conserved set natively so the comparison variables are the
+        # dumped cell/face averages directly (no primitive->conserved
+        # reconstruction in the analysis).
+        timestamps = [0.0, self.final_time]
+        ph.ElectromagDiagnostics(quantity="B", write_timestamps=timestamps)
+        for quantity in ["rho", "rhoV", "Etot"]:
+            ph.MHDDiagnostics(quantity=quantity, write_timestamps=timestamps)
+
+    @data(2)  # Linear2 runtime kernel
```
PhilipDeegan: are we expecting other variants? maybe redundant otherwise

src/core/numerics/ampere/ampere.hpp
PhilipDeegan: also what is that comment, talk about overkill

PhilipDeegan: it might actually be YAGNI - [this](https://github.com/PHAREHUB/PHARE/pull/1193/changes)  suggests hybrid can be ghost_box.shape - 1 and to skill the ghost schedule

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -194,38 +186,22 @@ class CompositeFieldRefiner : public IFieldRefineKernel<GridLayoutT, FieldT>
...
-        {
-            if (std::isnan(dst(fineLocal[0], fineLocal[1])))
-                dst(fineLocal[0], fineLocal[1]) = value;
-        }
-        else
-        {
-            if (std::isnan(dst(fineLocal[0], fineLocal[1], fineLocal[2])))
-                dst(fineLocal[0], fineLocal[1], fineLocal[2]) = value;
-        }
+        if (auto& dst_val = dst(fineLocal.toArray()); std::isnan(dst_val))
```
PhilipDeegan: you really need that `.toArray()` here?

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -176,13 +169,12 @@ class CompositeFieldRefiner : public IFieldRefineKernel<GridLayoutT, FieldT>
...
-          }(std::make_index_sequence<nCombined>{});
+    using Gatherer_t                 = double (*)(FieldT const&, Point_t const&);
+    static constexpr auto gatherers_ = core::for_N_make_array<nCombined>(
+        [](auto ic) { return Gatherer_t{&gatherStencil_<ic()>}; });
 
-    static double sampleCoarse_(FieldT const& src, Point_t const& anchorLocal, Point_t const& offset)
+    static double sampleCoarse_(FieldT const& src, Point_t const& anchorLocal,
+                                Point_t const& offset)
     {
         if constexpr (dimension == 1)
```
PhilipDeegan: is this not just `return src(anchorLocal + offset)`; ?

not sure this function needs to exist

src/amr/messengers/refinement_config.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef PHARE_REFINEMENT_CONFIG_HPP
+#define PHARE_REFINEMENT_CONFIG_HPP
+
+namespace PHARE::amr
```
UCaromel, line 10: This anticipates on possibility for limiting, and support to more refinement order that will come in the future. Not strictly needed right now.

tests/simulator/amr_convergence/test_alfven2d_amr.py
UCaromel: they are in the future, but the ddt part can come with them

UCaromel:
Converting this as draft for a bit. Some debugging tbd on run.py now that second order is the only operator.

tests/simulator/amr_convergence/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  file(COPY ${PYFILES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+endif()
+
+if(HighFive)
+  ## AMR space-time convergence studies (2D MHD); dump + read back h5 diagnostics.
+  ## One ctest entry per test method so `ctest -j` runs the independent sweeps
+  ## concurrently; inside an entry the sweep is sequential (order/drift need
+  ## all points together).
+  if(PHARE_PYTEST_SIMULATORS GREATER_EQUAL 2)
+    phare_mpi_python3_exec(21 ${PHARE_MPI_PROCS} amr-convergence-alfven2d-spatial-o2      test_alfven2d_amr.py   ${CMAKE_CURRENT_BINARY_DIR} AlfvenConvergenceTest.test_spatial_convergence)
```
PhilipDeegan, line 16: we just typically execute the file rather than specific tests within said files, this puts more pressure on the CPU during ctest executions, as ctest is just given a -j for total "processes", regardless of how many mpirun -n $N is used

src/amr/data/field/refine/coarse_cell_round_out.hpp
```diff
@@ -0,0 +1,128 @@
...
+NO_DISCARD inline std::string to_str(SAMRAI::hier::Box const& box)
+{
+    auto const dim = box.getDim().getValue();
+    std::string s{"[("};
+    for (SAMRAI::hier::Box::dir_t d = 0; d < dim; ++d)
+        s += std::to_string(box.lower(d)) + (d + 1 < dim ? "," : "),(");
+    for (SAMRAI::hier::Box::dir_t d = 0; d < dim; ++d)
+        s += std::to_string(box.upper(d)) + (d + 1 < dim ? "," : ")]");
+    return s;
+}
```
PhilipDeegan: this should probably be in `amr/utilities/box/amr_box.hpp`

PhilipDeegan:
[nightly build on Aug 8th](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_Users_Caromel_Build/96048) looks ok

not sure how important the conflicts are tho, may need rerunning

src/core/data/grid/impl/yee/gridlayout_hybrid_yee.hpp
```diff
@@ -396,6 +396,57 @@ class GridLayoutImplYee
...
+        {
+            auto make_p = [](int offset) {
+                Point<int, dimension> p{};
+                p[dir] = offset;
+                return p;
+            };
+
+            if constexpr (order == 0)
+            {
+                return std::array{WeightPoint{make_p(0), 1.0}};
```
UCaromel, line 439: torm

pyphare/pyphare/cpp/__init__.py
```diff
@@ -12,7 +12,8 @@
 _libs = {}
 
 
-def simulator_id(sim):
+def _simulator_id_parts(sim):
```
PhilipDeegan, line 15: this may be an improvement but I don't see why it's in this PR

pyphare/pyphare/pharein/simulation.py
```diff
@@ -904,6 +917,9 @@ def wrapper(simulation_object, **kwargs_in):
 
         kwargs["model_options"] = check_model_options(**kwargs)
 
+        refinement_order = check_refinement_operator(**kwargs)
+        kwargs["refinement_order"] = refinement_order
```
PhilipDeegan, line 921: `kwargs["refinement_order"] = check_refinement_operator(**kwargs)`

src/amr/data/field/refine/adpt_magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,457 @@
...
+using core::dirY;
+using core::dirZ;
+
+/**
+ * @brief Stage-2 (order-INdependent) cross-component divB touch-up of the Balsara ADPT
+ *        div-free magnetic prolongation.
+ *
+ * Paired with the fill-all composite kernel (stage 1, per-component, order-dialed): stage 1
+ * fills EVERY fine face of each B component from its own coarse faces; this stage equalizes,
+ * per coarse zone, the 2^d fine-subzone divergences of the filled data by adding the closed-form
```
PhilipDeegan, line 31: subzone? you mean overlap?

pyphare/pyphare/cpp/__init__.py
UCaromel: yes, i will open a standalone pr for this instead (#1323)

src/amr/data/field/refine/adpt_magnetic_refine_patch_strategy.hpp
UCaromel: i meant new fine cells, i am doing a claude comment rewrite pass right now.

src/amr/data/field/refine/composite_field_refiner.hpp
```diff
@@ -0,0 +1,199 @@
...
+                   SAMRAI::hier::Box const& intersectionBox,
+                   std::array<core::QtyCentering, dimension> const& centering,
+                   SAMRAI::hier::Box const& destFieldBox, SAMRAI::hier::Box const& sourceFieldBox,
+                   SAMRAI::hier::IntVector const& ratio) const override
+    {
+        for (std::size_t d = 0; d < dimension; ++d)
+            if (ratio(d) != 2)
+                throw std::runtime_error("CompositeFieldRefiner supports refinement ratio 2 only");
+
+        // Stage 1 fills every fine face, interior ones included; the stage-2 touch-up then
```
UCaromel, line 66: arguably these magnetic only part are not super idiomatic, i can propose something like this instead: 

```cpp
// composite_field_refiner.hpp — drop the isMagnetic param
protected:
    //! the fine region this kernel writes; the base writes exactly the overlap it was handed
    virtual SAMRAI::hier::Box destinationRegion_(
        SAMRAI::hier::Box const& overlap,
        std::array<core::QtyCentering, dimension> const&,
        SAMRAI::hier::Box const&) const { return overlap; }

// magnetic_composite_refiner.hpp — a real class, not an alias
class MagneticCompositeRefiner : public CompositeFieldRefiner<GridLayoutT, FieldT, order>
{
protected:
    SAMRAI::hier::Box destinationRegion_(...) const override
    {
        auto const primalCount = /* count primal centerings */;
        if (primalCount > 1) throw ...;             // at most one face normal
        if (primalCount == 0) return overlap;       // out-of-plane: no interior face
        return roundFieldBoxOutToCoarseCells<dimension>(overlap, centering) * destFieldBox;
    }
};
```

this is nicer but an extra virtual indirection per overlap (probably fine, to discuss ?)

UCaromel: this is likely better:
```cpp
// composite_field_refiner.hpp
namespace detail {
template<typename GridLayoutT, typename FieldT, std::size_t order>
struct CompositeFill                     // stateless, no vtable, holds the stencil tables
{
    static void fill(FieldT const& src, FieldT& dst,
                     SAMRAI::hier::Box const& region,   // caller has already resolved it
                     centering, destFieldBox, sourceFieldBox, ratio);   // ratio check inside
    static constexpr int coarseStencilWidth() { return order / 2; }
};
}

template<typename GridLayoutT, typename FieldT, std::size_t order>
class CompositeFieldRefiner final : public IFieldRefineKernel<GridLayoutT, FieldT>
{
    using Fill = detail::CompositeFill<GridLayoutT, FieldT, order>;
    void refineBox(..., overlap, centering, destFieldBox, srcFieldBox, ratio) const override
    { Fill::fill(src, dst, overlap, centering, destFieldBox, srcFieldBox, ratio); }
    int coarseStencilWidth() const override { return Fill::coarseStencilWidth(); }
};

// magnetic_composite_refiner.hpp — a sibling, not a subclass
template<typename GridLayoutT, typename FieldT, std::size_t order>
class MagneticCompositeRefiner final : public IFieldRefineKernel<GridLayoutT, FieldT>
{
    using Fill = detail::CompositeFill<GridLayoutT, FieldT, order>;

    //! B is primal in at most one direction (its face normal). With a normal, the gather must
    //! cover whole coarse cells so the stage-2 touch-up's reads exist; without one (By,Bz in 1D,
    //! Bz in 2D) there is no interior face and no round-out to do.
    static SAMRAI::hier::Box region(overlap, centering, destFieldBox)
    {
        auto const primalCount = /* count primal centerings */;
        if (primalCount > 1) throw std::runtime_error("...at most one primal (normal) direction");
        if (primalCount == 0) return overlap;
        return roundFieldBoxOutToCoarseCells<dimension>(overlap, centering) * destFieldBox;
    }

    void refineBox(..., overlap, centering, destFieldBox, srcFieldBox, ratio) const override
    { Fill::fill(src, dst, region(overlap, centering, destFieldBox), centering,
                 destFieldBox, srcFieldBox, ratio); }
    int coarseStencilWidth() const override { return Fill::coarseStencilWidth(); }
};
```

avoids a virtual call, and still give a clean separation of the magnetic specificity

src/amr/data/field/refine/adpt_magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,405 @@
...
+        auto const fine_field_box = core::for_N_make_array<N>([&](auto i) {
+            using PhysicalQuantity = std::decay_t<decltype(fields[i].physicalQuantity())>;
+
+            return FieldGeometry<gridlayout_type, PhysicalQuantity>::toFieldBox(
+                region, fields[i].physicalQuantity(), regionLayout);
+        });
+
+        // One stage-1 snapshot per pass: every correction must read stage-1 divergences.
+        // Recomputing from the live field couples sibling interior faces and breaks divB exactness.
+        DivCache cache;
```
PhilipDeegan, line 162: oh, this is heap allocation, could we not use a scratch space filled with nans instead?

src/amr/data/field/refine/adpt_magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,464 @@
...
+     * @brief The stage-1 divergence snapshot of one touch-up pass: one slot per fine cell of the
+     * reconstruction region, NaN meaning "not computed yet".
+     *
+     * Every correction has to read stage-1 divergences -- recomputing from the live field once a
+     * sibling interior face has been written couples the two faces and loses divB exactness -- so
+     * each subzone divergence is computed once per pass and reread afterwards. The region is a
+     * dense box of fine cells, so the slots are a flat array indexed by the cell's offset in it,
+     * and the array is kept between passes: only a region bigger than every earlier one allocates.
+     */
+    class DivScratch
```
PhilipDeegan: this could be out-of-line

src/amr/data/field/refine/adpt_magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,464 @@
...
+                offset = offset * static_cast<std::size_t>(shape_[d])
+                         + static_cast<std::size_t>(cell[d] - lower_[d]);
+            }
+
+            return offset;
+        }
+
+        core::Point<int, dimension> lower_;
+        core::Point<int, dimension> shape_;
+        std::vector<double> values_;
```
PhilipDeegan: my previous comment about dynamic allocation has gone missing, perhaps you missed it?

are you positive there is no way to to use already allocated storage?

src/amr/data/field/refine/adpt_magnetic_refine_patch_strategy.hpp
```diff
@@ -0,0 +1,464 @@
...
+            return;
+
+        auto const loc = layout.AMRToLocal(idx);
+        int const ix   = loc[dirX];
+        int const iy   = loc[dirY];
+
+        int const cyB = iy - 1;
+        int const cyA = iy;
+        int const cx0 = ix;                                   // own column
+        int const cx1 = ix + ((idx[dirX] % 2 == 0) ? 1 : -1); // sibling column
```
PhilipDeegan: could well be optimized out but this seems like an a bunch of unnecessary copies

you can just use ` by(loc)` for instance

PR #1281, opened 2026-07-24
Add 1d shock tube test.

iljah:


tests/functional/shock_tube/shock_tube.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+import matplotlib as mpl
+import numpy as np
+
+mpl.use("Agg")
+
+
+def shock_tube():
+    species_mass = 1.6726217770000001e-27
```
nicolasaunai, line 15: Simulation will not work with these SI units. The equations in the code base are normalized.

iljah: Right, I might update this at some point...

PR #1285, opened 2026-07-24
Test C++ 26

PhilipDeegan:
maybe it can give a simpler enum to string system


nicolasaunai:
maybe yes, and maybe the ResourcesManager object interface can also be gone.
We should assess availability of cpp26 features in compilers available on GENCI

igirault:
sure, but let us not forget [that](https://vittorioromeo.com/index/blog/refl_enum_to_string.html)

igirault:
Could we consider moving forward to c++23 first ? Only gcc 16 has partial support for reflections right now, while clang implements none.

PR #1287, opened 2026-07-26
vtkhdf: 64 bit Steps offsets

UCaromel:
## The bug

`VTKFileInitializer::initFileLevel` creates `Steps/Level<n>/{PointDataOffset/data, AMRBoxOffset, NumberOfAMRBox}` as **int32**, while the values written are `std::size_t` (`data_offset` and `box_offset` come from `getDimensions()[0]`). HDF5 clamps on datatype conversion, so past 2^31 rows of point data every subsequent step stores `INT32_MAX = 2147483647` — it saturates rather than wrapping. Readers then return the same frozen timestep for the whole tail of the run, silently, with no error.

From a 4096², 3072 rank Lundquist run, 101 dumps:

```
offsets[:4]  = [0, 34521088, 69042176, 103563264]        stride 34,521,088 rows/step
offsets[-4:] = [2147483647, 2147483647, 2147483647, 2147483647]

steps 98->99: offset delta 0,        raw PointData slices bit-identical
steps  4->5:  offset delta 34521088, slices differ (max|diff| 4.4e-2)
```

Frozen from step 63 of 101. An 8192² run in the same batch froze at step 16. The payload is intact — `PointData/data` has exactly `NSteps * rows_per_step` rows — only the index is wrong, so affected files can be recovered by reconstructing `offset[i] = i * total_rows // NSteps`.

## The fix

The three datasets are created as `std::int64_t` (`OffsetType`), matching VTK's own `vtkHDFWriter`, which routes every step offset through `vtkIdType` stored as `H5T_STD_I64LE`. Each write site casts explicitly rather than relying on an implicit narrowing conversion. Readers are unaffected — h5py and VTK both widen int32 transparently, so older files stay readable.

`AMRBoxOffset` and `NumberOfAMRBox` have the same defect but have not overflowed yet at our box counts; they are fixed here too rather than left as the same landmine. `CellDataOffset` and `FieldDataOffset` are empty groups with no datasets, so they were never affected.

`HierarchyData::{level_rank_data_size, n_boxes_per_level, level_data_size}` were `int` and summed `box.size() * X_TIMES` into `int`. That is signed overflow at 2^31 rows in a *single* step — the same threshold — so they are widened to `std::size_t`.

## The test

`test_steps_index` runs the existing 40²-cell config for 3 timesteps (~3 s) and asserts two things per level, for every `.vtkhdf` written:

1. **the declared width** of the three datasets is at least 8 bytes. This is the only observable that carries the defect at a size a test can afford to run — no toy run gets near 2^31 rows, so nothing value-based could ever fire.
2. **the step index rebuilt from the file's own geometry**: `data_offset[step]` must equal the running total of `prod(up_d - lo_d + 2) * X_TIMES` over that step's `AMRBox` rows, `box_offset[step]` the running total of `NumberOfAMRBox`, and both must close exactly on `PointData/data.shape[0]` and `AMRBox.shape[0]`. Exact equality, not monotonicity — a level absent at a step legitimately repeats its offset.

Verified both ways by rebuilding across the change:

| writer | `test_steps_index` |
| --- | --- |
| before | **FAILED** — `4 not greater than or equal to 8 : EM_B.vtkhdf Level0/AMRBoxOffset is int32, too narrow past INT32_MAX` |
| after | **OK**, 3.1 s |

Full file green after the fix (`Ran 2 tests in 4.114s / OK`). Build gate passed hybrid-only, MHD-only and coupled permutations.

## Note for reviewers

Existing dumps stay broken and need the read-side reconstruction. Separately, the vtkhdf reader (`pyphare/.../hierarchy/fromvtkhdf5.py`, which lives on a branch, not on master) reads these offsets with no consistency check at all — that is how this went unnoticed through a full production batch, and it is worth an assertion there independently of this fix.

https://claude.ai/code/session_01QgGaJc1S4JMwbsdtq2X9NU

### DIFF
--- a/src/diagnostic/detail/vtkh5_type_writer.hpp
+++ b/src/diagnostic/detail/vtkh5_type_writer.hpp
@@ -121,14 +121,16 @@ template<typename Writer>
 class H5TypeWriter : public PHARE::diagnostic::TypeWriter
 {
     using FloatType = std::conditional_t<PHARE_DIAG_DOUBLES, double, float>;
-    // VTKHDF step offsets/counts are 64 bit, as in VTK's own writer (H5T_STD_I64LE)
-    //  32 bits saturates for dumps of more than 2^31 rows of point data
-    using OffsetType                     = std::int64_t;
-    using HierData                       = HierarchyData<Writer::dimension>;
-    using ModelView                      = Writer::ModelView;
-    using physical_quantity_type         = ModelView::physical_quantity_type;
-    using Attributes                     = Writer::Attributes;
-    std::string static inline const base = "/VTKHDF";
+    // on disk type of the Steps/ offsets, mandated by VTK: its own writer stores them as
+    //  vtkIdType / H5T_STD_I64LE. In memory the counts stay std::size_t, which is what
+    //  box.size() and HighFive getDimensions/select/resize give and take. 32 bits here
+    //  saturates at INT32_MAX for dumps of more than 2^31 rows of point data.
+    using VTKOffsetType                        = std::int64_t;
+    using HierData                             = HierarchyData<Writer::dimension>;
+    using ModelView                            = Writer::ModelView;
+    using physical_quantity_type               = ModelView::physical_quantity_type;
+    using Attributes                           = Writer::Attributes;
+    std::string static inline const base       = "/VTKHDF";
     std::string static inline const level_base = base + "/Level";
     std::string static inline const step_level = base + "/Steps/Level";
     auto static inline const level_data_path
@@ -390,9 +392,9 @@ void H5TypeWriter<Writer>::VTKFileInitializer::initFileLevel(int const ilvl)
     auto const lvl = std::to_string(ilvl);
 
     h5file.create_resizable_2d_data_set<int, boxValsIn3D>(level_base + lvl + "/AMRBox");
-    h5file.create_resizable_1d_data_set<OffsetType>(step_level + lvl + "/AMRBoxOffset");
-    h5file.create_resizable_1d_data_set<OffsetType>(step_level + lvl + "/NumberOfAMRBox");
-    h5file.create_resizable_1d_data_set<OffsetType>(step_level + lvl + "/PointDataOffset/data");
+    h5file.create_resizable_1d_data_set<VTKOffsetType>(step_level + lvl + "/AMRBoxOffset");
+    h5file.create_resizable_1d_data_set<VTKOffsetType>(step_level + lvl + "/NumberOfAMRBox");
+    h5file.create_resizable_1d_data_set<VTKOffsetType>(step_level + lvl + "/PointDataOffset/data");
 
     auto level_group = h5file.file().getGroup(level_base + lvl);
     if (!level_group.hasAttribute("Spacing"))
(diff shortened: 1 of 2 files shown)


UCaromel:
I think the visualisation bug i saw and talked about a few times on bigger runs (type run0055) on paraview might actually also be this issue. Wrong offsets making the actual the data look very weird. This is consistent with the data looking fine at the first time per file on these, and the fact that the offsets are per level could explain also why L1 and L2 looked fine while L1 had issues -> much less patches on these finer levels.

tests/simulator/test_vtk_diagnostics.py
```diff
@@ -191,6 +205,64 @@ def test_dump_diags(self, ndim, interp, simInput):
             except ModuleNotFoundError:
                 print("WARNING: vtk python module not found - cannot make plots")
 
+    @data(*permute_multistep({}))
+    @unpack
+    def test_steps_index(self, ndim, interp, simInput):
```
nicolasaunai: in a few months I'm not sure that we'll understand why this test. 
Can we make sure from the name and its code that what it tests and why becomes "obvious"

nicolasaunai:
> I think the visualisation bug i saw and talked about a few times on bigger runs (type run0055) on paraview might actually also be this issue. Wrong offsets making the actual the data look very weird. This is consistent with the data looking fine at the first time per file on these, and the fact that the offsets are per level could explain also why L1 and L2 looked fine while L1 had issues -> much less patches on these finer levels.

YES !!  I was going to suggest this indeed. These were typical of wrong offsets... easy to test if you have the data still

src/diagnostic/detail/vtkh5_type_writer.hpp
```diff
@@ -110,15 +111,19 @@ struct HierarchyData
...
 
 
 template<typename Writer>
 class H5TypeWriter : public PHARE::diagnostic::TypeWriter
 {
-    using FloatType                      = std::conditional_t<PHARE_DIAG_DOUBLES, double, float>;
+    using FloatType = std::conditional_t<PHARE_DIAG_DOUBLES, double, float>;
+    // VTKHDF step offsets/counts are 64 bit, as in VTK's own writer (H5T_STD_I64LE)
+    //  32 bits saturates for dumps of more than 2^31 rows of point data
+    using OffsetType                     = std::int64_t;
```
PhilipDeegan: is this not a bit of a disparity if you set the vectors to `std::size_t` which is `unsigned`?

UCaromel: I think the offsets are always unsigned, the int64_t is only for vtkhdf. There should be a single static_cast between the 2.

PhilipDeegan: `9223372036854775807` is pretty large [I guess](https://coliru.stacked-crooked.com/a/1a1ededc628d902e) ;)

PhilipDeegan (approved):

nicolasaunai (approved):

src/diagnostic/detail/vtkh5_type_writer.hpp
UCaromel: One can only hope this is enought 🙏

UCaromel:
Same data, before the offset fix:
image
After the offset fix:
image

PR #1295, opened 2026-07-28
pressure tensor schedule issue

PhilipDeegan:
Demonstrate pressure tensor schedule bug with two simulations, one restarted to flush the schedules, and one not

fix attempt uses a weak_ptr to the patch_level so if it's stale we just make a new schedule

[build showing test failure on first commit ](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_BuildGithubPrGccSs/95457?buildTab=log&linesState=1733&logView=flowAware&focusLine=1996)

### DIFF
--- a/src/diagnostic/diagnostic_model_view.hpp
+++ b/src/diagnostic/diagnostic_model_view.hpp
@@ -244,31 +244,29 @@ protected:
         {
             using PlusEqualsOp = core::PlusEquals<typename VecField::value_type>;
 
-            auto const curLevel = hierarchy.getPatchLevel(ilvl);
-            auto it              = MTschedules.find(ilvl);
-            // a regrid replaces the PatchLevel object at ilvl with a new one, even
-            // though the level number is unchanged, so an ilvl-only cache is not
-            // enough to detect staleness; compare against the level the cached
-            // schedule was actually built from.
-            bool const stale = it != MTschedules.end() and it->second.level.lock() != curLevel;
+            auto const level   = hierarchy.getPatchLevel(ilvl);
+            auto schedule_iter = MTschedules.find(ilvl);
+            auto const create_schedule
+                = schedule_iter == MTschedules.end() or schedule_iter->second.level.lock() != level;
 
-            if (it == MTschedules.end() or stale)
-                it = MTschedules
-                         .insert_or_assign(
-                             ilvl,
-                             Entry{curLevel,
-                                   MTalgo->createSchedule(
-                                       curLevel, 0,
-                                       std::make_shared<amr::FieldBorderOpTransactionFactory<
-                                           typename Super::TensorFieldData_t, PlusEqualsOp>>())})
-                         .first;
+            if (create_schedule)
+                schedule_iter
+                    = MTschedules
+                          .insert_or_assign(
+                              ilvl,
+                              Entry{level,
+                                    MTalgo->createSchedule(
+                                        level, 0,
+                                        std::make_shared<amr::FieldBorderOpTransactionFactory<
+                                            typename Super::TensorFieldData_t, PlusEqualsOp>>())})
+                          .first;
 
-            return *it->second.schedule;
+            return *schedule_iter->second.schedule;
         }
 
         struct Entry
         {
-            std::weak_ptr<SAMRAI::hier::PatchLevel> level;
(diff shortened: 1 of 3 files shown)


src/diagnostic/diagnostic_model_view.hpp
```diff
@@ -243,18 +243,38 @@ class ModelView<Hierarchy, Model, std::enable_if_t<solver::is_hybrid_model_v<Mod
...
-                                  typename Super::TensorFieldData_t, PlusEqualsOp>>()));
-            return *MTschedules[ilvl];
+
+            auto const curLevel = hierarchy.getPatchLevel(ilvl);
+            auto it              = MTschedules.find(ilvl);
+            // a regrid replaces the PatchLevel object at ilvl with a new one, even
+            // though the level number is unchanged, so an ilvl-only cache is not
+            // enough to detect staleness; compare against the level the cached
+            // schedule was actually built from.
+            bool const stale = it != MTschedules.end() and it->second.level.lock() != curLevel;
```
PhilipDeegan: https://en.cppreference.com/cpp/memory/weak_ptr/lock

src/diagnostic/diagnostic_model_view.hpp
```diff
@@ -243,18 +243,38 @@ class ModelView<Hierarchy, Model, std::enable_if_t<solver::is_hybrid_model_v<Mod
...
-                              hierarchy.getPatchLevel(ilvl), 0,
-                              std::make_shared<amr::FieldBorderOpTransactionFactory<
-                                  typename Super::TensorFieldData_t, PlusEqualsOp>>()));
-            return *MTschedules[ilvl];
+
+            auto const curLevel = hierarchy.getPatchLevel(ilvl);
+            auto it              = MTschedules.find(ilvl);
+            // a regrid replaces the PatchLevel object at ilvl with a new one, even
+            // though the level number is unchanged, so an ilvl-only cache is not
+            // enough to detect staleness; compare against the level the cached
```
nicolasaunai: //  so an ilvl-only cache is not
            // enough to detect staleness; compare against the level the cached
            // schedule was actually built from.

what?

PhilipDeegan: the master schedules are per level

https://github.com/PHAREHUB/PHARE/blob/master/src/diagnostic/diagnostic_model_view.hpp#L252

which we keep once created, so that is the ilvl cache

tests/simulator/refinement/test_regridding.py
```diff
@@ -0,0 +1,281 @@
...
+
+time_step = 0.005
+time_step_nbr = 4
+final_time = time_step * time_step_nbr
+
+# ModelView::MTAlgo::getOrCreateSchedule (src/diagnostic/diagnostic_model_view.hpp)
+# caches the border-overlap schedule it uses to compute the (per population) momentum/
+# pressure tensor by SAMRAI level number only, and never rebuilds it when that level is
+# regridded (unlike the regular messenger pipeline, see multiphysics_integrator.hpp).
+# L2 is regridded by tagging on essentially every substep of its parent level, so a schedule
```
nicolasaunai: every 2 steps

PhilipDeegan: no I think the point is that L2 will be regridded during an L1 substep, generally

nicolasaunai (approved):

src/diagnostic/diagnostic_model_view.hpp
```diff
@@ -244,31 +244,29 @@ class ModelView<Hierarchy, Model, std::enable_if_t<solver::is_hybrid_model_v<Mod
...
-                                   MTalgo->createSchedule(
-                                       curLevel, 0,
-                                       std::make_shared<amr::FieldBorderOpTransactionFactory<
-                                           typename Super::TensorFieldData_t, PlusEqualsOp>>())})
-                         .first;
-
-            return *it->second.schedule;
+            auto const level   = hierarchy.getPatchLevel(ilvl);
+            auto schedule_iter = MTschedules.find(ilvl);
+            auto const create_schedule
```
nicolasaunai, line 249: I would rather read  `if (needNewSchedule)` :p

nicolasaunai (approved):

PR #1297, opened 2026-07-29
pharein: default nesting_buffer to 1, only apply/validate it when AMR is used

### LINKED ISSUE #1196: nesting buffer default is 0
should it not be 1 now?

https://github.com/PHAREHUB/PHARE/blob/4b2c3946e035e37727bccf9c7350abe1f2feb399/pyphare/pyphare/pharein/simulation.py#L583

nicolasaunai:
Fixes #1196.

**Problem:** `nesting_buffer` (minimum coarse-cell gap between a level's border and any refined patch) defaulted to `0`. SAMRAI itself defaults `proper_nesting_buffer` to `1` and only accepts `0` with a warning ([`PatchHierarchy.cpp:57,306-330`](https://github.com/PHAREHUB/PHARE/blob/master/subprojects/samrai/source/SAMRAI/hier/PatchHierarchy.cpp#L57)) — `0` is known to let L1 patches touch periodic borders.

**Changes:**
- `nesting_buffer` now defaults to `1`, and must be `>= 1` whenever the simulation has refinement levels.
- It is only meaningful when there *is* AMR (`refinement_boxes` set, or `max_nbr_levels > 1`). Otherwise it's left unset and not sent to SAMRAI at all — same pattern already used for `smallest_patch_size`/`largest_patch_size`, and consistent with SAMRAI itself ignoring the value when `max_levels == 1`.
- Dropped a check comparing `nesting_buffer` to `cells - largest_patch_size`: not derived from any real SAMRAI constraint (the actual geometric invariant is enforced per-box elsewhere, in `check_refinement_boxes`), and it broke as soon as `largest_patch_size == cells`.
- Adjusted the handful of `test_validation.py` cases that only worked with a `0`-cell nesting gap. One case relied on a finer box spanning the touching boundary of two same-level sibling boxes — that's geometrically impossible once any positive `nesting_buffer` is required, so it moved from valid to invalid.

**Tested:** `py3_validation` (60/60) and `py3_diagnostics` (28/28, including `test_dump_diags`, the test that broke when `nesting_buffer` was first bumped to 1) both pass.

### DIFF
--- a/pyphare/pyphare/pharein/simulation.py
+++ b/pyphare/pyphare/pharein/simulation.py
@@ -594,7 +594,7 @@ def check_nesting_buffer(ndim, **kwargs):
         return None
 
     # SAMRAI itself defaults proper_nesting_buffer to 1 and warns when it is 0
-    # (subprojects/samrai/source/SAMRAI/hier/PatchHierarchy.cpp:57,306-330)
+    # (subprojects/samrai/source/SAMRAI/hier/PatchHierarchy.cpp)
     nesting_buffer = phare_utilities.np_array_ify(kwargs.get("nesting_buffer", 1), ndim)
 
     if nesting_buffer.size != ndim:
--- a/pyphare/pyphare/pharein/simulation.py
+++ b/pyphare/pyphare/pharein/simulation.py
@@ -581,9 +581,15 @@ def check_refinement(**kwargs):
 
 def check_nesting_buffer(ndim, **kwargs):
     refinement_boxes = kwargs.get("refinement_boxes")
-    has_amr = (
-        refinement_boxes is not None and len(refinement_boxes) > 0
-    ) or kwargs.get("max_nbr_levels", 1) > 1
+    has_refinement_boxes = refinement_boxes is not None and len(refinement_boxes) > 0
+
+    if check_refinement(**kwargs) == "boxes":
+        # check_refinement_boxes() normalizes to (None, 1) whenever
+        # refinement_boxes is empty, ignoring any max_nbr_levels the user
+        # passed in that case, so has_amr must follow the same rule here.
+        has_amr = has_refinement_boxes
+    else:
+        has_amr = kwargs.get("max_nbr_levels", 1) > 1
 
     if not has_amr:
         if "nesting_buffer" in kwargs:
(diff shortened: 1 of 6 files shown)


UCaromel (approved):

src/amr/wrappers/hierarchy.hpp
```diff
@@ -382,8 +382,11 @@ auto patchHierarchyDatabase(PHARE::initializer::PHAREDict const& amr)
     auto maxLevelNumber = amr["max_nbr_levels"].template to<int>();
     hierDB->putInteger("max_levels", maxLevelNumber);
 
-    std::vector<int> nesting_buffer = amr["nesting_buffer"].template to<std::vector<int>>();
-    hierDB->putIntegerVector("proper_nesting_buffer", nesting_buffer);
+    if (amr.contains("nesting_buffer"))
+    {
+        std::vector<int> nesting_buffer = amr["nesting_buffer"].template to<std::vector<int>>();
```
PhilipDeegan: -> `std::vector<int> nesting_buffer = amr["nesting_buffer"];`

you can drop the `.template ...` if you're setting the type on the left hand side now

pyphare/pyphare/pharein/simulation.py
```diff
@@ -580,16 +580,33 @@ def check_refinement(**kwargs):
...
+    if not has_amr:
+        if "nesting_buffer" in kwargs:
+            print(
+                f"WARNING, nesting_buffer({kwargs['nesting_buffer']}) has no effect"
+                + " on a simulation with no refinement levels, ignoring"
+            )
+        return None
+
+    # SAMRAI itself defaults proper_nesting_buffer to 1 and warns when it is 0
+    # (subprojects/samrai/source/SAMRAI/hier/PatchHierarchy.cpp:57,306-330)
```
PhilipDeegan: probably don't include the line numbers as they may not last if samrai changes something

PhilipDeegan (approved):

pyphare/pyphare/pharein/simulation.py
```diff
@@ -580,7 +580,22 @@ def check_refinement(**kwargs):
...
+    nesting_buffer = phare_utilities.np_array_ify(kwargs.get("nesting_buffer", 1), ndim)
```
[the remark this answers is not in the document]
nicolasaunai: yes

PR #1277, opened 2026-07-23
doc: overhaul pharein documentation

nicolasaunai:
addresses #1273 and limited to pharein

PR #1300, opened 2026-07-30
release: bump SAMRAI and HighFive pins for the upcoming 0.1.0 release

nicolasaunai:
## Summary

In preparation for PHARE's first release (0.1.0), reviewed and updated the pinned dependency versions in `res/cmake/release.cmake`.

- **SAMRAI**: `v-4-3-0` → `v-4-5-1`
- **HighFive**: `v3.0.0-beta2` → `v3.3.0` (moves off a beta onto a stable tag)
- **pybind11**: left at `v2.13.6`. The latest upstream is a major version (`v3.x`), which is a real API migration — deferring that to its own PR rather than bundling it into a release-prep change.
- **cppdict**: left at `v1.0.0` (already the latest tag; it's our own LaboratoryOfPlasmaPhysics library).

Note: these pins only take effect when `release.ver.txt` exists (see `CMakeLists.txt`); normal development builds are unaffected and continue to float on `develop`/`master`/`main` of each dependency as before.

## Why

`release.cmake` had never actually been exercised — PHARE has never cut a release, and CI never builds against these pinned versions (only against the floating branches). So the existing pins were both stale and untested against current PHARE code.

## Testing

Verified each bump independently (isolated clone per dependency, to avoid disturbing the shared dev checkouts under `subprojects/`), with a full `cmake --build` + `ctest` run against `release.ver.txt` present:

| Dependency | Version | Build | Tests |
|---|---|---|---|
| SAMRAI | v-4-5-1 | clean, 0 errors | 76/76 passed |
| HighFive | v3.3.0 | clean, 0 errors | 76/76 passed |

Both full test suites (`ctest -j$(nproc)`, all 76 registered tests including the HighFive-exercising `py3_restarts`, `py3_diagnostics`, `test-diagnostics_*`) passed with zero failures.

## Release note candidate (v0.1.0)

Draft for the GitHub Release once this is merged and the `0.1.x` branch is cut:

> # PHARE 0.1.0
>
> First tagged release of PHARE — a parallel multi-formalism platform for simulating space plasmas, with patch-based Adaptive Mesh Refinement (AMR), built on LLNL's SAMRAI.
>
> ## What's included
> - AMR Hybrid PIC
>   - isothermal electrons
>   - 1st, 2nd, 3rd order b-spline interpolation kernels
> - AMR Resistive & Hall MHD, Godunov finite volume code
> - All formalisms available in 1D, 2D, 3D
>
> ## Pinned dependencies for this release
> - SAMRAI v-4-5-1
> - HighFive v3.3.0
> - pybind11 v2.13.6
> - cppdict v1.0.0
>
> (These are frozen for this release via `release.ver.txt`; normal development on `master` continues to track upstream `develop`/`main` branches of these dependencies.)
>
> ## Known issues
> See `ISSUES.TXT` for a list of known environment-specific issues (OpenMPI symbol resolution, MPI4PY/C++ MPI init ordering, static-library linking caveats).
>
> ## Reporting problems
> Run `python3 tools/report.py` to generate a diagnostic bundle, and open an issue at https://github.com/PHAREHUB/PHARE/issues/new.

https://claude.ai/code/session_01SjgiFbY4nkCp7jihVsYFcn


PhilipDeegan:
using the last cppdict release will not compile as it does not include [this ](https://github.com/LaboratoryOfPlasmaPhysics/cppdict/pull/22)

we should cut a release of cppdict so we can upgrade it to C++20 anyway

PhilipDeegan:
not sure the last highfive release will work either but I guess we will find out

.gitignore
```diff
@@ -1,5 +1,6 @@
 bin
 build
+.worktrees
```
PhilipDeegan: ?

PhilipDeegan:
you could add `release.ver.txt` and find out

nicolasaunai:
> using the last cppdict release will not compile as it does not include [this ](https://github.com/LaboratoryOfPlasmaPhysics/cppdict/pull/22)
> 
> we should cut a release of cppdict so we can upgrade it to C++20 anyway

maybe there should be more release there then indeed

nicolasaunai:
> you could add `release.ver.txt` and find out

find out what

.gitignore
nicolasaunai: ??

nicolasaunai:
> not sure the last highfive release will work either but I guess we will find out

builds and all tests pass

PhilipDeegan:
>  these pins only take effect when release.ver.txt exists

.gitignore
PhilipDeegan: ‽‽

UCaromel:
> not sure the last highfive release will work either but I guess we will find out

Highfive had issues not so long ago right ? I would guess fixed if they made a major release and the tests pass

.gitignore
nicolasaunai: 🫣

nicolasaunai:
Pushed `release.ver.txt` temporarily to let CI actually exercise the release pins (removing it again before merge). Result: CI failed, but not on cppdict — worth explaining why.

**Failure:**
```
src/python3/cpp_etc.cpp:79:40: error: no member named 'smart_holder' in namespace 'pybind11'
```
Everything else built cleanly up to that point (including code compiled against cppdict headers — no cppdict errors surfaced).

**Root cause:** our normal CI (`cmake_ubuntu.yml`) runs in `ghcr.io/pharchive/phare_dep/ubuntu:24.04` with `-DSAMRAI_ROOT=/usr/local` and a pre-activated venv. That means SAMRAI and pybind11 normally come from that image's pre-built environment, not from a GitHub clone — `find_package(SAMRAI PATHS /usr/local)` and `find_package(pybind11 ... CONFIG QUIET)` both short-circuit `release.cmake`/`dep/*.cmake` entirely in the default configuration. The pre-installed pybind11 in that image must be a `smart_holder`-patched build, since `smart_holder` only ever existed on pybind11's `archive/smart_holder` branch and was never merged into `pybind/pybind11` mainline, at any tag or on current `master`.

The moment `release.ver.txt` set `PHARE_PYBIND11_VERSION`, the build fell through to cloning the public `pybind/pybind11` repo instead of using the image's patched copy — and the public repo doesn't have `smart_holder` anywhere. So this isn't something the SAMRAI/HighFive bump broke — it looks like the pybind11 pin in `release.cmake` (`v2.13.6`, untouched by this PR) has never actually been buildable from a clean clone, it just was never tested until now since nothing normally exercises `release.cmake`.

Practical consequence: this CI run also didn't test our SAMRAI bump either, since `-DSAMRAI_ROOT=/usr/local` bypasses the pin regardless. The SAMRAI/HighFive validation we have is from local isolated-clone builds (76/76 tests each), not this CI run.

Still open: cppdict at `v1.0.0` — the build died before we got confirmation either way, though nothing cppdict-related failed in what did run.

Does anyone know where the image's pybind11 build comes from (Dockerfile/build recipe for `ghcr.io/pharchive/phare_dep`)? We'd need to point `release.cmake`'s pybind11 pin at whatever fork/branch that actually is, rather than the public tag, for the release mechanism to work at all.

res/cmake/release.cmake
```diff
@@ -3,14 +3,14 @@
 #  KEEP SIMPLE!
 #
 
-# https://github.com/LLNL/SAMRAI/releases/tag/v-4-3-0
-SET(PHARE_SAMRAI_VERSION "v-4-3-0")
+# https://github.com/LLNL/SAMRAI/releases/tag/v-4-5-1
+SET(PHARE_SAMRAI_VERSION "v-4-5-1")
 
 # https://github.com/pybind/pybind11/releases/tag/v2.13.6
 SET(PHARE_PYBIND11_VERSION "v2.13.6")
```
PhilipDeegan: you probably want to update this to `v3.0.4`

nicolasaunai:
should wait for:
- #1271 
- #1278 
- #1277

PR #1309, opened 2026-08-05
diagnostics/resource manager multimodel

### LINKED ISSUE #1197: Lazy-evaluator collection for additional diagnostics
# Current way to add diags
Right now the pattern to add a new diag `X` is to add a `X_diag_` field in the `model` and the `diagnostic_model_view`, and to implement a corresponding compute function that gets duplicated for each diagnostic type (pahreh5, vtkhdf).

# Proposition
Instead, store a registry of lazy evaluators, one per diagnostic. At write-time, the writer calls each registered evaluator, writes the result into a shared scratch field, then flushes to disk.

# Advantages
- no duplication of diag computation in each type of diagnostic.
- decoupling of diag writing and diag computation.
- single memory allocation for the scratch field, for all current and future diags.

# Drawbacks / open questions
- should this be implemented on the python side rather than cpp one ?

### LINKED ISSUE #1219: Possible MHD diagnostic redundancies
We now how the following

```
    VecField V_diag_{"diagnostics_V_", core::MHDQuantity::Vector::V};
    Field P_diag_{"diagnostics_P_", core::MHDQuantity::Scalar::P};
```

which are samrai allocated per patch fields - it's possible we can just use the existing temporary patch fields and not need these

### LINKED ISSUE #1275: VTKHDF emergency dump with more than 2 AMR levels is messed up
In paraview dump exist, but nothing can be vizualized. Can be observed on this failing case : [blast.py](https://github.com/user-attachments/files/30309344/blast.py)

PhilipDeegan:
Master has a resource manager per model which doesn't really make a lot of sense as it deals with SAMRAI static global patch data registry, and also master diagnostics is only really configured for one model

introduces easily extendible temporary/runtime resources using variants
drops pressure tensor from ions and ion pops given it's only used for diagnostics, I think @UCaromel may use it for coupling so tbc

diagnostics needsCompute is now an noop as all computed diagnostics are on an as needed basis, per active timestamp for computed diagnostics, I guess in future it may be used for time averaging or somesuch

closes #1275
closes #1219
closes #1197


src/amr/resources_manager/resources_manager.hpp
```diff
@@ -645,6 +580,14 @@ namespace amr
             }
         }
 
+        // every live instance of this exact ResourcesManager specialization registers itself
+        // here, so ALL_IDS()/registerForRestarts() can aggregate across the models sharing it
+        NO_DISCARD static auto& instances_()
+        {
+            static std::vector<std::map<std::string, ResourcesInfo>*> instances;
```
PhilipDeegan: possibly can just be a class member

PR #1314, opened 2026-08-06
Ruff errors assessed

PhilipDeegan:
New version of ruff has many more checks active now.

before (e.g. master)

```
Found 717 errors.
```

after

```
Found 189 errors.
```


PhilipDeegan:
probably merge after #1315

Issue #1127, opened 2026-01-14
boundary conditions in PHARE


nicolasaunai:
implement non periodic boundary conditions in the MHD part of PHARE

- [ ] document the python interface
- [ ] document which are the available boundary conditions

- [ ] 1D
- [ ] 2D
- [ ] 3D

PR #1318, opened 2026-08-31
Minimal boundary conditions

igirault:
Adds base mechanism for domain-edge boundary conditions applied to fields, and make it possible to use it by the MHD model.

## Issue

 Relates to #1127.

## Usage 

To be backward-compatible, the non-periodic directions must be indicated as "physical" in the "boundary_types" Simulation's kwarg lis. Up to now this list was only allowed to specify "periodic" for each direction. Periodic directions remains the default behavior. I think "boundary_types" should be renamed as "direction_types" or "is_periodic" (in this last case should becomes a list of bools)

For non-periodic directions, boundary_conditions shall be specified, with this kind of SImulation's kwargs:

```python
    boundary_types="physical",  # if scalar, all directions types are set with this value
    boundary_conditions={  #  assuming the simulation is 1D, we must prescribe xlower and xupper as boundary conditions
        "xlower": {"type": "super-magnetofast-inflow",
                   "data": {"velocity": [U, 0.0, 0.0],
                            "density": 1.0,
                            "pressure": 1.0,
                            "B": [Bx, By, 0.0]}},
        "xupper": {"type": "open"},
```

Supported boundary condition types: `none`, `reflective`, `super-magnetofast-inflow`, `open`.

## Design

We resort to the SAMRAI mechanism for enforcing boundary conditions: overriding the SAMRAI's `RefinePatchStrategy` class by a `FieldRefinePatchStrategy`. Its main function, `setPhysicalBoundaryConditions` defers the work to newly created `core` classes, mostly `BoundaryManager` and `IFieldBoundaryCondition`. This last one is abstract, and is implemented for each elementary field boundary condition (Neumann, Dirichlet, ...). The `BoundaryManager` belongs to the model, and basically associates one `IFieldBoundaryCondition` per physical quantity and per physical/non-periodic boundary.

More detail will be added here.

## Tests

- C++ unit tests per individual field boundary condition type.
- pyphare validation tests for the configuration dict.
- Functional: mhd_shock_with_super_magnetofast_inflow, and a two-level mhd_harris_with_boundaries exercising a refined level at a physical boundary.

## To be done

Some point are left open:
- concepts for GridLayout, Scalar Vs Vector FieldData.
- usage of BoxIterator (and its reversed), shall be removed when new box iterations are available.
- handling of temporary refinement patches near boundaries/
- some utility definition (`ParseDimXYZType`) moved, or new ones created  (`neighbor` in `point.hpp`)


src/amr/messengers/mhd_messenger.hpp
```diff
@@ -405,23 +404,16 @@ namespace amr
             for (auto& patch : level)
             {
                 auto dataOnPatch = resourcesManager_->setOnPatch(
-                    *patch, mhdModel.state.rho, mhdModel.state.V, mhdModel.state.P,
```
igirault, line 408: unrelated deletions, but legit

src/core/utilities/box/box.hpp
```diff
@@ -19,6 +19,8 @@ namespace PHARE::core
 template<typename Type, std::size_t dim>
 class box_iterator;
 
+template<typename Type, std::size_t dim>
```
igirault, line 22: must me removed once new box iterator is available

tests/functional/mhd_shock_with_super_magnetofast_inflow/mhd_shock.py
```diff
@@ -0,0 +1,268 @@
...
+        if self.simulator is not None:
+            self.simulator.reset()
+        self.simulator = None
+        ph.global_vars.sim = None
+
+    def test_run(self):
+        # self.register_diag_dir_for_cleanup(diag_dir)
+        Simulator(config()).run().reset()
+        print_case_info()
+        assert_inflow_holds_left_state(diag_dir)
```
PhilipDeegan: we've started executing functional tests in dry run mode on TC (why [this job](https://hephaistos.lpp.polytechnique.fr/teamcity/buildConfiguration/Phare_Phare_BuildGithubPrGccSs/96449?buildTab=log&linesState=1750&logView=flowAware&focusLine=2314) is failing)

so you probably want to change this to something like 


```
sim = config()
Simulator(sim).run().reset()
if not sim.dry_run and cpp.mpi_rank() == 0:
    print_case_info()
    assert_inflow_holds_left_state(diag_dir)
```

Issue #1283, opened 2026-07-24
Reduce number of ghosts in MHD.


UCaromel:
### Why

Number of ghosts in MHD is higher than i needs to be. Right now it is 6 with WENO-Z, which is 2 more than what we need, even with the current way of computing the current density J (Ampere on the ghostbox - 1, preparation for boundary conditions to avoid having to re-derive J from B analytically for every boundary condition type). 

### What

Three potential way to address this: 
- compute the fluxes only on the domain box, which a sub-box of where the reconstruction are done (transverse directions get reconstructed of the reconstructions stencil width on the ghost, to prepare the coefficient for UCT). Need to assess if this can be done without the need for extra storage for the L and R states, but i expect it probably could.
- Use the ct electric field in the poynting flux: (https://github.com/UCaromel/PHARE/tree/feature/ct-energy-flux-port), this way the laplacian evaluation never has to be done on the reconstructed faces, but only ever on the edge centered electric field. 
- Use whistler waves in the riemann fan instead of hyper resistivity to stabilise the hall term.

nicolasaunai:
Could decrease the probability same level patch overlaps  (#1284) occur.

nicolasaunai:
Could decrease the probability same level patch overlaps  (#1284) occur.

PR #1319, opened 2026-08-31
Feature/mhd edge native dissipative flux (reduce ghost number constraint).

UCaromel:
## Issue
#1283

## What this implements

`nbrGhostsFromReconstruction<R>()` was `roundUpToEven(R + 2)`, so WENOZ/MP5 MHD builds reserve **6** field ghost layers. The extra layer was bought by one thing: with hyper-resistivity on, laplJ was computed from reconstructed values, which required +1 cell in each directions.

The non-ideal (resistive **and** hyper-resistive) flux contributions now come from J at its native **edge** location, so the Laplacian never reaches past layer 2, the `+1` shell is gone, and the constant becomes `roundUpToEven(R + 1)`: WENOZ/MP5 **6 → 4**, Linear/WENO3/Constant unchanged (rounding absorbs it).

## Design

`E_diss = eta*J - coef*laplacian(J)` is formed on the edge, multiplied **there** by the edge-projected transverse B — projection does not commute with multiplication, so the product must be formed before projecting — and the product is projected to the face. The flux's non-ideal terms are then the face projection of the *same* edge quantities the CT already uses to advance B, where before the two used different discretisations of ∇²J times different B's.

## Tests

Harris bitwise identical in ideal. Differences in the order of O(1e-5) with spatial hyper resistivity on harris at 0.4 dx. Expected since different numerics, lowering at O(dx^2) (hyper constant) or O(dx^4) (hyper spatial). Expected as second order representation of the laplacian (multiplied by dx^2 in spatial case).

### DIFF
--- a/src/amr/solvers/time_integrator/compute_fluxes.hpp
+++ b/src/amr/solvers/time_integrator/compute_fluxes.hpp
@@ -60,7 +60,7 @@ public:
             TimeSetter{level, model, newTime}(state.B, state.J);
         }
 
-        FVMethod_t{level, model, fVMethodInfo_}(fvm_, ct_, state, fluxes, newTime);
+        FVMethod_t{level, model, fVMethodInfo_}(ct_, state, fluxes, newTime);
 
         // unecessary if we decide to store both primitive and conservative variables
         ToConservativeConverter_t{level, model}(state, to_conservative_gamma_, newTime);
@@ -68,15 +68,10 @@ public:
         ConstrainedTransport_t{level, model, constrainedTransportInfo_}(ct_, state);
     }
 
-    void registerResources(MHDModel& model)
-    {
-        model.resourcesManager->registerResources(fvm_);
-        model.resourcesManager->registerResources(ct_);
-    }
+    void registerResources(MHDModel& model) { model.resourcesManager->registerResources(ct_); }
 
     void allocate(MHDModel& model, auto& patch, double const allocateTime) const
     {
-        model.resourcesManager->allocate(fvm_, patch, allocateTime);
         model.resourcesManager->allocate(ct_, patch, allocateTime);
     }
 
@@ -84,9 +79,7 @@ private:
     FVMethodInfo_t fVMethodInfo_;
     ConstrainedTransportInfo_t constrainedTransportInfo_;
 
-    // Ampere_t ampere_;
-    core::GodunovState<VecField, Equations_t> fvm_{};
-    core::UpwindConstrainedTransportState<VecField, Hall, Resistivity> ct_{};
+    core::UpwindConstrainedTransportState<VecField, Hall, Resistivity, HyperResistivity> ct_{};
     // ToPrimitiveConverter_t to_primitive_;
     // ToConservativeConverter_t to_conservative_;
     double to_primitive_gamma_;
--- a/src/core/numerics/MHD_equations/MHD_equations.hpp
+++ b/src/core/numerics/MHD_equations/MHD_equations.hpp
@@ -90,21 +90,6 @@ public:
         return f;
     }
 
-    // template<auto direction>
-    // auto compute(auto const& u, auto const& J, auto const& LaplJ) const
-    // {
-    //     PerIndex f = compute<direction>(u);
-    //
-    //     if constexpr (Hall)
-    //         hall_contribution_<direction>(u.rho, u.B, J, f.B, f.P);
-    //     if constexpr (Resistivity)
-    //         resistive_contributions_<direction>(eta_, u.B, J, f.B, f.P);
-    //
-    //     resistive_contributions_<direction>(nu_, u.B, -LaplJ, f.B, f.P);
-    //
-    //     return f;
-    // }
-
     template<auto direction>
     void resistive_contributions(auto const& coef, auto const& Bt, auto const& Jt, auto& F_B,
                                  auto& F_Etot) const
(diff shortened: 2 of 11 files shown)


src/amr/solvers/time_integrator/compute_fluxes.hpp
```diff
@@ -60,33 +60,26 @@ class ComputeFluxes
...
 private:
     FVMethodInfo_t fVMethodInfo_;
     ConstrainedTransportInfo_t constrainedTransportInfo_;
 
-    // Ampere_t ampere_;
-    core::GodunovState<VecField, Equations_t> fvm_{};
-    core::UpwindConstrainedTransportState<VecField, Hall, Resistivity> ct_{};
+    core::UpwindConstrainedTransportState<VecField, Hall, Resistivity, HyperResistivity> ct_{};
     // ToPrimitiveConverter_t to_primitive_;
     // ToConservativeConverter_t to_conservative_;
```
PhilipDeegan, line 84: do we need to keep these comments?

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -156,164 +151,147 @@ class Godunov : public GodunovInfo
...
-        auto& Bt = getBt_<direction>(fvm_state);
-
-        Bt(Component::X)(idx) = Bidx.x;
-        Bt(Component::Y)(idx) = Bidx.y;
-        Bt(Component::Z)(idx) = Bidx.z;
+        auto constexpr wps = Stencil();
+        double r           = 0.;
+        for (auto const& wp : wps)
+            r += wp.coef * at_edge(idx + wp.indexes);
+        return r;
```
PhilipDeegan, line 188: could be `return core::sum_from(...);`

https://github.com/PHAREHUB/PHARE/blob/master/src/core/utilities/types.hpp#L269

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -156,164 +151,147 @@ class Godunov : public GodunovInfo
...
-            return fvm_state.bt_y;
-        else if constexpr (direction == Direction::Z)
-            return fvm_state.bt_z;
+        auto const meshSize = layout_.meshSize();
+        if constexpr (dimension == 1)
+            return meshSize[0];
+        else if constexpr (dimension == 2)
+            return std::min({meshSize[0], meshSize[1]});
+        else
+            return std::min({meshSize[0], meshSize[1], meshSize[2]});
```
PhilipDeegan, line 199: `return *std::min_element(...);`

https://en.cppreference.com/cpp/algorithm/min_element

src/amr/solvers/time_integrator/compute_fluxes.hpp
UCaromel: Removed - they were dead placeholders, the converters are constructed inline in operator().

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
UCaromel: Done, now sum_from over the stencil weight points.

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
UCaromel: Done, min_element over meshSize() - the if constexpr dimension ladder is gone.

src/core/numerics/MHD_equations/MHD_equations.hpp
```diff
@@ -90,21 +90,6 @@ class MHDEquations
         return f;
     }
 
-    // template<auto direction>
```
PhilipDeegan, line 93: torm?

UCaromel: rmed

PhilipDeegan (approved):

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -87,6 +81,12 @@ class Godunov : public GodunovInfo
 
     constexpr static auto Hall = Equations::hall;
 
+    // Zero margin by design: ampere computes J on the ghost box shrinked by one, so J is valid on
+    // ghost_width - 1 layers, and the reconstruction reads it out to nghosts on the transverse
+    // grow shell. Any consumer reaching further than that needs one more ghost layer.
```
igirault: this comment can be deleted I think

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -182,48 +175,23 @@ class Godunov : public GodunovInfo
...
             };
 
-            // adding resistive contributions to energy taking advantage of the already computed jt
-            // fluxes for the laplacian computation. This probably doesn't need the grow as the
-            // required quantities for ct are already saved.
-            auto addResistiveContributions = [&](auto doResistiveT, auto doHyperT, auto hyperT) {
+            // Non-ideal (resistive + hyper-resistive) flux contributions, taken from J at its
+            // native edge location: the E_diss x B product is formed on the edge and projected
+            // to the face. This keeps the Laplacian stencil confined to the edge-J reconstruction
+            // width, so it never needs J past the transverse-grow layer.
```
igirault: could this comment be a bit clearer and less claude-like ?

src/core/numerics/godunov_fluxes/godunov_fluxes.hpp
```diff
@@ -244,103 +212,115 @@ class Godunov : public GodunovInfo
...
+    // transverse B component onto the opposite edge, so the E_diss*B product can be formed
+    // pointwise on a single edge before being projected (projection does not commute with
+    // multiplication).
+    template<auto direction, bool Resistivity, bool HyperResistivity, HyperMode hyper,
+             auto EdgeFirstToFace, auto EdgeSecondToFace, auto BSecondToEFirst,
+             auto BFirstToESecond>
+    void non_ideal_face_contribution_(auto const& state, auto const& ct_state,
+                                      MeshIndex<dimension> idx, auto const& Jfirst,
+                                      auto const& Jsecond, auto const& Bfirst, auto const& Bsecond,
+                                      auto const& Bnormal, auto& F_B, auto& F_Etot) const
```
igirault: tasty, looks like fortran

src/core/numerics/MHD_equations/MHD_equations.hpp
```diff
@@ -111,6 +111,37 @@ class MHDEquations
...
+            F_B.z += -projSecond;
+        }
+        if constexpr (direction == Direction::Z)
+        {
+            F_B.x += -projSecond;
+            F_B.y += projFirst;
+        }
+        F_Etot += crossFirst - crossSecond;
+    }
+
```
igirault: I think having this here is wrong. There should not be any stuff related to discretization in this file.

igirault (commented):
Excellent intent we clearly want this. The PR was super good looking until I came accross `godunov_fluxes.hpp` and `MHD_equations.hpp`. Clearly my feeling is that there should be a rewrite there because the code looks complicated, and the function added to `MHD_equations.hpp` does not feel to belong here

PR #1323, opened 2026-09-14
pyphare: say which permutation a missing pybindlibs module needs

UCaromel:
A list of options whose field combination is not in any res/sim permutation file fails with Python's bare import error, naming the module but neither what to do about it nor what the build does contain:

    ModuleNotFoundError: No module named
    'pybindlibs.cpp_2_1_4_SSPRK4_5_WENOZ_None_Rusanov_false_false_false'

Re-raise it with the res/sim line that would build the module, and the list of permutations the loaded build actually has:

    This run needs the build permutation
        2,1,4,SSPRK4_5,WENOZ,None,Rusanov,false,false,false
    Add it to a res/sim permutation file (res/sim/all.txt) and rebuild.
    Permutations currently built:
        2_SSPRK4_5_WENOZ_None_Rusanov_false_false_false
        ...


PhilipDeegan (approved):

pyphare/pyphare/cpp/__init__.py
```diff
@@ -12,7 +12,8 @@
 _libs = {}
 
 
-def simulator_id(sim):
+def _simulator_id_parts(sim):
```
PhilipDeegan, line 15: private functions can be at the bottom as the whole script is read before any function can be called

PhilipDeegan:
we could potentially have something like [this](https://github.com/PhilipDeegan/PHARE/blob/tiling_multithreaded/pyphare/pyphare/cpp/__init__.py#L52) 

where we try to build the lib if missing, pretty handy, but maybe it's not something we want to encourage people to use given it would mean one rank is compiling and the rest are waiting

PhilipDeegan:
there was pending comments on this

igirault:
Ah sorry we were discussing it yesterday and I thought this was good to go

PR #1325, opened 2026-09-16
rm phare_simulator_options.hpp and rework dependency hierarchy for mh…

PhilipDeegan:
phare_simulator_options.hpp was being used in places that had no real reason to know what a simulator was

things have been moved around so that only local headers for enums are required


src/simulator/simulator_def.hpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+};
+
+struct SimOpts
+{
+    std::size_t dimension     = 1;
+    std::size_t interp_order  = 1;
+    std::size_t nbRefinedPart = core::defaultNbrRefinedParts(dimension, interp_order);
+
+    MHDOpts::TimeIntegratorType time_integrator_type = MHDOpts::TimeIntegratorType::Default;
```
UCaromel: Do we think it makes more sense to keep the default being mhdoff as currently on master in SimOpts ?

src/simulator/simulator_def.hpp
```diff
@@ -0,0 +1,122 @@
...
+    bool mhd_enabled    = reconstruction_type != MHDOpts::ReconstructionType::Default;
+
+    // time_integrator_type/riemann_solver_type must agree with reconstruction_type on
+    // off-ness; a member fn keeps SimOpts an aggregate. slope_limiter_type is excluded:
+    // MHDOpts::SlopeLimiterType::None is both its off-sentinel and a legitimate active
+    // choice (e.g. Constant reconstruction pairs with limiter=None while mhd is enabled).
+    constexpr bool mhd_axes_consistent() const
+    {
+        bool const mhd_is_off = !mhd_enabled;
+        return mhd_is_off == (time_integrator_type == MHDOpts::TimeIntegratorType::Default)
```
UCaromel: SlopeLimiter == None or MHDOff ?

PhilipDeegan: oh there should be no diff, I will check that, I made this from another branch

src/simulator/simulator_def.hpp
PhilipDeegan: this is PR is more about where the code lives rather than the actual values, there should be no diff from master except where the code is placed

